PR #272, opened 2015-03-01
UI Cleanup - Better Code Inspection UX

ckuhn203:
This pull request is to address issues detailed in #258.
It doesn't address it in exactly the way specified, but is perhaps good enough for a v1.2 release.
Everything is working, but I'm not absolutely certain I did not adversely affect performance of an already time consuming process. This needs to be reviewed and tested on a larger project before merging.


ckuhn203:
Code Review also posted here https://codereview.stackexchange.com/questions/82922/notifying-the-ui-that-issues-were-found

ckuhn203:
This can probably be merged as is, but it does not resolve any issues with the Toolbar. The toolbar is still running synchronously and will probably freeze large projects.

PR #428, opened 2015-05-03
Inno installer script updates.

rossknudsen:
Hi guys,

Here are some changes to the installer script based on comments from previous PR.


ckuhn203:
Looks good to me, granted. I don't know much about InnoScript, but it look like all prior concerns have been addressed. :+1:

PR #447, opened 2015-05-09
Architecture Change

ckuhn203:
This is a pretty big change to the project's architecture. Previously, we more or less had everything shoved inside of the Rubberduck.dll file. This meant that with each new release, we were breaking references to the type library for any client code (including Unit Testing). It also meant that code completely unrelated to the grammar file also had to be licensed under the GPL license. This is my attempt to remedy that.

There are now 5 separate projects and dll files.

 - VBEditor: Core VBE Extension functionality. (MIT)
 - Unit Testing: Everything in the old `UnitTesting` folder/namespace moved here. Uses the parser, which uses the grammar, so this remains GPL.
 - Source Control: Exposes our Source Control API. (MIT)
 - Parsing: Parses. uses the GPL grammar. 'nuff said.
 - Rubberduck: This is the actual add-in and GUI

Personally, I really like this change. I think it makes it much easier to navigate the project and find the functionality we're interested in. It also has the ***huge*** benefit that now we only have to break API references if the API themselves change.

However, this is big and risky, so I'd like to hear some feedback and have this scrutinized a bit before we merge the changes in. Thoughts?


ckuhn203:
Damn it. Wrong branch.

PR #501, opened 2015-05-18
Fix reorder bug

Hosch250:


Hosch250:
Whoops, this fixes a bug that caused the parameters of method calls to be reordered incorrectly.

PR #616, opened 2015-06-12
Appveyor - Set up project for continuous integration

ckuhn203:
I copied the 2007 Office PIAs into the project and linked to them there rather than the GAC. Builds and tests are working on the server, but I really think this needs a manual test on your local machine before we pull this. 

Closes #607. Closes #113.

PR #617, opened 2015-06-13
Converted from PIAs to NetOffice libraries.

rossknudsen:
Hey y'all,

Here is what I have managed to get working with NetOffice in place of PIA's.  It runs on my end (without crashing) but some functionality is not in place.  I'm **not** really wanting this to be pulled but wanted this branch available for you to see and try out (you should be able to pull my branch to do that).

I'm really keen on your feedback about whether you would like to pursue this.  It could wait for the following release since you are so close and probably don't want to delay it further.

Enjoy.

### DIFF
--- a/RetailCoder.VBE/App.cs
+++ b/RetailCoder.VBE/App.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.Windows.Forms;
-using Microsoft.Vbe.Interop;
+using NetOffice.VBIDEApi;
 using Rubberduck.Inspections;
 using Rubberduck.Parsing;
 using Rubberduck.Parsing.VBA;
--- a/RetailCoder.VBE/Extension.cs
+++ b/RetailCoder.VBE/Extension.cs
@@ -3,7 +3,7 @@ using System.ComponentModel;
 using System.Runtime.InteropServices;
 using System.Windows.Forms;
 using Extensibility;
-using Microsoft.Vbe.Interop;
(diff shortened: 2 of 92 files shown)


ckuhn203:
Don't pay attention to the build failure. I had to make some changes to the project to get the build server to work that haven't been pulled yet. 

It's great to know that #113 is a possibility if it turns out my attempt to make the project more portable didn't work. As of now, I can't say whether or not we'll pursue this though. I'm afraid of running into more inconsistencies with the library as the project grows. Let's give @retailcoder a chance to test #616. 

Or, if anyone else would like to test it, here's the Git commands to pull that branch local. 

    git remote add ducky https://github.com/ckuhn203/Rubberduck.git
    git fetch ducky
    git checkout ducky/appveyor

ckuhn203:
Also, would you mind expanding on what functionality "isn't in place"?

rossknudsen:
Sorry I'm spreading my information all over the show.  An example of what is not working:
You can open the Code Explorer but double clicking on an item doesn't open the respective code module.

I expect its because the API of NetOffice is slightly different to the way the COM API of the PIAs work.  I had to make a few changes here and there to get it to launch.  I didn't want to solve every problem for a proof-of-concept, if it wasn't to be accepted.

ckuhn203:
Ahhhh... yes. I could see how that would break our hack to force focus onto a code pane. So, yeah. Let's only pursue this if I failed at making the project portable.

ckuhn203:
#616 worked, so I'm going to shelve this and #113. Thanks for again for putting the effort in, though.

PR #628, opened 2015-06-15
Fix #602 for Office 2013 x64

### LINKED ISSUE #602: Switching Source Control Branches
When you switch source control branches, Excel crashes.

Hosch250:


ckuhn203:
Help me out here buddy. What does this fix? I see you moved from the struct to more args, but I'm having trouble understanding why.

Hosch250:
It is this block that the issue starts in:

    private void Refresh()
    {
        //Because refreshing removes all components, we need to store the current selection,
        // so we can correctly reset it once the files are imported from the repository.
        var selection = Project.VBE.ActiveCodePane.GetSelection();
        var name = selection.QualifiedName.Component.Name;

        Project.RemoveAllComponents();
        Project.ImportSourceFiles(CurrentRepository.LocalLocation);

        Project.VBE.SetSelection(selection.QualifiedName.Project, selection.Selection, name);
    }

When you call `Project.RemoveAllComponents();`, `selection.QualifiedName.Component`, which is a reference variable, is mostly destroyed, leaving a bunch of exceptions instead of fields/properties, including `selection.QualifiedName.Component.Name`.  We use this name in `Project.VBE.SetSelection()`, and it throws an exception the old way because it is destroyed.  So, I save the name directly, and pass each component used in instead of digging it out of `selection`.

PR #636, opened 2015-06-17
Unit testing - Moved Reference Ensurance from TestEngine to Middle Tier

ckuhn203:
The TestRunner shouldn't be messing around with the references. It's one thing to do it in the GUI. In the GUI, the reference is required to make the code compile, but someone using the TestRunner could be legitimately latebinding their code.

I lost the tests for it. Even after the last great TestExplorer refactor, it's still got a ton of dependencies and calls to static methods. I made some changes to make testing easier down the road, but wasn't able to get them done at this time.


ckuhn203:
It just occurred to me that I didn't test the TestRunner. Be sure to test this PR before merging.

ckuhn203:
image

TestRunner still works. Merging.

PR #642, opened 2015-06-18
Close issue #637 with tests

Hosch250:
Completely redid UI.  Did not remove duplicates from config file.


ckuhn203:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

ckuhn203:
Don't feel bad about the squash... I couldn't get it to work either. I'll go ahead and review the hard way. =;)-

RetailCoder.VBE/Settings/ConfigurationLoader.cs
```diff
@@ -113,7 +114,7 @@ public Configuration GetDefaultConfiguration()
         public ToDoMarker[] GetDefaultTodoMarkers()
         {
             var note = new ToDoMarker(RubberduckUI.ToDoMarkerNote, TodoPriority.Low);
-            var todo = new ToDoMarker(RubberduckUI.ToDoMarkerToDo, TodoPriority.Normal);
+            var todo = new ToDoMarker(RubberduckUI.ToDoMarkerToDo, TodoPriority.Medium);
```
ckuhn203: .....This is a breaking change. I don't know about this. I know we said we'd stuff all the breaking changes for the config into this release though, so... okay. We really need to treat the serialized classes like we would treat COM interfaces though. It's always okay to add to the interface, but never to change or remove from it.

ckuhn203:
The new form looks and works great. Since we're breaking the config for localization, I'm okay with the priority level break. Needs to be well documented though. I'll open an issue.

RetailCoder.VBE/Settings/ConfigurationLoader.cs
Hosch250: I thought I might as well change that since we were displaying "Medium" anyway.

PR #652, opened 2015-06-18
update appveyor webhook to the live Duga url

Zomis:


Zomis:
I think it is precisely what you think it is.

ckuhn203:
Thank you @Zomis!

PR #684, opened 2015-07-04
Rename tests in master too

Hosch250:


ckuhn203:
I'm not sure we want to pull this into master right now. Ideally, each commit to master signifies a release. I can't see creating a new release just to patch some test names.

Hosch250:
@ckuhn203 OK, I didn't realize you got the tests out before v.1.4.1

PR #701, opened 2015-07-08
Add Remove Params tests

Hosch250:


RubberduckTests/Refactoring/RemoveParametersTests.cs
```diff
@@ -349,5 +818,29 @@ private static Mock<IRefactoringPresenterFactory<IRemoveParametersPresenter>> Se
             factory.Setup(f => f.Create()).Returns(presenter.Object);
             return factory;
         }
+
+        private void SetupProject(string inputCode)
```
ckuhn203: This should be throwing a warning about hiding base class methods. Is this intended? Why not use the base implementation?

RubberduckTests/Refactoring/ReorderParametersTests.cs
```diff
@@ -140,5 +742,53 @@ private static Mock<IRefactoringPresenterFactory<IReorderParametersPresenter>> S
             factory.Setup(f => f.Create()).Returns(presenter.Object);
             return factory;
         }
+
+        private void SetupProject(string inputCode)
```
ckuhn203: Same deal. Either a change needed to be made to the base class, or a new subclass should be created.

PR #748, opened 2015-08-20
fix links in readme.md

yadimon:
from "retailcoder" to "rubberduck-vba" repo.


ckuhn203:
You're awesome! Thank you very much!

PR #779, opened 2015-09-18
Replaced "Menu Bar" by index number of the control bar so it should w…

INOPIAE:
…ork on all langugae installations


Vogel612:
"On all language Installations"... This may or may not be true. I'd personally like to see that list you have attached in #778 as named constants, So as to not rely on "they only translated the first 4 menu elements"

PR #905, opened 2015-12-27
Display type in UI for extract interface

Hosch250:


retailcoder:
Why would there be conflicts here? ...

retailcoder:
Please synchronize the PR

PR #1028, opened 2016-02-05
fix visual studio 2015 invalidcastexception for built-in declarations

autoboosh:
Apparently Visual Studio 2015 uses a new compiler which generates different code that in turn affects what GetNestedTypes() returns. See: http://stackoverflow.com/questions/31542076/behavior-of-assembly-gettypes-changed-in-visual-studio-2015.

This change resulted in a compiler generated class being returned for the built-in declarations which lead to an invalid cast exception. The fix simply ignores the compiler generated types.

Hosch250 tested the fix in Visual Studio 2013 and it seemed to work. I'm not sure if it'll work on Vista as well though.

### DIFF
--- a/Rubberduck.Parsing/Symbols/AdodbObjectModel.cs
+++ b/Rubberduck.Parsing/Symbols/AdodbObjectModel.cs
@@ -2,9 +2,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


ckuhn203:
This is a very welcomed fix. 
I'll leave it to the more active members to verify it works, but looks good at a glance.

Hosch250:
@autoboosh This will have nothing to do with working on Vista.  The only thing pertaining to that is the compiler we use to build the release.

PR #1117, opened 2016-02-19
About window

Hosch250:


Hosch250:
Seeing as I accidentally had part of it merged already, I figured I might as well get a fully working one in.

PR #1188, opened 2016-03-15
Translation

INOPIAE:
I worked on the German translation.
Would it make sense to separate the text from AboutWindow_SpecialThanksList into Headings that can be translated and the persons that are kept in the code?
The German text in the Indent Settings is wider then the canvas. Please check how this can be fixed. I thought this should be fixed automatically..


Hosch250:
We know about the text in the settings bug, but have not yet fixed it; we are going to make the text wrap around.  As for the Special Thanks list, I will separate them.  That way, it will only need to be changed in one place.  Thanks for the notice and translations.

PR #1221, opened 2016-03-20
Translation

INOPIAE:
Worked on the German translation


Hosch250:
Thanks for keeping abreast of this, @INOPIAE.  BTW, I just addressed your concerns in the settings and about windows recently.

PR #1255, opened 2016-03-25
Split combined parser into parser and lexer and fix a few reported grammar issues

autoboosh:


autoboosh:
There's no more commits to add. I'll add the annotation API with a separate PR.

Issue #1326, opened 2016-04-14
resolver/parser bug: addressof should have identifierreference created for referenced procedure


autoboosh:
Here's some example code:

```
Declare Function CallWindowProc Lib "user32" Alias "CallWindowProcA" ( _
      ByVal lpPrevWndFunc As Long, _
      ByVal hwnd As Long, _
      ByVal msg As Long, _
      ByVal wParam As Long, _
      ByVal lParam As Long) _
      As Long

Sub CallbackTest2()
   CallWindowProc AddressOf CallbackMsg, 0, 0, 0, 0
End Sub

Public Function CallbackMsg _
      (msg As String, _
      ByVal nUnused1 As Long, _
      ByVal nUnused2 As Long, _
      ByVal nUnused3 As Long)

   MsgBox "CallBack Work :)" & msg
End Function
```

`CallbackMsg `should have an identifierreference created for it but it doesn't (because it is used in the AddressOf expression in the `CallbackTest2 `method). Possible reason: addressof is treated specially in the parser.

Example code taken from http://stackoverflow.com/questions/24997849/vba-addressof-crash-office-app.

@retailcoder If the parser should be changed, let me know and I'll take care of it.

autoboosh:
Fixed with https://github.com/rubberduck-vba/Rubberduck/pull/1429

PR #1357, opened 2016-04-23
Allow valueStmts in "Implements" directive

autoboosh:
This change also features the introduction of the binding process based on expressions, mirroring the VBA spec as closely as possible. For this purpose a new expression parser has been introduced.


autoboosh:
Added some tests that try to validate on a basic level whether the "resolution namespaces" priorities are honoured.

autoboosh:
Please don't merge this yet. I just realized that adding member access expressions shouldn't be too difficult and will allow implements statements like "VBAProject.ITest" to be used.

autoboosh:
With the latest commit binding resolution of member access expressions in the type binding context are now possible. That means that the following example code:

`VBAProject.ITest`

Now creates two references, one to the VBAProject and one to the ITest class. The way it works is by creating a tree of IExpressionBinding. This tree gets "resolved" to a IBoundExpression tree. This tree is then visited/walked by the identifier resolver and for each visited IBoundExpression the identifier reference is added to the resolved declaration.

Tests are still missing, I'll add those asap.

autoboosh:
This PR should be ready to merge. If there's something that has to be fixed, let me know.

Regarding this issue: https://github.com/rubberduck-vba/Rubberduck/issues/1365. Expresions are slowly integrated one by one. The problem in that issue is in relation with call statements which allow many more expressions than are currently implemented. But we'll get there :)

autoboosh:
The latest commit essentially adds support for AddressOf resolution (see also https://github.com/rubberduck-vba/Rubberduck/issues/1326) since the "procedure pointer binding context" is only used in the AddressOf expression.

autoboosh:
The parser had problems dealing with the new keywords so I jumped at the opportunity and removed the ambiguous/certain identifiers and replaced them with "normal" identifiers.

autoboosh:
Latest commit replaces the old "as type" resolution with the new binding resolution. It seems to work OK. Referenced projects such as DAO get resolved without problems.

Also, name lookups are now case insensitive. Some other small bug fixes included as well.

autoboosh:
That's it for the changes. I'll try and do the "normal" expression resolution in a separate PR because that'll be quite a big change as well if that's OK.

PR #1409, opened 2016-04-29
Fixed the QuickFix menu localization

Vogel612:
see #1405, this should've happened instead of the hack at #1408


RetailCoder.VBE/Inspections/InspectionsUI.de.resx
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
Vogel612: hrngh... fml this is really annoying. next commit with TranslationHelper will change this back again.

PR #1429, opened 2016-05-04
refactored resolver

autoboosh:


retailcoder:
Just dumping notes / question marks:

How does `Width` resolve to a member of `VBA._HiddenModule` here, given `ProgressBar` is known to be a `DeclarationType:Control`?

image

Same here, except `Caption` doesn't resolve at all:

image

>     Unknown identifier in 'Control': 'Caption'
>     Unknown identifier in '(unknown)': 'Caption'

`QueryClose` members resolving correctly is a new and welcome improvement:

image

 :+1:

`As [ClassModuleName]` doesn't seem to produce identifier references as it should:

image

...but `As [UserDefinedType]` works:

image

Something is wrong with the resolving logic of "AsTypeClause":

image

The resolver should be looking for a *type* here - it shouldn't matter that there's a "closer" member that matches the identifier, the member isn't a type.

I like that we resolve custom members in subclasses:

image

But we need to look for base members too:

image

Does this resolve out of sheer luck?

image

...because `Workbook` in `As Workbook` isn't picked up as a reference to the `_Workbook` type (probably the same problem noted above though).

We should be able to resolve `Application`:

image

>      Unknown identifier in '(unknown)': 'Application'

...and its `ActiveWorkbook` member:

image

Although...

>     Unknown identifier in '(unknown)': 'Cells'

Not sure why this member wouldn't resolve:

image

Imported function calls don't seem to resolve either:

image

>     Unknown identifier in '(unknown)': 'Sleep'

The `Raise` and `Clear` members of `ErrObject` don't resolve (they didn't before either, not sure why) - that's crazy because every property resolves fine:

image

This is scaring me, and points to a silent parse tree bug:

>     Unknown identifier in '(unknown)': 'Else'
>     Unknown identifier in '(unknown)': 'Then'
>     Unknown identifier in '(unknown)': 'ElseIf'

retailcoder:
Regarding the last point about the silent parse tree bug; I seem to repro the problem with this procedure:

    Private Sub view_Activated()
    
        On Error GoTo CleanFail
        ThrowIfNotInitialized
    
        If Not this.instance Is Nothing Then
            ExecuteInstanceMethod
        Else
            ExecuteMacro
        End If
    
    CleanExit:
        view.Hide
        Exit Sub
    
    CleanFail:
        MsgBox Err.Number & vbTab & Err.Description, vbCritical, "Error"
        Resume CleanExit
    End Sub

>     Unknown identifier in '(unknown)': 'Else'

retailcoder:
I had comments, but they were based on my local branch not being sync'd with remote. Apologies. I added a failing test for a parser bug, and after syncing the new test passes. Re-reviewing now :)

retailcoder:
The `Err` function is known to return an `ErrObject`, yet the `ErrObject` members don't seem to resolve - pre-merge, `ErrObject` members *did* resolve, ...except the methods for some reason:

image

image

image

`Application` and its members don't resolve either:

image

The parse tree seems fine, so I suppose `Application.Run` parses as a member/procedure call, which would make the rest of the code line an `argListContext` - in which case I'm not sure I understand why these don't resolve (`this` and its members resolve fine elsewhere in the module):

image

autoboosh:
Regarding the Err/Application issues. In the case of the ErrObject, the declaration containing the properties/methods is actually called _ErrObject which seems to be the same problem as we've seen before. It finds the Err property, which is inside an accessible module called VBA.Information. Its type is ErrObject however thus when it tries to find the e.g. Description property in ErrObject it won't find it because it's actually in _ErrObject. Same problem with Application. It's actually the interface that contains all the declarations, which is also called _Application. I'm surprised it worked before, how did we look up those names?

retailcoder:
@autoboosh haha.. by cheating!

        public IEnumerable<Declaration> MatchName(string name)
        {
            var normalizedName = name.ToLowerInvariant();
            Declaration[] result;
            if (_declarationsByName.TryGetValue(normalizedName, out result))
            {
                return result;
            }
            if (_declarationsByName.TryGetValue("_" + normalizedName, out result))
            {
                return result;
            }
            if (_declarationsByName.TryGetValue("i" + normalizedName, out result))
            {
                return result;
            }
            if (_declarationsByName.TryGetValue("_i" + normalizedName, out result))
            {
                return result;
            }

            return new List<Declaration>();
        }

autoboosh:
Ah, yes of course. No wonder it's broken now.

How about we add a "Implements interfaces" property to the ClassModuleDeclaration, a list of declarations. Then we could walk up the hierarchy if we can't find a member. We could do this for user code as well so that e.g. refactorings and inspections have easier access to the implemented interfaces.

I'm not sure if that should cover the subclasses/inheritance stuff we found with ThisWorkbook, Sheet1 etc. though.

What do you think?

Oh and also, I've tried to make that TextBox1.Text = "Something" example work in MS Access. But there aren't even any control declarations created because the VBComponent.Designer is always null. Any ideas?

retailcoder:
That would be THE ideal solution! Yes, by all means go ahead with it! I'll see if I can tweak the COM reflection stuff to get those implemented interfaces; the only problem will be to determine that a worksheet document is a `Worksheet` object (and that `ThisWorkbook` is a `Workbook`), and then it will all just fall into place.

As for Access, it never worked (I believe there's an issue about it somewhere), I'm not sure how to go about it.

autoboosh:
Isn't the Worksheet/ThisWorkbook problem covered by looking up the parent property?

I'll try Excel with those control tests then, thanks for the info.

retailcoder:
Sort of. It's incomplete; the parent of `ThisWorkbook` gives us the `Application` object, and the parent of `Sheet1` gives us a `Workbook` object... which kinda breaks the type resolution. There doesn't seem to be any easy way of getting the actual underlying type of a document module, which is utterly annoying because then we have to implement host-specific resolver code and I'd really do anything to avoid that.

autoboosh:
Ah, too bad.

This is slightly off-topic but do you know if we have access to aliases in the type info libraries? E.g. if we create a declaration for the _B_var_Left, do we know that it's also called Left? If so, perhaps we could simply add a list of aliases to each declaration and include that in the lookups as well? Just a thought.

retailcoder:
If it's visible *somewhere* in the `ITypeInfo` /  `FUNCDESC`, then yeah we should be able to know what aliases there are: not sure how, but we'll figure it out.

Worst-case, I don't really have a problem with hard-coding these few special cases either: it takes whatever it takes :smiley:

autoboosh:
The latest commit adds the possibility of comments that look like annotations but are not.

I'm not sure if that control declaration test is even executable by the way, since we would have to load all built-in declarations.

retailcoder:
> *I'm not sure if that control declaration test is even executable by the way, since we would have to load all built-in declarations.*

@autoboosh that's correct. but I have a plan.. I think the test setup can manually add a fake built-in `MSForms.Control` declaration to the parser state.

retailcoder:
When late-bound, `Assert` member calls don't resolve as references to the module-level `Assert` field in a test module.

When early-bound, references to `Rubberduck.AssertClass` resolve correctly... but somehow not its members. Any ideas? Could it have something to do with the way we're registering it?

autoboosh:
The first two (Rubberduck.AssertClass + Excel.Worksheet) are the result of the same bug. Didn't pass the correct context to create the selection with. That's fixed now:

rd_project

rd_assertclass

Thanks for pointing that out :)

The Worksheet.Name one is the interface problem. As soon as those type hierarchies are linked up properly it should work. I did it manually and it does work:

            var ws = (ClassModuleDeclaration)declarations.Where(p => p.DeclarationType == DeclarationType.ClassModule && p.IdentifierName == "Worksheet").FirstOrDefault();
            var wsInterface = (ClassModuleDeclaration)declarations.Where(p => p.DeclarationType == DeclarationType.ClassModule && p.IdentifierName == "_Worksheet").FirstOrDefault();
            ws.AddSupertype(wsInterface);
            wsInterface.AddSubtype(ws);

Resolving Name then works as expected:

image

autoboosh:
Regarding the Control tests. For it to work we'd have to change how the declarations for controls are created a bit, because the current resolver doesn't look at DeclarationType.Control.
We could make them default instance variable class modules and link them up with the project as a parent. We could just add a flag "IsControl" so that they are still identifiable by the rest of the code base.

Also it seems that the designer.Controls=null problem is related to the designer being dynamic and the Controls property being explicitly implemented. Dynamic variables apparently can't deal with explicit interfaces so that's why it couldn't find the Controls property. Casting the designer to UserForm fixes the problem.

autoboosh:
MatchName is still in use. Just saw your question, totally missed that one. It only looks for exact matches now though.

autoboosh:
This PR is getting quite big. What do you think of ignoring the Control-Test for now, merging this PR and then looking at the hierarchy stuff separately?

retailcoder:
I'm worried about the regression of controls no longer being picked up as fields of the form class. But you're right, that PR is huge enough as it is. Ignore the test and I'll merge it.

autoboosh:
Found a GUID/IDENTIFIER precedence bug that should be fixed now.

The control test is disabled for now. It wouldn't work without figuring out what kind of control it is anyway.

Also, there's a DeclarationType.UserForm that doesn't seem to be used anywhere when creating a declaration. Those won't work with the resolver right now. Some inspections/refactorings seem to check for that type though.

Other than that this should be mergable. If any issues appear I'll try to fix them as soon as possible.

retailcoder:
DeclarationType.UserForm was a mistake, it should be nuked from orbit.. to be sure. :-)

PR #1535, opened 2016-05-18
Close #1531.  Remove some dead bug comments.

### LINKED ISSUE #1531: Inspections: Left to Left$ gone wrong
It's trying to add the $ to the wrong place:
image

```
Option Explicit

Public Function tester() As String

Dim str As String

str = "teeeeeeeeeeeeeeeeeest"


tester = Left(str, 4)

End Function
```

Should be Left$(arg,arg) not Left(arg,arg)$. the latter doesn't compile.

But anyway the parser seems to be getting more stable ^^ it doesn't crash as much as just a few days before :) great work !

Hosch250:


RetailCoder.VBE/Inspections/UntypedFunctionUsageInspectionResult.cs
```diff
@@ -32,18 +32,20 @@ public override string Description
     public class UntypedFunctionUsageQuickFix : CodeInspectionQuickFix
     {
         public UntypedFunctionUsageQuickFix(ParserRuleContext context, QualifiedSelection selection) 
-            : base(context, selection, string.Format(InspectionsUI.QuickFixUseTypedFunction_, context.GetText(), context.GetText() + "$"))
+            : base(context, selection,
+                  string.Format(InspectionsUI.QuickFixUseTypedFunction_,
+                      context.GetText(), context.GetText().Insert(context.GetText().IndexOf('('), "$")))
```
Vogel612: This seems to be blowing up with #1537 now ... Might need to check our assumptions here

PR #1569, opened 2016-05-23
Logging

autoboosh:


autoboosh:
Looks like I broke something? It was a simple find and replace commit anyway, I'll create a separate branch and come back later :)

Issue #1606, opened 2016-05-27
Opening settings dialog throws indenter error


autoboosh:
Here's the error:

```
System.Windows.Data Error: 17 : Cannot get 'PreviewSampleCode' value (type 'String') from '' (type 'IndenterSettingsViewModel'). BindingExpression:Path=PreviewSampleCode; DataItem='IndenterSettingsViewModel' (HashCode=61415348); target element is 'TextBlock' (Name=''); target property is 'Text' (type 'String') TargetInvocationException:'System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentOutOfRangeException: 'count' must be non-negative.
Parameter name: count
   at System.String.CtorCharCount(Char c, Int32 count)
   at Rubberduck.SmartIndenter.Indenter.Indent(String[] codeLines, String moduleName, Boolean reportProgress, Int32 linesAlreadyRebuilt) in D:\Temp\Rubberduck\Rubberduck.SmartIndenter\Indenter.cs:line 471
   at Rubberduck.UI.Settings.IndenterSettingsViewModel.get_PreviewSampleCode() in D:\Temp\Rubberduck\RetailCoder.VBE\UI\Settings\IndenterSettingsViewModel.cs:line 309
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at MS.Internal.Data.PropertyPathWorker.GetValue(Object item, Int32 level)
   at MS.Internal.Data.PropertyPathWorker.RawValue(Int32 k)'
```

Line 471 in Indenter.cs:

`currentLine += new string(' ', gap) + right;`

There's also another error right after the stack trace, not sure if it's the same issue:

[...]

Issue #1291, opened 2016-04-03
VB6 SmartIndenter applies indentation incorrectly for comments with line continuations


ThunderFrame:
If I comment out a procedure (using the VBE toolbar), and then SmartIndent... The line-continued comments are indented...

```vb
Option Explicit

Sub MySub1()
'VBE Tab width = 2
'With `Align continued strings and parameters` setting set to `Unchecked`
    Dim x1 As _
            Integer: _
            Dim x2 As _
            Integer
    ReDim x2(0) As _
            Integer
End Sub
```
becomes
```vb
Option Explicit

'Sub MySub1()
''VBE Tab width = 2
''With `Align continued strings and parameters` setting set to `Unchecked`
'    Dim x1 As _
     '            Integer: _
     '            Dim x2 As _
     '            Integer
'    ReDim x2(0) As _
     '            Integer
'End Sub
'
```
And with `continued strings` set as Checked...
```vb
Option Explicit

Sub MySub1()
'VBE Tab width = 2
'With `Align continued strings and parameters` setting set to `Checked`
    Dim x1 As _
        Integer: _
        Dim x2 As _
            Integer
    ReDim x2(0) As _
          Integer
End Sub
```
it becomes...
```vb
Option Explicit

'Sub MySub1()
''VBE Tab width = 2
''With `Align continued strings and parameters` setting set to `Checked`
'    Dim x1 As _
     '        Integer: _
     '        Dim x2 As _
     '            Integer
'    ReDim x2(0) As _
     '          Integer
'End Sub
```

Issue #1286, opened 2016-04-02
VB6 SmartIndenter applies indentation incorrectly for Reserved Words in Type Declarations


ThunderFrame:
The VB6 SmartIndenter has a bug when indenting Type Declarations that include Member names that use reserved words. Specifically, the keywords `Enum`, `Function`, `Sub` and `Type`.

**`Enum` and `Type` are not indented correctly after SmartIndenter runs**
If `Enum` or `Type` is used as a member name, then SmartIndenter does not indent *that* member at all. All other members are indented correctly.
```vb
      'Original code, which is deliberately over-indented
      Private Type MyType
          x1 As Integer
          Enum As Integer
          x2 As Integer
          Type As Integer
          x3 As Integer
      End Type

'Code after SmartIndenter has run
Private Type MyType
    x1 As Integer
Enum As Integer
    x2 As Integer
Type As Integer
    x3 As Integer
End Type
```

**`Function` and `Sub` are not indented correctly after SmartIndenter runs**
If `Function` or `Sub` is used as a member name, then SmartIndenter does not indent *that member and all subsequent members* correctly.

```vb
      'Original code, which is deliberately over-indented
      Private Type MyFuncType
          x1 As Integer
          Function As Integer
          x2 As Integer
          x3 As Integer
      End Type

      Private Type MySubType
          x1 As Integer
          Sub As Integer
          x2 As Integer
          x3 As Integer
      End Type

'Code after SmartIndenter has run
Private Type MyFuncType
    x1 As Integer
Function As Integer
x2 As Integer
x3 As Integer
End Type

Private Type MySubType
    x1 As Integer
Sub As Integer
x2 As Integer
x3 As Integer
End Type
```

comintern:
Does VBA accept _any_ identifier as an Type member?

Private Type Foo
    If As Integer
    Select As Integer
    Split As String
    For As Integer
    Dim As Integer
    Then As Integer
    UBound As Variant
    To As Integer
    Or As Integer
    Case As Integer
End Type

Private Sub WTF()
    Dim Bar As Foo
    
    With Bar
        .UBound = Split(.Split)
        If Not .Then Xor .If Then
            For .For = UBound(.UBound) To .Dim
                Select Case .Select
                    Case .Case
                        .Or = .Or Or .Or + UBound(.UBound)
                End Select
            Next .For
        End If
        Debug.Print .Or
    End With
End Sub

ThunderFrame:
[...]

PR #1624, opened 2016-05-29
Indenter.BallOfMud = new MuchSmallerBallOfMud(); Changed Assembly load to host context.

comintern:
Closes [#1291](https://github.com/rubberduck-vba/Rubberduck/issues/1291), [#1286](https://github.com/rubberduck-vba/Rubberduck/issues/1286), and [#1606](https://github.com/rubberduck-vba/Rubberduck/issues/1606).

### DIFF
--- a/RetailCoder.VBE/Extension.cs
+++ b/RetailCoder.VBE/Extension.cs
@@ -55,13 +55,14 @@ namespace Rubberduck
 
         Assembly LoadFromSameFolder(object sender, ResolveEventArgs args)
         {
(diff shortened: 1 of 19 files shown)


comintern:
It seems like passing the `ITypeInfo` to `CreateParameterDeclaration` in #1583 changed the lifespan of the object, so the timing of releasing the structs is off again.  So... rolling back the changes from #1549 prevents the access violations when the typelibs are being read and defers them to when the host is closed.  Exit time access violation > runtime access violation.

PR #1658, opened 2016-06-01
Convert GetDeclarationsForReference return to List<Declaration> 

comintern:
Not needed and need to rule out the state machine's interaction with OLE.


comintern:
<s>CreateParameterDeclaration</s>  CreateMemberDeclaration passing a FUNCDESC as an out parameter didn't allow releasing the struct properly if `(memberDescriptor.callconv != CALLCONV.CC_STDCALL)` because it doesn't have access to the pointer.

PR #1707, opened 2016-06-05
Adds Unit Tests to Outlook

ThunderFrame:
@ckuhn203 This one's for you.


ckuhn203:
Beautiful hack! Just a couple of things. Why the empty catch? Mind adding some comments explaining why we need the hack to begin with? The current team know, but future maintainers might not immediately  understand that Outlook doesn't have an `Application.Run()` method.

Issue #1722, opened 2016-06-07
Workbook created via SaveAs ends up with `HelpFile/ProjectID` collision, causes parse problems


ThunderFrame:
I have 2 workbooks. The 2nd was created by doing a SaveAs from the first. I notice that TestChecker1.xlsm has HelpFile value **29288625** and TestChecker2.xlsm has helpfile value **29288625**

*So much for one-in-a-million chance*

I start a fresh copy of Excel 2016 32-bit and open TestChecker1, followed by TestChecker2. Only then do I open the VBE.

Rubberduck parses, and Code Explorer shows me:
![Imgur](http://i.imgur.com/Jo0FFID.png)

Notice the absence of TestChecker2, so I close TestChecker2 and Reopen it, and Rubberduck reparses:
![Imgur](http://i.imgur.com/5nra2He.png)

So Test Explorer and Code Explorer get muddled with the 2nd workbook.

PR #1723, opened 2016-06-07
Careful assignment of ProjectID - Avoids project collisions

ThunderFrame:
All @Hosch250's work

I did get some problems trying to execute tests across both projects, but I suspect that's another issue.


ThunderFrame:
Closes #1722

PR #1800, opened 2016-06-12
fix tokens should allow line continuations (fixes #1786)

### LINKED ISSUE #1786: Line break in End With, and End *Procedure* statements breaks parser
>mismatched input '\r\n' expecting ':='

```vb
Sub test1()
  With Application
  End _
  With
End Sub
```

autoboosh:
There were a number of other tokens with the same problem. They should all be fixed now, hopefully I didn't miss any. Just like @ThunderFrame said, end_enum and end_type only allow whitespace (no line continuations). I'm not quite sure why that is but there's probably a reason for that I don't understand. To stay on the safe side those were left as they were.


ThunderFrame:
@autoboosh You know we're getting close to a perfect grammar when we start finding inconsistencies in the *actual* VBE grammar.

Issue #1841, opened 2016-06-16
Indenter settings


Talorthain:
Hi,

Changing the settings for the indentation, dont save when click OK. The defaults work correctly.

File version is: Rubberduck.Setup.2.0.1b (updated 1)

Regards

Anthony

p.s newest version is a big jump in the right direction. Brilliant work.

![settings after](https://cloud.githubusercontent.com/assets/6796503/16110939/419bdfde-33a8-11e6-9452-bffc991ba4f8.PNG)
![settings](https://cloud.githubusercontent.com/assets/6796503/16110940/419db71e-33a8-11e6-82ef-dedb01be0f49.PNG)

Hosch250:
No repro here.  That is very peculiar...

comintern:
No repro here either.

Talorthain:
I set the options and when I click ok and go back in, they are back to the defaults. Could this be an antivirus issue?

 

I am using office 2016 desktop – Excel.

 

Regards

 

Anthony

Hosch250:
No, this isn't an AV issue.

Talorthain:
I had no issues when I installed the program and I used the program on a clean excel project.

Talorthain:
I will have to reinstall it. I had to remove, becuase I am working on a time critical project (e.g my bosses expect it sooner rather than later!!)

I will do tomorrow..

Talorthain:
Hi,
I have reinstalled the program and I am getting the same issue. What is the file you need? I cant see a config file in the rubberduck folder...

Talorthain:
Hi, I have reinstalled. And I have the same issue. what file do you need me to attach, because I can see a config file in the rubberduck folder?

Hosch250:
OK, that may well be the bug.  Which folder are you looking in?  It should be in `C:\Users\{user}\AppData\Roaming\Rubberduck`.  And you don't get a `rubberduck.config` file if you press OK in the settings view?  What about if you press Reset, then OK?

Talorthain:
Hi,

 

                I have just checked the location you have stated and I don’t even have a rubberduck folder, never mind a config file. I have tried reset and ok, but nothing.

 

                I am also getting a crash on exit of excel.

 

 

Regards

 

Anthony

comintern:
[...]

PR #1870, opened 2016-06-18
Ensure settings path exists. See #1841

comintern:


RetailCoder.VBE/App.cs
```diff
@@ -169,13 +170,29 @@ private void _configService_SettingsChanged(object sender, ConfigurationChangedE
...
+            try
+            {
+                if (!Directory.Exists(ApplicationConstants.LOG_FOLDER_PATH))
+                {
+                    Directory.CreateDirectory(ApplicationConstants.LOG_FOLDER_PATH);
+                }
+            }
+            catch
+            {
+                //Does this need to display some sort of dialog?
```
Vogel612: I'm not even sure we should catch here. I daresay a Fail-Fast is better. Blow up rather sooner than later and either let this bubble or show a dialog and stop loading RD.  
Since we would blow up sooner or later anyways with the directory not existing or having failed to be created we can basically stop the world (at least the RD world)

Hosch250: Make sense.  Just show some sort of dialog and re-throw.  Let it bubble into Extension.cs and shut RD down.

PR #1890, opened 2016-06-21
Make resolve sequence resolve references on all components, modified …

Hosch250:
…or not.

### DIFF
--- a/Rubberduck.Parsing/VBA/RubberduckParser.cs
+++ b/Rubberduck.Parsing/VBA/RubberduckParser.cs
@@ -613,11 +613,9 @@ namespace Rubberduck.Parsing.VBA
 
         private void ResolveReferences(DeclarationFinder finder, VBComponent component, IParseTree tree)
         {
+            Debug.Assert(State.Status == ParserState.ResolvedDeclarations);
+
             var state = _state.GetModuleState(component);
-            if (state != ParserState.Resolving)
-            {
-                return;
-            }
             var qualifiedName = new QualifiedModuleName(component);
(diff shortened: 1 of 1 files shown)


Rubberduck.Parsing/VBA/RubberduckParser.cs
```diff
@@ -613,11 +613,9 @@ private Declaration CreateProjectDeclaration(QualifiedModuleName projectQualifie
 
         private void ResolveReferences(DeclarationFinder finder, VBComponent component, IParseTree tree)
         {
+            Debug.Assert(State.Status == ParserState.ResolvedDeclarations);
+
             var state = _state.GetModuleState(component);
```
Vogel612: you should probably remove this one, since it causes some major mayhem down below in the catch-blocks

Hosch250: Which one?

Vogel612: `var state` should actually be unnecessary for the rest of the method, since the catch blocks should access `_state` instead of setting this variable.

Hosch250: try
                {
                    ...
                    state = ParserState.Ready;
                }
                catch (Exception exception)
                {
                    ...
                    state = ParserState.ResolverError;
                }

            _state.SetModuleState(component, state);

Hosch250: You are right, however, that I can reduce calculations here.

Issue #2099, opened 2016-07-20
RD had an error, #2


daFreeMan:
I hit <kbd>Ctrl</kbd><kbd>M</kbd> to indent a module and got a RD error.

[rubberduck.log.txt](https://github.com/rubberduck-vba/Rubberduck/files/373902/rubberduck.log.txt)

Vogel612:
Relevant stacktrace from inside the log:
```
2016-07-20 10:23:04.8895;ERROR;DynamicInjectorce6ff4e92e28401192dcbfde42d3c26e;Ninject.ActivationException: Error activating IInspection using binding from IInspection to method
A cyclical dependency was detected between the constructors of two services.

Activation path:
  3) Injection of dependency IInspection into parameter inspections of constructor of type ConfigurationLoader
  2) Injection of dependency IGeneralConfigService into parameter configService of constructor of type TestExplorerViewModel
  1) Request for TestExplorerViewModel

Suggestions:
  1) Ensure that you have not declared a dependency for IInspection on any implementations of the service.
  2) Consider combining the services into a single one to remove the cycle.
  3) Use property injection instead of constructor injection, and implement IInitializable
     if you need initialization logic to be run after property values have been injected.

   at Ninject.Activation.Context.Resolve()
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Linq.Enumerable.<CastIterator>d__94`1.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.HashSet`1.UnionWith(IEnumerable`1 other)
   at System.Collections.Generic.HashSet`1..ctor(IEnumerable`1 collection, IEqualityComparer`1 comparer)
   at Rubberduck.Settings.CodeInspectionConfigProvider.GetDefaultCodeInspections() in C:\Users\hosch\Documents\Visual Studio 2015\Projects\Rubberduck\RetailCoder.VBE\Settings\CodeInspectionConfigProvider.cs:line 47
   at Rubberduck.Settings.CodeInspectionConfigProvider.Create(IEnumerable`1 inspections) in C:\Users\hosch\Documents\Visual Studio 2015\Projects\Rubberduck\RetailCoder.VBE\Settings\CodeInspectionConfigProvider.cs:line 29
[...]

PR #2115, opened 2016-07-22
Remove IEnumerable<IInspection> dependency from CodeInspectionSettings

### LINKED ISSUE #2105: 2.0.6 Run-time error '-2146233088 in Source Control
When changing from branch next to branch master in Source Control, I've got Run-time error '-2146233088 (80131500)' after I got a message saying Checkout failed - 1 conflict prevents checkout. 

Error activating IInspection using binding from IInspection to method
A cyclical dependency was detected between the constructors of two services.

Activation path:
  4) Injection of dependency IInspection into parameter inspections of constructor of type ConfigurationLoader
  3) Injection of dependency ConfigurationLoader into parameter configLoader of constructor of type FolderHelper
  2) Injection of dependency FolderHelper into parameter folderHelper of constructor of type CodeExplorerViewModel
  1) Request for CodeExplorerViewModel

Suggestions:
  1) Ensure that you have not declared a dependency for IInspection on any implementations of the service.
  2) Consider combining the services into a single one to remove the cycle.
  3) Use property injection instead of constructor injection, and implement IInitializable
     if you need initialization logic to be run after property values have been injected.

comintern:
Closes #2105 and #2099.  Note this is a **temporary fix** that breaks backward compatibility in the rubberduck.config file. Consider it the minimum payment due on the technical debt card.


comintern:
f9cbca9 closes #1764.

Vogel612:
@comintern is there a follow-up issue for the [technical-debt]. Or did this clear some cruft and I just misunderstood what you were trying to say?

comintern:
@Vogel612 - No follow up issue - that just refers to the fact that there are still a couple of coupling issues in CodeInspectionSettings that make it fragile to work with.

Issue #2255, opened 2016-09-30
RD should have inspection for Enum Type value is assigned without Enum member reference


ThunderFrame:
Types declared as Enum should be assigned with Enum members and *not* with literal values or ambiguous member return values.

```vb
Option Explicit

Private Enum foo
  useprod = 1
  UseDev = 2
  UseUAT = 4
End Enum

Sub foo()
  Dim f As foo
  
  'Explicit assignments without member reference
  'This code uses the Enum, and depends upon 2 being a valid member
  f = 2
  f = GetAmbiguousFoo
    
  'Should be something like
  f = UseUAT
  'Or...
  f = useprod Or UseUAT
  'Or...
  f = GetFoo

End Sub

Private Function GetFoo() As foo
  GetFoo = UseUAT
End Function

Private Function GetAmbiguousFoo() As Long
  GetAmbiguousFoo = UseUAT
End Function
```

ThunderFrame:
If this inspection does end up being at the Enum *member* level, then it will need to consider Aliases and/or bitwise usages.

```vb
Private Enum foo
  useprod = 1
  UseDev = 2
  UseUAT = 4
  UseProduction = useprod
  UseProdOrDev = useprod Or useDev
End Enum
```

If `UseProduction` is used in consuming code, but `useprod` is never used outside of the Enum, then `UseProduction` could be removed, but only if usage of `UseProduction` was changed to use `useprod`

If neither `useprod` nor `usedev` are used in consuming code, but `UseProdOrDev` *is* used, then neither `useprod` nor `usedev` are necessarily removable.

PR #2462, opened 2016-12-26
Big refactor of COM reflection.  All your TypeLib are belong to us.

### LINKED ISSUE #1700: TLB crawler should collect `helpfile` and `helpcontext` IDs for TypeInfo members
Type libraries can have `helpfile` and `helpcontext` attributes. If we collect those details, we can facilitate access to offline help, as per #1053 .

Here's an example from the Excel OleViewer:

``` text
  helpstring("Microsoft Excel 16.0 Object Library"),
  helpfile("VBAXL10.CHM"),
  helpcontext(0x0000ffff),
```

comintern:
Note - serialization still isn't hooked back up because of hash collisions of `Declarations` (see my comment on #2447 for the reason _why_ there are hash collisions). I'll probably need to "de-consolidate" the Enums and Types (oh well).

- Now pulling Enum values, Const values, and default values for `Optional` parameters (ref #419, #2255, #2263).
- Button state might work better now, I found a non-local variable being used inside a closure in the COM collector.
- Now collecting "strongly typed" parameters that aren't reference types - i.e. VbMsgBoxResult.
- Reference collection should be much faster, type annotation is probably slower (see above).
- Now collecting all of the documentation information from the tlb (still not doing anything with it, but it's there for whatever use).  Closes #1700, ref #1053.
- Resolution of the default member should work better - it was just selecting the first member as default instead of testing the flags.
- General clean-up of the functionality.

Other things I forgot.  Note - this is currently a bit of a WIP, there may be things that got broken that I'm not aware of.


comintern:
Sorted enums, serialization/deserialization is sorted out. This duck can fly.

PR #2481, opened 2017-01-05
More tests for symbols and refactorings for CostumDeclarations

MDoerner:
I added some more tests and refactored the CostumDeclarations.

In particular, I brought the `AliasDeclaration` and `FormEventDeclarations` in line with the other `CostumDeclarations` by letting them use a `DeclarationFinder` to find the referenced  modules and not to hard-code anything in the declarations themselves.
Moreover, I changed the point in time when the declarations are queried from the `RubberduckParserState` for the `DebugDeclarations`. They are now refreshed on every load instead of only at the time of the construction of the `DebugDeclarations` object. 
I tried the same for the `SpecialFormDeclarations` but after the change the tests simply stalled.


MDoerner:
ref #2464

PR #2517, opened 2017-01-11
Test Categorization, fixes #2499 [WIP]

### LINKED ISSUE #2499: Categorize Tests
Currently a significant number of tests is "uncategorized". This means we don't have a reasonable way of checking which tests are testing "impactfully". Additionally the RubberduckTests project is getting humungous. 

It might be useful to extract tests for specific suprojects into their own testprojects and / or categorize them to provide a clean way of evaluating test runs. Also it helps to not have to run each and every unit-test when developing somewhere, beacause the 1.6k tests are taking quite some time by now. 

Ref #218 too. As of now about 1k tests are not marked with any category. It's also possible to mark tests with multiple categories to allow for something akin to TestSuites

Vogel612:


Vogel612:
hrngh ... github, just let me edit the remote head ref ...

PR #2570, opened 2017-01-23
COM collector fixes.

comintern:


comintern:
WARNING: This breaks any previously serialized files.

PR #2735, opened 2017-02-24
Add a modern folder dialog style.

comintern:
Left the old one in as a fallback on the off chance that RD is running on XP or earlier :cringe:


comintern:
@ckuhn203 - The XP style was already there, so I figured "why not".  Good point on the R# suppressions though - the newer dialog doesn't support hiding the new folder button.  I'll comment it in my next commit.  ;-)

ckuhn203:
Thanks @comintern!

Issue #2737, opened 2017-02-24
The `DeclarationFinder` cannot cope with multiple classes implementing the same interface.


MDoerner:
The problem is that the collection `interfaceMembers` in the constructor of the `DeclarationFinder` contains an item for each time an interface is implemented.

This collides with the lazy constructor of `_interfaceMembers`.

```
_interfaceMembers = new Lazy<ConcurrentDictionary<Declaration, Declaration[]>>(() =>
               new ConcurrentDictionary<Declaration, Declaration[]>(interfaceMembers.ToDictionary(item => item.InterfaceModule, item => item.InterfaceMembers.ToArray())), true);
```

Because there are duplicate keys, `ToDictionary` fails with an `ArgumentException`. 
This can only be observed when the inspections run because the constructor is lazy

Actually, it can only be observed on the second parse because the `DeclarationFinder` does not get refreshed after resolving the references. Because of this, the inspections in the first parse do not see which interfaces get implemented.

PR #2738, opened 2017-02-24
Remove references of not reparsed components.

MDoerner:
Made the `ParseCoordinator` remove all references upon reparse and not only those to the built-in declarations and the modules that have been modified.

This will have to change again, when I implement selective reference resolving.

This PR will make issue #2737 more obscure since there will never be a `DeclarationFinder` at whose time of construction any references are are resolved.


MDoerner:
I made the `DeclarationFinder` use distinct declarations for the interfaces. This should fix #2733.
However, it could still us a test to verify this does not break again. Well, that is true for everything in the `Declataion Finder`.

PR #2766, opened 2017-03-02
Fix extreme UI sluggishness, avoid throws in SubclassingWindow.

comintern:
Ref #2763


comintern:
Going ahead and merging this - consider it a hot-fix.

PR #2788, opened 2017-03-04
Updated German i18n for the .12 release

Vogel612:


retailcoder:
Awesome, good job @Vogel612!

PR #2942, opened 2017-04-01
WIP: Expose parse cancellation token to parser state handlers

Hosch250:
Check carefully before merging.


Rubberduck.Parsing/VBA/ParserStateChangedCallbackRunner.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            foreach (var callback in callbacks)
+            {
+                if (token.IsCancellationRequested) { break; }
+
+                Task.Run(() => callback(token), token);
+            }
+        }
+    }
+}
```
Hosch250: Sure.

Rubberduck.Parsing/VBA/RubberduckParserState.cs
```diff
@@ -68,8 +61,46 @@ internal void RefreshFinder(IHostApplication host)
...
+                    {
+                        callbacks = new ConcurrentDictionary<Action<CancellationToken>, byte>();
+                        _callbacks.Add(value, callbacks);
+                    }
+                }
+                else
+                {
+                    callbacks = _callbacks[value];
+                }
+
```
Hosch250: Yeah, for some reason, I thought I'd done that.

Rubberduck.Parsing/VBA/RubberduckParserState.cs
```diff
@@ -130,6 +162,8 @@ private void Sinks_ProjectRemoved(object sender, ProjectEventArgs e)
             OnParseRequested(sender);
         }
 
+        internal CancellationToken ParseCancellationToken { private get; set; }
+
```
Hosch250: Bother. I'll have to redesign this.

Hosch250:
Yeah, but when I got into it, I don't see any method of doing that cleanly. `ParseCoordinator` only knows about the state, not vice-versa. Mat said to have the registry in the `RPS`, because that is what the callers have. One method would involve firing an event "CallbackAdded," or something, the other would be exposing the dictionary of states/callbacks. I don't like either of these...

MDoerner:
Two more thoughts:

First, if we extracted the logic for storing, registering, unregistering, and running all registered callbacks for a state into a class implementing an `IParserStateChangeCallbackManager` interface, we could expose this both to the state and the coordinator; the state would delegate the registration and unregistration to the callback manager and the coordinator would trigger all callbacks via the callback manager without knowing what it really triggers.

Second, do we really want to remove the state changed event? We could use it to trigger the UI updates independently of running the callbacks, especially of the command bar. Then the callbacks would be reserved for addidional functionality running at certain points in the parsing run like the unit tests and inspections.

Hosch250:
Hmm, I like that first idea. Not too sure why we would need the event anymore, the way I set this up (or wait, we might need to update our UI's synchronously because of COM?).

PR #2976, opened 2017-04-16
german I18n

Vogel612:


Vogel612:
Note that not all resource strings were required to be changed in 9a3d145 because German can sometimes be less expressive than English ... Also I did already somewhat adjust the meaning in the direction the change was going, so adjustments haven't been necessary for all cases.

PR #3001, opened 2017-05-07
Preprocessing Without Changing the Code

MDoerner:
So far, the preprocessor took the module code, evaluated the `#If` blocks and then replaced all lines with precompiler directives and all dead lines with a single newline. This led to issue #2938 because the rewriters see only the tokens stream faor the modified code.

This PR addresses this issue. Instead of changing the code itself, the new preprocessor changes the channels of the tokens in the original token stream. Instead of replacing lines the parser should not see by new lines, the new preprocessor sets the channel of the corresponding tokens to `TokenConstants.HiddenChannel`, exclusding newlines and line continuations to keep line counts intact. This has the effect that the parser, which listens only to the default channel, does not see the tokens. However, they are still there and get returned when calling `GetText` on the rewriter.

So, this PR fixes issue #2938.

I took the liberty to also include a few more parser tests covering corner cases of call statements.


MDoerner:
I completely missed the issue with line numbers in #2938. This has not be addressed by this PR. (If it has been addressed it was an unintentional side effect.) So, this PR probably only fixes half of issue #2938.

Issue #2636, opened 2017-02-04
Line labels with code after them are not recognized as line labels.


SystemsModelling:
Two warnings about use of Again: label

(1)
	Variable 'Again' is used but not assigned	 
doubleclick points to white space left of "Dim Fname"
Function MacGetSaveAsFilenameExcel(MyInitialFilename As String, FileExtension As String)
      'Ron de Bruin, 03-April-2015
      'Custom function for the Mac to save the activeworkbook in the format you want.
      'If FileExtension = "" you can save in the following formats : xls, xlsx, xlsm, xlsb
      'You can also set FileExtension to the extension you want like "xlsx" for example
          Dim FName As Variant
          Dim FileFormatValue As Long
          Dim TestIfOpen As Workbook
          Dim FileExtGetSaveAsFilename As String

736 Again:           FName = False

Maybe it's the double label? Line number 736 AND a Goto target?
(2)
delete 736 and redo Code Inspection
Now get 
     	Variable 'Again' is never assigned	 

same pointing to white space

comintern:
For reference, the code is from [this link](http://www.rondebruin.nl/mac/mac016.htm), and the problem spot is here:
```
    Dim FileExtGetSaveAsFilename As String

Again:         FName = False
    'Call VBA GetSaveAsFilename
```
It appears that RD is treating the `:` marker in the label as a statement separator instead of a label marker.  TBH I had no idea that you could put code on the same line as a label (that should probably be an inspection itself, because the syntax on that line is completely ambiguous at a glance.

Good find!

comintern:
This is a bug in the grammar, not in the resolver.  The token `Again` is a match for `LExprContext` and the token is being consumed as a `SimpleNameExprContext`.

The problem is that the label rule is...

`identifierStatementLabel : unrestrictedIdentifier whiteSpace? COLON;`

...but the issue seems to be that there is nothing that anchors `identifierStatementLabel` to the beginning of the line. This is related to #1873, and it's possible that adding a "beginning of line" rule would be the direction to go here.  The tricky part is even a BOL is ambiguous because it can't be preceded with a line continuation.  For example, in this code `Bar:` is a label:

```
Sub Foo()
    Dim x As Long, y As Long, result As Long
Bar: result = x + y
    Debug.Print result
End Sub

Sub Bar()
    Debug.Print "Bar called"
End Sub
```
In this code, `Bar:` is (obviously) not:
```
Sub Foo()
[...]

PR #3014, opened 2017-05-11
Line Numbers

MDoerner:
This PR adresses #2938 and #2636.

In other words, with the grammar changes in introduced here, RD is able to parse and resolve line numbers and line labels with code on the same line.


MDoerner:
One thing I do not know ois whether RD currently hand;es correctly that there can be a label and a member of the same name, not hiding each other.


ambiguousvba

retailcoder:
:+1:  labels are always locally scoped; the resolver already handles that ambiguity.

PR #3372, opened 2017-09-12
Apply Refresh Message to Code Inspections window

shadowofsilicon:
This PR refactors the "Rubberduck doesn't see anything, Refresh plz" message from the Code Explorer into a separate reusable element, and applies it to the Code Inspections window.


Vogel612:
Wonder what's up with the deleted user opening the PR though?

PR #3410, opened 2017-10-03
Shadowed declarations performance tweaks

MDoerner:
This PR contains some performance tweaks for the `ShadowedDeclarationInspection`.

The is the same but looping through collections and walking trees is minimized using the work the parsing engine has already done.


rkapka:
@MDoerner I could swear that `IsPrivateModule` didn't work for `Option Private Module` and that's why I created that listener...

MDoerner:
@rkapka Note that I replaced `declaration` with `Declaration.GetModuleParent(declaration)`.

PR #3504, opened 2017-10-22
Simplify shadowed declaration tests

MDoerner:
This PR simplyfies the tests for the `ShadowedDeclarationInspection`. Instead of generating a separate VBE for each posiible combinattion in a test and then parsing it, as much is put into the same VBE, project or even component. This reduces the necessary amount of parsing consideraby. For me, this brings the time to run all tests down to 3m and 12s from previously 7m and 35s.

Moreover, all `RubberduckParserstate`s in the tests for this inspection are now wrapped in a using statement. This reduces the memory load on my system from a bit over 3GB while running tests to a bit over 2GB. This should actually be done in all tests using a `RubberduckParserState` since only disposing of it makes it unregister from the static events for project and component removal and addition.

As a small extra, this PR contains a fix for the registration of `IParseTreeInspaction`s and quickfixes, wich I messed up in my port to CW.

Moreover, I fixed three tests that made a very unfortunate choice of a component name, which hid away a bug in the inspection. Apart fom the unfortunate choice, these tests made wrong assumptions about the expected outcome. I fixed the assumptions and then fixed the inspection to honor the new assumptions.

Please note that, appart from these three inspections, this PR is purely a refactoring and doe not change the assumptions about the correct results of the inspection.

There are still some problems with the assumptions in the tests, cf. issue #3508.


MDoerner:
There is actually no bug on the inspection; the assumption of the three failing tests is just wrong in the first place.

I will fix them later today.

MDoerner:
No longer WIP.

Vogel612:
LGTM, but I can say nothing re the correctness of the tests

PR #3516, opened 2017-10-26
COMException fix and others

IvenBach:
COMException fix, a few string interpolation fixes, and explicit `{ }` for `if ( )` block.


IvenBach:
Should fix Issue #3058.

PR #3524, opened 2017-10-31
Folder annotation scope

IvenBach:
Fixes Issue #3393.


Vogel612:
Tip for the future: instead of merging `next` into your feature-branch, it doesn't make a new commit if you base your feature branch onto next like so: `git checkout -b feature-branch upstream/next`

PR #3573, opened 2017-11-27
Rubberduck.inspections c#6&7 update

IvenBach:
Updates for the Rubberduck.Inspections folder.


Rubberduck.Inspections/Concrete/ParameterCanBeByValInspection.cs
```diff
@@ -125,29 +124,16 @@ private static bool IsUsedAsByRefParam(IEnumerable<Declaration> declarations, De
...
+            //TODO convert to method syntax
+            return (from item in procedureCalls
+                        from declaration in items
+                            .Where(arg => arg.DeclarationType == DeclarationType.Parameter && arg.ParentScope == item.Key.Scope)
+                            .OrderBy(arg => arg.Selection.StartLine)
+                            .ThenBy(arg => arg.Selection.StartColumn)
+                            .ToArray()
+                        where ((VBAParser.ArgContext)declaration.Context).BYVAL() == null
+                        select declaration)
+                    .Any(dec => dec.References.Any(reference => reference.IsAssignment));
```
Hosch250: I fail to see how this is more clear than the original. And it is certainly slower.

Rubberduck.Inspections/Concrete/ShadowedDeclarationInspection.cs
```diff
@@ -276,15 +259,14 @@ private static bool DeclarationInTheSameComponentCanBeShadowed(Declaration origi
...
-                return false;
-            }
-
-            if (originalDeclaration.DeclarationType == DeclarationType.Parameter || originalDeclaration.DeclarationType == DeclarationType.UserDefinedTypeMember ||
-                originalDeclaration.DeclarationType == DeclarationType.LineLabel)
+            switch (originalDeclaration.DeclarationType)
             {
-                return false;
+                case DeclarationType.UserDefinedType:
+                    return false;
```
Hosch250: Why is this one set in its own block?

Rubberduck.Inspections/QuickFixes/AddIdentifierToWhiteListQuickFix.cs
```diff
@@ -27,13 +27,10 @@ public override void Fix(IInspectionResult result)
...
-        public override string Description(IInspectionResult result)
-        {
-            return InspectionsUI.WhiteListIdentifierQuickFix;
-        }
+        public override string Description(IInspectionResult result) => InspectionsUI.WhiteListIdentifierQuickFix;
 
-        public override bool CanFixInProcedure { get; } = false;
-        public override bool CanFixInModule { get; } = false;
-        public override bool CanFixInProject { get; } = false;
+        public override bool CanFixInProcedure => false;
```
Hosch250: I sure hope you realize the difference between these... It's not a problem here, but you'd better know it because it will result in different values in certain cases.

Hosch250: Also, I prefer the first here.

Rubberduck.Inspections/Concrete/ShadowedDeclarationInspection.cs
IvenBach: Didn't catch it could be combined.

PR #3588, opened 2017-12-02
Feature extract method Update 2

bclothier:
Continuation of previous PR that was accidentally merged prematurely.


RetailCoder.VBE/UI/RubberduckUI.resx
```diff
@@ -2176,4 +2176,7 @@ Would you like to import them to Rubberduck?</value>
   <data name="ExtractMethod_NoneSelected" xml:space="preserve">
     <value>(None)</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="ExtractMethod_CompilationDirectivesWarning" xml:space="preserve">
+    <value>The module contains compilation directives (e.g. #Const, #If, ##lse or #End If). The refactoring can only operate on the active blocks. You must manually update the inactive blocks after the refactoring.</value>
```
Vogel612: typo ##lse should be #Else

Rubberduck.VBEEditor/Extensions/CodeModuleExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+            return regex.Matches(rawCode).Count > 0;
+        }
+
+        public static bool ContainsCompilationDirectives(this ICodeModule codeModule)
+        {
+            var rawCode = string.Join(Environment.NewLine,
+                codeModule.GetLines(1, codeModule.CountOfLines));
+            var regex = new Regex(@"^(\s?)+#", RegexOptions.Multiline);
+            return (regex.Matches(rawCode).Count > 0);
+        }
```
Vogel612: I think I'd rewrite this to call something like:

```
private static bool ContainsCompilationDirectives(string rawCode) 
{
    // local static may not be supported, if it isn't extract it into a constant
    static var regex = new Regex(@"^\s*#", RegexOptions.Multiline);
    return regex.Matches(rawCode).Count > 0;
}
```

and just call this from the other extension methods

Rubberduck.VBEEditor/Extensions/CodeModuleExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        public static bool ContainsCompilationDirectives(this ICodeModule codeModule)
+        {
+            var rawCode = string.Join(Environment.NewLine,
+                codeModule.GetLines(1, codeModule.CountOfLines));
+            return ContainsCompilationDirectives(rawCode);
+        }
+
+        private static bool ContainsCompilationDirectives(string rawCode)
+        {
+            const string regexExpression = @"^(\s?)+#";
```
Vogel612: I did change the regex. Why did you not?

bclothier: Apologies - it's only because I overlooked it was changed. My bad.

PR #3609, opened 2017-12-13
Rubberduck.smart indenter c#6&7 update

IvenBach:


Rubberduck.SmartIndenter/AbsoluteCodeLine.cs
```diff
@@ -120,58 +120,34 @@ public string Escaped
...
-                       string.IsNullOrEmpty(EndOfLineComment) &&
-                       HasContinuation &&
-                       ((IsDeclarationContinuation && Segments.Count() == 1) || DeclarationRegex.IsMatch(Segments.Last()));
-            }
-        }
+        public bool HasDeclarationContinuation => (!IsProcedureStart && !ProcedureStartIgnoreRegex.IsMatch(_code)) &&
+                                                  !ContainsOnlyComment &&
+                                                  string.IsNullOrEmpty(EndOfLineComment) &&
+                                                  HasContinuation &&
+                                                  ((IsDeclarationContinuation && Segments.Count() == 1) || DeclarationRegex.IsMatch(Segments.Last()));
```
bclothier: This code is too hard to follow for me.

1) I would have the expressions in its own block, like this:
```
public bool HasDeclarationContinuation =>
     <Holy be all end all giant jumbo expression>
```

2) I would have each condition on their own lines. The first line has 2 conditions while subsequent lines has only one conditional. The last line need more handling - either put them in a new indent block to split the `||` condition into their own line.

PR #3615, opened 2017-12-17
COM Release for menu items and command bar

MDoerner:
This PR introduces `Marshal.ReleaseCOMObject` again for the rubberduck command bar and all the `ICommandBarControl`s. The reasoning is that we create them and should dispose of them on shutdown because we attached resources to them.

I opted to introduce a new interface for the relaese and to reimplement the release in the individual base classes in order to be able to make this awaylable selectively. E.g. a component should not be releasable, since we do not own it and because it gets passed around quite a lot.

Compared to my build from repulled `next` today, a lot of access violations vanished. Hower, there are still 29 of them and I also still get 6 exceptions via Winforms.

In addition to the COM release, I tweaked the shutdown code for the `DockableWindowHost` in order to guarantee unsubscription of the event handler on the parent subclassing window.


Rubberduck.VBEEditor/SafeComWrappers/Abstract/IReleasableCOMObject.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Rubberduck.VBEditor.SafeComWrappers.Abstract
+{
+    public interface IReleasableCOMObject
```
bclothier: I think it'd be more consistent to name it `IReleasableComObject`.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
```diff
@@ -164,5 +165,31 @@ public override int GetHashCode()
...
+            {
+                _isReleased = true;
+                return;
+            }
+
+            try
+            {
+                if (final)
+                {
+                    Marshal.FinalReleaseComObject(Target);
```
bclothier: According to the documentation, the return should be `0` if it's successful. I think it's good idea to log (and maybe throw? I doubt) if the return from the `Marshal.FinalReleaseComObject` is non-zero.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
```diff
@@ -164,5 +165,31 @@ public override int GetHashCode()
...
+                }
+                else
+                {
+                    Marshal.ReleaseComObject(Target);
+                }
+            }
+            finally
+            {
+                _isReleased = true;
```
bclothier: I wonder if we really should be setting it to `true` irrespective of the returns from the `Marshal.FinalReleaseComObject` or `Marshal.ReleaseComObject`? If it's meant to ensure that we should not use the object from the .NET side, that might be a case for implementing the `IDisposable` pattern? Otherwise, we're kind of lying because a non-zero return from either `Marshal.FinalReleaseComObject` or `Marshal.ReleaseComObject` doesn't mean the COM object is in fact released.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBarControl.cs
```diff
@@ -209,5 +210,31 @@ public override int GetHashCode()
...
+            {
+                _isReleased = true;
+                return;
+            }
+
+            try
+            {
+                if (final)
+                {
+                    Marshal.FinalReleaseComObject(Target);
```
bclothier: Ditto

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBarControl.cs
```diff
@@ -209,5 +210,31 @@ public override int GetHashCode()
...
+                }
+                else
+                {
+                    Marshal.ReleaseComObject(Target);
+                }
+            }
+            finally
+            {
+                _isReleased = true;
```
bclothier: Ditto

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBarControls.cs
```diff
@@ -70,5 +71,33 @@ public override int GetHashCode()
...
+            {
+                _isReleased = true;
+                return;
+            }
+
+            try
+            {
+                if (final)
+                {
+                    Marshal.FinalReleaseComObject(Target);
```
bclothier: Ditto

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBarControls.cs
```diff
@@ -70,5 +71,33 @@ public override int GetHashCode()
...
+                }
+                else
+                {
+                    Marshal.ReleaseComObject(Target);
+                }
+            }
+            finally
+            {
+                _isReleased = true;
```
bclothier: Ditto

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
MDoerner: To be honest, I just used the commented out code from `SafeComWrapper`.

retailcoder: @bclothier good idea. At *trace* level?

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
retailcoder: @bclothier this may or may not be why reloading RD after unloading bricks the VBE... thoughts?

bclothier:
Well.... I was considering the scenario where the `FinalReleaseComObject` returns a nonzero result... That is IMO, a fatal error because its basically saying we can't release it and probably should not be at this moment. So an Error level seems more apropos there.

That said, we ought to *trace* the non-final calls to `ReleaseComObject`. That would greatly aid us in tracking whether non-final calls allow us to gently put down the referenced COM object in appropriate order. We shouldn't be relying on the `FinalReleaseComObject` to do it for us - that's somehow akin to lobbing a nuclear bomb in your room to clean it.

We need a pattern here.

IMO, `_released` ought be an `int` with default of `null` and take the returns from both `ReleaseComObject` and `FinalReleaseComObject` so that we can then expose a property `bool Released => ( _released == 0 )` which would be more truthful.

I hesitate to suggest but we might consider whether we should require a finalizer so that we can check the `Released`, and if not zero, effect a `FinalReleaseComObject` with a log at *warning* level indicating that we just employed the nuclear option.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
PeterMTaylor: Sounds good for trace level support. More information earlier the better as we can ease off how much input to justify low level COM data we need later down the track.

bclothier: Well.... I was considering the scenario where the `FinalReleaseComObject` returns a nonzero result... That is IMO, a fatal error because its basically saying we can't release it and probably should not be at this moment. So an Error level seems more apropos there.

That said, we ought to *trace* the non-final calls to `ReleaseComObject`. That would greatly aid us in tracking whether non-final calls allow us to gently put down the referenced COM object in appropriate order. We shouldn't be relying on the `FinalReleaseComObject` to do it for us - that's somehow akin to lobbing a nuclear bomb in your room to clean it.

We need a pattern here.

IMO, `_released` ought be an `int` with default of `null` and take the returns from both `ReleaseComObject` and `FinalReleaseComObject` so that we can then expose a property `bool Released => ( _released == 0 )` which would be more truthful.

I hesitate to suggest but we might consider whether we should require a finalizer so that we can check the `Released`, and if not zero, effect a `FinalReleaseComObject` with a log at *warning* level indicating that we just employed the nuclear option.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
bclothier: Hard to say if *that* is the cause of bricking after reloading/unloading. We need more data. But that line definitely isn't truthful.

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
MDoerner: Before I push and, thus, make this comment thread stale, I would like to comment on the idea with the finalizer. I think it would not really help. When this class gets GCed, probably the RCW it wrapps does, too. In that case, there is no use in prolonging the life of the object and the RCW. Moreover, The CLR might hand back the very same RCW to multiple instances of a wrapper class. In that case, we must not final release at the point the wrapper gets finalized.

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
```diff
@@ -1,40 +1,61 @@
...
+                _logger.Warn($"Tried to release a COM wrapper of type {this.GetType()} whose target is not a COM object.");
+                return;
+            }
+
+            try
+            {
+                if (final)
+                {
+                    _rcwReferenceCount = Marshal.FinalReleaseComObject(Target);
+                    _logger.Trace($"Final released COM wrapper of type {this.GetType()} with remaining reference count {_rcwReferenceCount}.");
```
bclothier: I really think that whenever the `_rcwReferenceCount` is non-zero after a `FinalReleaseComObject`, that is an error condition, and we should log at `Error` or at least `Warning` level. If it's zero, then that's expected and we can use `Trace`.

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
```diff
@@ -1,40 +1,61 @@
...
+                }
+                else
+                {
+                    _rcwReferenceCount = Marshal.ReleaseComObject(Target);
+                    _logger.Trace($"Released COM wrapper of type {this.GetType()} with remaining reference count {_rcwReferenceCount}.");
+                }
+            }
+            catch(COMException exception)
+            {
+                _logger.Warn(exception,$"Failed to release COM wrapper of type {this.GetType()}.");
```
bclothier: I think the log should still log the `_rcwReferenceCount` in the `catch` block with a note that it may be the previous value. That may help in correlating whenever there are multiple threads which may cause one of thread to throw?

Rubberduck.VBEEditor/SafeComWrappers/Office.Core/CommandBar.cs
bclothier: @MDoerner, good point regarding not calling the `FinalReleaseComObject` inside a finalizer.  I agree. That said, I still think we should log whenever a class has been finalized without the `_rcwReferenceCount` being zero. Otherwise, the log will not show any missed classes and unless we implement the dispose pattern, the finalizer is only place where we ought test whether its zero and issue a warning when it is not.

PR #3631, opened 2017-12-20
Catch missing resource strings

IvenBach:
`ObsoleteErrorSyntaxInspectionName` previously wasn't found to be missing.

Note: got from @Hosch250's https://github.com/rubberduck-vba/Rubberduck/pull/3621/commits/ae447223291d6fb078ed159ee070cd23a0e3bec2


RubberduckTests/Inspections/GeneralInspectionTests.cs
```diff
@@ -24,13 +24,27 @@ public void InitResources()
             RubberduckUI.Culture = Thread.CurrentThread.CurrentUICulture;
         }
 
+        private static List<Type> GetAllBaseTypes(Type type)
```
Vogel612: this could be somewhat cleaner as an `IEnumerable<Type>` where you'd `yield` the separate results.

PR #3644, opened 2017-12-25
More shutdown tweaks

MDoerner:
This PR contains some more tweaks relating to the shutdown issue.

Following the leads provided by @WaynePhillipsEA I added disposal code for some more com wrappers.
Doing this in the IoC installer allowed to remove the final release code in the app menu disposal. 

In addition, I changed the disposal of the subclasing windows and dockable hosts somewhat to allow them to be collected when we are unloaded via the menu. 

Atm, for me shutdown has a chance of about 50% to have no access violation and of 50% to have one. However, when I comment out the forced GC at shutdown, I get lots of access violations.

### DIFF
--- a/RetailCoder.VBE/Root/RubberduckIoCInstaller.cs
+++ b/RetailCoder.VBE/Root/RubberduckIoCInstaller.cs
@@ -257,9 +257,7 @@ namespace Rubberduck.Root
         private void RegisterRubberduckMenu(IWindsorContainer container)
         {
             const int windowMenuId = 30009;
-            var commandBars = _vbe.CommandBars;
-            var menuBar = commandBars[MenuBar];
-            var controls = menuBar.Controls;
+            var controls = MainCommandBarControls(MenuBar);
             var beforeIndex = FindRubberduckMenuInsertionIndex(controls, windowMenuId);
             var menuItemTypes = RubberduckMenuItems();
             RegisterMenu<RubberduckParentMenu>(container, controls, beforeIndex, menuItemTypes);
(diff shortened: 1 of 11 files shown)


Rubberduck.VBEEditor/SafeComWrappers/VBA/Windows.cs
```diff
@@ -1,5 +1,8 @@
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using Microsoft.Office.Interop.Outlook;
```
bclothier: Is that intentional? I don't see a change that warrant uses of the `Microsoft.Office.Interop.Outlook`?

RetailCoder.VBE/Root/RubberduckIoCInstaller.cs
```diff
@@ -296,24 +294,36 @@ private static int FindRubberduckMenuInsertionIndex(ICommandBarControls controls
...
 
             return controls.Count;
         }
 
+        private ICommandBarControls MainCommandBarControls(int commandBarIndex)
+        {
+            ICommandBarControls controls;
+            using (var commandBars = _vbe.CommandBars)
+            {
+                using (var menuBar = commandBars[commandBarIndex])
```
bclothier: Nit: I believe we can reduce the nesting levels here:
```
using (var commandBars = _vbe.CommandBars)
using (var menuBar = commandBars[commandBarIndex])
{
  ...
}
```

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
```diff
@@ -54,8 +54,17 @@ public void Release(bool final = false)
...
                 {
                     _rcwReferenceCount = Marshal.ReleaseComObject(Target);
-                    _logger.Trace($"Released COM wrapper of type {this.GetType()} with remaining reference count {_rcwReferenceCount}.");
+                    if (_rcwReferenceCount >= 0)
+                    {
+                        _logger.Trace($"Released COM wrapper of type {this.GetType()} with remaining reference count {_rcwReferenceCount}.");
+                    }
+                    else
+                    {
+                        _logger.Warn($"Released COM wrapper of type {this.GetType()} whose underlying RCW has already been released from outside the SafeComWrapper.");
```
bclothier: I think the warning message should still log the new `_rcwReferenceCount` to aid in tracking.

WaynePhillipsEA:
Looking good.   By turning off the explicit GC in the shutdown, I'm detecting these 3 reference leaks:

1) VBProjects event subscription leak, leaving IConnectionPointContainer and IVBProjects references open
2) Parent is leaking in this line... child.Tag = $"{Item.Parent.Name}::{Item.Tag}::{item.GetType().Name}";
3) Parent is also leaking here (i think)...   System.Diagnostics.Debug.WriteLine($"Added handler for: {Parent.Name} '{Target.Caption}' (tag: {Tag}, hashcode:{Target.GetHashCode()})");

Rubberduck.VBEEditor/SafeComWrappers/VBA/Windows.cs
MDoerner: No, these changes are not intentional. I will remove them in my next PR.

ThunderFrame:
Interesting that there were more problems after a reparse. I wonder if there might be other actions that might expose more problems?

- Run unit tests
- Open multiple projects, reparse, close 1 or more projects, reparse
- Reparse, then change the codepane selection.
- Reparse, then switch to a form designer, and select a control
- Use RD to import/export modules
- Use source control

bclothier:
@ThunderFrame,  AIUI, they are just symptoms of the same problem -- we are not cleaning up everywhere we should be. Considering @WaynePhillipsEA's remarks, we need to track all references to all those that derives from the `SafeComWrapper` class and wrap them in an `using` blocks for the local references, in addition to cleanup of non-local references.

ThunderFrame:
@bclothier Absolutely, but @WaynePhillipsEA's tool isn't going to catch problems in features that aren't exercised. Hopefully the abstractions help centralise the focus/fixes, but Unit Testing (and Assert/Fakes) makes use of COM in 2 directions, and *only* if you actually run tests.

WaynePhillipsEA:
@Thunderframe, yes that's a good idea.  I'll do one step at a time, trying to get a consistent full clean shutdown, then try various other RD tasks to try to find anything remaining.  I'm planning on getting the tool out to you guys in the next few days -- I just need to create a little UI which will indicate all live COM references and help identify them at OnDisconnection.

PR #3660, opened 2017-12-31
Wayne Phillip's dockable toolwindow host

MDoerner:
This PR substitutes the `DockableWindowHost` with the one provided by Wayne Phillips provided in issue #3614.

I made some adjustments to it:

1) I fixed the `null` check in the `IntPtr` by making it compare to `IntPtr.Zero` instead of `null`.
2) The disposal of the `ExposedUserControl` now follows the C# disposable pattern correctly.
3) I extracted the user control shutdown and the individual steps into their own private methods.
4) I added a GC call to the user control shutdown to make shutdown after a parse encounter less access violations. (User control shutdown now happens after `OnDisconnect` has completed.) When we finally dispose all `SafeComWrapper`s, we should be able to remove this again.
5) I made the `Release` method remove the `ExposedUserControl`'s child controls in an attempt to stuff our managed memory leak.
6) I moved the helpers, COM interfaces and base wrappers into their own namespace. (Originally I wanted to move them into the `Rubberduck.VBEditor` project, but that would have required to register that for COM interop as well.)


bclothier:
With this PR, I do get this unhandled exception, after loading a VBA project, parsing it then closing Access:

```
Unhandled exception at 0x75A6BB49 (ole32.dll) in MSACCESS.EXE: 0xC0000005: Access violation reading location 0x0922ACBC.
```

However, in spite of the unhandled exception, Access seems to shuts down with exit code `0x0` and I do not have a orphaned process left in the task manager. I still cannot unload & reload in a single session, also.

RetailCoder.VBE/UI/DockableWindowHost.cs
```diff
@@ -7,86 +7,87 @@
...
-            [FieldOffset(0)]
-            public uint Value;
-            [FieldOffset(0)]
-            public readonly ushort LowWord;
-            [FieldOffset(2)]
-            public readonly ushort HighWord;
-        }
+    // ExposedUserControl - wrapper for UserControl that also exposes the underlying 
+    // IOleObject and IOleInPlaceObject COM interfaces implemented by it
+    public class ExposedUserControl : UserControl
```
bclothier: I think this should be its own file.

RetailCoder.VBE/UI/DockableWindowHost.cs
```diff
@@ -130,37 +131,346 @@ protected override bool ProcessKeyPreview(ref Message m)
...
+            // Moreover, we can release the UserControl COM references, as Close() should be the very last call into the IOleObject interface.
+            PerformUserControlShutdown();
+
+            return hr;
+        }
 
-            Debug.WriteLine("DockableWindowHost release called.");
+        private void PerformUserControlShutdown()
+        {
+            ReleaseCOMReferenceOfSctiveXControl();
```
bclothier: Typo here - Also I think the naming should be `ReleaseComReferenceOfActiveXControl` --- it could be also shortened to `ReleaseActiveXControlReference`, possibly.

WaynePhillipsEA:
@bclothier, I am also getting that exception, on occasion.  If you look at the beginning of the callstack you'll see it is within CorExitProcess, which is when the CLR shutting down.  I'm only getting this exception when there are reported leaks in my logs telling me that some of the objects mentioned earlier are not being explicitly released still.  When the logs show a clean shutdown (ie when GC happens to work well)  there are no exceptions at all.  Therefore I'm quietly confident that once we are explicitly releasing everything perfectly (and Max is making great progress with that), these odd exceptions will be gone.  Nearly there folks

RetailCoder.VBE/UI/DockableWindowHost.cs
MDoerner: I would prefer to leave this in the same file, but at the bottom, until anything but the dockable host really needs an exposed user control. (I cannot really see what that would be.) This would follow the same reasoning as the `ParentWindow` class, which is also defined in this file.

PR #3663, opened 2018-01-01
Removal of TimerHook and update Kernel32/User32/HotKey

### LINKED ISSUE #3662: TimerHook Windows API issues
A few issues identified in TimerHook.cs;

1) the timerId generated is never stored back to the class _timerId field.  Therefore, in Detach(), a zero timerId is passed to KillTimer() and GlobalDeleteAtom().  

Subsequently, your timer hook never gets released, and so this can cause the WinAPI to try calling into your managed callback even after the CLR has been unloaded.  The callback will be a released memory address after the CLR unload, and so a potential crash situation.

2) GlobalAddAtom can fail with zero return value, but this is unchecked.
3) SetTimer can fail with a zero return value, but this is unchecked.

Easy to fix :)  Happy new year

bclothier:
Closes #3662; add explicit logging for errors arising from Win32 API which will not necessarily throw exceptions.


RetailCoder.VBE/Common/TimerHook.cs
```diff
@@ -31,8 +33,27 @@ public void Attach()
             try
             {
                 var timerId = (IntPtr)Kernel32.GlobalAddAtom(Guid.NewGuid().ToString());
```
bclothier: So I'm reading the documentation on the [SetTimer](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644906(v=vs.85).aspx) and there is no mention of needing a global atom. I'm not sure what purpose the global atom is supposed to serve here, given that we are generating a new GUID every time we instantiate the class which will never match any existing entries in the global atom table. Can anyone explain why we need this? Otherwise, this strikes me as a cargo-cult programming...

RetailCoder.VBE/Common/TimerHook.cs
```diff
@@ -1,14 +1,16 @@
 using System;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using Rubberduck.Common.WinAPI;
+using NLog;
 
 namespace Rubberduck.Common
 {
     public class TimerHook : IAttachable, IDisposable
```
bclothier: Is this class even in use? It is referenced in the `RubberduckIocInstaller` but by time RD loads, the class is never instantiated nor its methods called....

RetailCoder.VBE/Common/TimerHook.cs
WaynePhillipsEA: +1, although you need to supply a timer ID of some sort.  The relevant part of the documentation says "If the hWnd parameter is not NULL and the window specified by hWnd already has a timer with the value nIDEvent, then the existing timer is replaced by the new timer".  The rest of it (e.g. saying that a new timer ID will be allocated for you) refers to where hWnd is being passed a NULL, so doesn't seem to apply.  

Using a global atom is clearly not a requirement though.  You just need to be prepared for your specified ID to be already assigned to a timer (e.g. by another addin, or the VBE itself), and try another if it fails (which you should do even with the global atom anyway).

RetailCoder.VBE/Common/TimerHook.cs
WaynePhillipsEA: ... indeed, that does appear to be the case!   I love wasting my time ;-)

MDoerner: I think it is a remenant of the old timed logger interceptor, which has been removed because there has been the feeling that the inability to proply use reflection on the intercepted inspections (to get their type) was too much of a problem.

bclothier: Ok, if there's no objection, I will revise the PR to delete the file altogether and remove it from the IoC installer.

Vogel612: sounds good. If it's never used, it should be deleted 👍

Issue #3050, opened 2017-06-05
Inspection for inappropriate Type comparisons in Select Case Blocks


ThunderFrame:
The following block has a Boolean expression in the `Select Case` statement, but the `Case` statements are using non-Boolean types:

```vb
  Select Case VBA.Rnd > 0.5
    Case -5 'Implicitly True
      Stop
    Case -4 'Unreachable
      Stop
    Case Else
      Stop
  End Select
```

Rubberduck should have an inspection that identifies the type mismatches, and the implied casts that they entail.

In the case above, the fix could be either of the following (where my preference would be for the first):

```vb
  Select Case VBA.Rnd
    Case > 0.5
      Stop
    Case Else
      Stop
  End Select
```

or

```vb
  Select Case VBA.Rnd > 0.5
    Case True
      Stop
    Case Else
      Stop
  End Select
```

The Variant comparison bug #3047 can show up here too.

ThunderFrame:
references #1584

bclothier:
We can consider this closed by PR #3868 - additional issues with the inspection should be opened separately.

Issue #1584, opened 2016-05-25
Code inspection for unreachable Case expressions and statements


ThunderFrame:
> If testexpression matches any Case expressionlist clause, the statements following that Case statement run up to the next Case, Case Else, or End Select statement. Control then passes to the statement following End Select. If testexpression matches an expressionlist clause in more than one Case clause, only the statements following the first match run.

In large Select Case blocks, it's all too easy to inadvertently add Case statements that have intersections with prior Case blocks. (I found one in production code today)

That means that these blocks have unreachable case statements:

``` vb
Const x As String ="Bar"
Select Case x
  Case "Foo", "Bar"
    'Do FooBar
  Case "Foo"
    'Unreachable
  Case "Bar"
    'Unreachable
  Case "Foo"
    'Unreachable
End Select
```

``` vb
Const x as Long = 7

Select Case x
  Case 1 to 100
    'Do FooBar
  Case 50
    'Unreachable
End Select
```

We need to handle the special case of `Select Case True`

``` vb
Sub foo()
  Dim x As Long: x = InputBox("Enter a value for x")
  Dim y As Long: y = InputBox("Enter a value for y")
  Select Case True
    Case x = 8
      Stop
    Case y = 9
      Stop
    Case Else
      Stop
  End Select
End Sub
```

comintern:
The `Select Case True` function highlights why parsing P-Code or compiled code helps to identify dead code:

```vba
Private Sub Foo()
    Select Case True
        Case Bar()
            Debug.Print "Bar"
        Case Baz()
            'Heuristically unreachable.
            Debug.Print "Baz"
    End Select
End Sub

Private Function Bar() As Boolean
    'Should trigger an inspection that it always returns true.
    Bar = True
End Function

Private Function Baz() As Boolean
    'Should trigger an inspection that it always returns false.
    Baz = False
End Function
```

ThunderFrame:
Some Case handlers will be unreachable by virtue of their type. Here's a Boolean condition in the `Select Case` statement.

```vb
  Select Case VBA.Rnd > 0.5
    Case True
      Stop
    Case False
      Stop
    Case Else 'Unreachable
      Stop
  End Select
```

While some Case handlers will be unreachable by virtue of implicit type casting:
```vb
  Select Case VBA.Rnd > 0.5
    Case -5 'Implicitly True
      Stop
    Case -4 'Unreachable
      Stop
    Case Else
      Stop
  End Select
```

[...]

PR #3664, opened 2018-01-02
Unreachable Select Case Inspection

BZngr:
Related to #1584 and #3050 

1. Detects unreachable Cases Clauses caused by duplicate, invalid comparison-types,  or prior super-set conditions.
2. Detects unreachable Case Else clauses where all possible cases are previously handled.
3. Clauses that contain invalid comparison types are not flagged if there are other valid comparison types present within the same Case Clause.
4. Attempts to makes comparisons of mixed types that would be implicitly converted and evaluated by VBA.
5. Supports evaluation of enums.  Enums Select Case statements will not mark the Case Else clause as unreachable if all enum values are handled.  Enums will be inspected in the same manner as Longs.  Case Else will be flagged as unreachable if and only if all Long values are handled by prior Case Clauses.
6. Will pass on inspecting Select Case statements where the Select Case statement includes math on the Select Case variable (e.g.,` Select Case x ^ 2 / 4`).


Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+
+        public override IInspectionListener Listener { get; } =
+            new UnreachableCaseInspectionListener();
+
+        public override CodeInspectionType InspectionType => CodeInspectionType.CodeQualityIssues;
+
+        private static Dictionary<string, string> CompareInversions = new Dictionary<string, string>()
+        {
+            { CompareTokens.EQ, CompareTokens.EQ },
+            { CompareTokens.NEQ, CompareTokens.NEQ },
```
Vogel612: why is NEQ the inverse of NEQ and EQ the inverse of EQ? Shouldn't it be:
```
{ CompareTokens.EQ, CompareTokens.NEQ },
{ CompareTokens.NEQ, CompareTokens.EQ },
```

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+            { CompareTokens.LT, CompareTokens.GT },
+            { CompareTokens.LTE, CompareTokens.GTE },
```
Vogel612: Truly inverse comparisons would be `{ LT, GTE }, { LTE, GT }`

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+            }
+
+            var expression = ParserRuleContextHelper.GetDescendent<SimpleNameExprContext>(firstLExpr).GetText();
+            if (SymbolList.TypeHintToTypeName.ContainsKey(expression.Last().ToString()))
+            {
+                return SetTheTypeNames(selectStmtDO, SymbolList.TypeHintToTypeName[expression.Last().ToString()]);
+            }
+
+            var idRefs = (State.DeclarationFinder.MatchName(expression).Select(dec => dec.References))
+                .SelectMany(rf => rf).Where(idr => ParserRuleContextHelper.HasParent(idr.Context, selectStmtDO.SelectExpressionContext));
```
Vogel612: ParserRuleContextHelper seems like it should be extension methods... `idr.Context.HasParent(selectStmtDO.SelectExpressionContext)` would be a little cleaner

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+                if (caseClause.ResultType == ClauseEvaluationResult.NoResult)
+                {
+                    caseClause = InspectCaseClause(caseClause, selectStmtDO.SummaryCaseClauses);
+                    selectStmtDO = UpdateSummaryClauses(selectStmtDO, caseClause);
+
+                    if (caseClause.RangeClauseDOs.All(rg => rg.ResultType != ClauseEvaluationResult.NoResult))
+                    {
+                        caseClause.ResultType = caseClause.RangeClauseDOs.All(rg => rg.ResultType == ClauseEvaluationResult.Unreachable)
+                            ? ClauseEvaluationResult.Unreachable : caseClause.RangeClauseDOs.All(rg => rg.ResultType == ClauseEvaluationResult.MismatchType)
+                                ? ClauseEvaluationResult.MismatchType : ClauseEvaluationResult.NoResult;
```
Vogel612: Could this be more explicit and less ternaries, please? I find it hard to follow this

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+            }
+            return caseClauseDO;
+        }
+
+        private RangeClauseDataObject InitializeRangeClauseDataObject(RangeClauseDataObject rangeClauseDO, string targetTypeName, string refName )
+        {
+            rangeClauseDO.TypeNameTarget = targetTypeName;
+            rangeClauseDO.IdReferenceName = refName;
+            rangeClauseDO.UsesIsClause = HasChildToken(rangeClauseDO.Context, Tokens.Is);
+            rangeClauseDO.IsValueRange = HasChildToken(rangeClauseDO.Context, Tokens.To);
```
Vogel612: HasChildToken should also be an extension method

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+                    if (!parentData.EvaluateAsIsClause)
+                    {
+                        parentData.EvaluateAsIsClause = CompareOperations.Keys.Contains(child.GetText());
+                    }
+                    parentData.Operator = child.GetText();
+                    contextEvals = AddEvaluationData(contextEvals, parentContext, parentData);
+                    continue;
+                }
+
+                if (IsBinaryOperatorContext(child) || IsUnaryOperandContext( child) )
```
Vogel612: extra space after opening paren

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+                        SelectCaseRefName = valueDeclaration.IdentifierName
+                    };
+
+                    parentData = AddEvaluationData(parentData, child, exprEval);
+                    return ResolveContextValue(parentData, child).First().Value.Result.AsString();
+                }
+
+                if (child is LiteralExprContext)
+                {
+                    if(child.Parent is EnumerationStmt_ConstantContext)
```
Vogel612: missing space before opening paren

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+                }
+            }
+            return string.Empty;
+        }
+
+        private List<ParserRuleContext> GetRHSContexts(List<IParseTree> contexts)
+        {
+            var contextsOfInterest = new List<ParserRuleContext>();
+            var eqIndex = contexts.FindIndex(ch => ch.GetText().Equals(CompareTokens.EQ));
+            if(eqIndex == contexts.Count)
```
Vogel612: missing space before paren

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+            else
+            {
+                summaryClauses.SingleValues.Add(rangeClauseDO.SingleValue);
+            }
+            return summaryClauses;
+        }
+
+        private bool EvaluateCaseElseAccessibility(SummaryCaseCoverage summaryClauses, string typeName)
+        {
+            if(summaryClauses.CaseElseIsUnreachable) { return summaryClauses.CaseElseIsUnreachable; }
```
Vogel612: missing opening paren

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
```diff
@@ -0,0 +1,1309 @@
...
+            if (summaryClauses.IsLTMax != null && summaryClauses.IsGTMin != null)
+            {
+                if (summaryClauses.IsLTMax > summaryClauses.IsGTMin
+                        || (summaryClauses.IsLTMax >= summaryClauses.IsGTMin
+                                && summaryClauses.SingleValues.Contains(summaryClauses.IsLTMax)))
+                {
+                    return true;
+                }
+
+                else if(summaryClauses.Ranges.Count > 0)
```
Vogel612: missing space before opening paren

Rubberduck.Inspections/Concrete/UnreachableCaseInspectionValue.cs
```diff
@@ -0,0 +1,619 @@
...
+            return thisValue._operatorIsLT(thisValue, compValue);
+        }
+
+        public static bool operator ==(UnreachableCaseInspectionValue thisValue, UnreachableCaseInspectionValue compValue)
+        {
+            if(ReferenceEquals(null, thisValue))
+            {
+                return ReferenceEquals(null, compValue);
+            }
+            else
```
Vogel612: opening the else block here is not necessary :)

Rubberduck.Inspections/Concrete/UnreachableCaseInspectionValue.cs
```diff
@@ -0,0 +1,619 @@
...
+            }
+        }
+
+        public static bool operator !=(UnreachableCaseInspectionValue thisValue, UnreachableCaseInspectionValue compValue)
+        {
+            if (ReferenceEquals(null, thisValue))
+            {
+                return !ReferenceEquals(null, compValue);
+            }
+            else
```
Vogel612: opening the else block here is not necessary

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Yeah...I think a better name for the dictionary might be "AlgebraicLogicInversions".  The goal is to be able to perform some (very) limited algebra for logical Range Clause expressions like '6 < x'.  In order to move 'x' to the LHS, the desired results is x > 6.  So, for '6 = x' the algebra results desired is 'x = 6'.  Same for NEQ.  I'm definitely open to a better name.  Note: I am interested in moving the variable to the LHS so that the expression can be evaluated in the same way as an IS clause. ('x < 6' == 'Is < 6' when 'x' is the Select Case expression)

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: See comment above - the goal is algebraic inversion rather than logical inversion.

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
Hosch250: Pretty sure that is an extension method already. If it isn't, Vogel's right.

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
Hosch250: Using C# 6, I'm pretty sure you could format these as `[CompareTokens.EQ] = CompareTokens.EQ`, etc.

BZngr: @Hosch250 yes!  Updated this and other Dictionaries accordingly

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Added extension method(s)

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Modified...yes...more readable.

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Added extension method(s).  If the intent is to refactor ParserRuleContextHelper completely into extension methods, then I would like to make that a separate Issue/PR.  I left ParserRuleContextHelper as-is.

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Cleaned-up (and others like it)

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Cleaned-up (and others like it)

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Cleaned-up (and others like it)

Rubberduck.Inspections/Concrete/UnreachableCaseInspection.cs
BZngr: Cleaned-up (and others like it)

Rubberduck.Inspections/Concrete/UnreachableCaseInspectionValue.cs
BZngr: Cleaned-up (and others like it)

bclothier:
@BZngr in regard to this point:

2. Detects unreachable Case Else clauses where all possible cases are previously handled.

I'm just curious - even if one had a function that takes an `Enum` as the parameter and the `Select Case` within the function enumerated all the values for that `Enum`, it is still valid to do this:

`MyFunction(1234) 'Look, a random number!`

which is implicitly converted but still won't match any of the `Case` statement. 

I often add a guard statement similar to this:

```
   Case Else
     Err.Raise 5, "MyFunction", "Invalid value given for the enumeration."
```

Will the inspection flag that as unreachable?

BZngr:
Updated to remove strict enum evaluations for unreachable Case and CaseElse clauses.  Enums are now evaluated in the same manner as Longs.

BZngr:
I'm closing this pull request and plan resubmit again later.  Even though it finds many scenarios of unreachable code, I believe it could use more refactoring/rework to make it easier to understand and review.

And a question:
It seems that most (if not all) other inspections seem  to 'fit' into a single inspection class/file.  So, I tried to do the same with this inspection - and still ended up with two classes/files.  To reduce the file/class count, the solution has several supporting Structs and processes them.  But, I think there may be read/review-ability benefits to refactoring some of the Structs into classes.  I'm thinking that the additional files should have their own folder (a subfolder of Inspections\Concrete)  to organize the supporting classes.  Since I do not see this organization/approach for the other existing inspections, I wanted to check if this was an acceptable way to organize my code.  Any guidance here would be appreciated - thanks.

bclothier:
RE: the question - FWIW if you look at Extract Method, there are several files, way more than other refactoring normally needs. 

I wouldn't be surprised if we have other inspections where fitting everything in a one file don't make sense. I think that as long we follow the same convention used by refactoring -- creating a folder instead of a file, it should be OK to do that for inspections, too.

wqweto:
@BZngr FYI, following is a valid VB code that does not work as intended:
```
Select Case x
  Case x ^ 2 = 49
    'OK
  Case x = 7
    'Unreachable, but not detected - math/algebra on the Select Case variable yet to be supported
  Case Is < 50
    ...
End Select
```
The expressions in first two `Case` statements are boolean so it makes sence to only `Select Case True` or `False`.

Basicly `Case x < 50` is different than `Case Is < 50`  for a `Select Case x` -- first one is probably wrong.

Noticed you are using similar in `UnreachableCaseInspectionTests`

bclothier:
@wqweto  I don't know if this is appropriate for this inspection. This is definitely something we need to inspect but this strikes me as more about inspecting for implicit type conversion than for unreachable code because VBA allow us to do implicit voodoo conversions where we can write expressions that results in a boolean which is then converted to whatever `x` is (presumably an integer of some size), which would cause unexpected results as you said.  Furthermore this problem would not be limited to only the `Select Case` so for that reason, I think it's best as its own inspection, apart from this PR. I suggest that you open an issue if there's not an inspection already for that type of implicit type conversions in a comparison.

bclothier:
BTW, see #3050

PR #3669, opened 2018-01-03
Remove duplicated resx file

Vogel612:


Vogel612:
got obsoletified somehow... nuked

PR #3670, opened 2018-01-04
Introducing the COM safe

MDoerner:
This PR introduces the COM safe, a cache for all `ISafeComWrapper`s that disposes them all as part of its own disposal.

I went with an implementation as a singleton with a static factory because this seemed to be the least invasive method to attach it to all `SafeComWrappers`. This way, only the base class needs to know of the existence of the cache. Although the cache can be accessed fro the static method on the factory I saved a reference to it on the wrappers in order to ensure that, even if the cache is reset, the wrappers try to remove themselves from the correct cache on disposal.

Apart from introducing the Com safe, I removed the `GC.Collect` call at the end of the shutdown code. I get a shutdown without access violations with this if I do nothing before closing. If I parse oin my test project, I get exactly one access violation.

With the GC call in place, I occasionally get Expections for _ race on RCWCleanup_.

The downside of releasing all `SafeComWrappers` is that for some reason, we now release RCWs that already got released from somewhere else. I have no idea from whare that could be since we hold a reference to all COM objects we requested. 

This PR should not be merged until PR #3660 got merged so that it can be merged into this PR.


Rubberduck.VBEEditor/ComManagement/ComSafe.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class ComSafe: IComSafe
+    {
+        //We override the equality comparison and hash code because subclasses of SafeComWrapper<T> override the corresponding methods.
+        //We need to distinguish between the individual wrapper instances no matter whether they are semantically equal.
+        private readonly ConcurrentDictionary<ISafeComWrapper, byte> _comWrapperCache = new ConcurrentDictionary<ISafeComWrapper, byte>(new ReferenceEqualityComparer());
+
+
+        public void Add(ISafeComWrapper comWrapper)
+        {
+            _comWrapperCache.AddOrUpdate(comWrapper, key => 1, (key, value) => value);
```
bclothier: TBH, this is kind of confusing because we are using `comWrapper` as the key, `key` (which is always `1` as the value. If you do not want to use the hash code for the object as its key, I think a comment is needed here to make clear that we do mean to use the `comWrapper` as the key, not as the value because we don't have a `ConcurrentList<T>` here.

Rubberduck.VBEEditor/ComManagement/ComSafe.cs
```diff
@@ -0,0 +1,57 @@
...
+        private static Lazy<ComSafe> _comSafe = new Lazy<ComSafe>();
+
+        public static IComSafe GetCurrentComSafe()
+        {
+            return _comSafe.Value;
+        }
+
+        public static void ResetComSafe()
+        {
+            _comSafe = new Lazy<ComSafe>();
```
bclothier: Do we need to attempt to dispose the old `_comSafe` and its content before we reset it to a new instance?

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
```diff
@@ -119,20 +124,25 @@ public void Dispose()
...
+                if (_isDisposed)
+                {
+                    return;
+                }
+                _isDisposed = true;
             }
 
             if (disposing && !HasBeenReleased)
             {
                 Release();
```
bclothier: Hmm. We take a lock on the `_isDisposed` to avoid race conditions, yet we don't do the same for the `_rcwReferenceCount`? The condition here tests `disposing` which isn't affected by the `_isDisposed`, so it seems to me we still have a race condition on the `Release()`?

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
```diff
@@ -111,34 +111,36 @@ public void ImportSourceFile(string path)
...
-                component.CodeModule.Clear();
-
                 var codeString = File.ReadAllText(path, Encoding.UTF8);
-                component.CodeModule.AddFromString(codeString);
+                using (var codeModule = component.CodeModule)
+                {
+                    codeModule.Clear();
+                    codeModule.AddFromString(codeString);
+                }
+                component.Dispose();
```
bclothier: If you can't do a `using` on the `component`, it should be at least in the `finally` block to ensure that exceptions between the successful assignment of `component` up to the `Dispose()` will not leak the variable.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
```diff
@@ -147,18 +149,25 @@ public void ImportSourceFile(string path)
...
                 var correctCodeString = string.Join(Environment.NewLine, codeLines.Skip(declarationsStartLine - 1).ToArray());
 
-                component.CodeModule.Clear();
-                component.CodeModule.AddFromString(correctCodeString);
+                using (var codeModule = component.CodeModule)
+                {
+                    codeModule.Clear();
+                    codeModule.AddFromString(correctCodeString);
+                }
+                component.Dispose();
```
bclothier: Same comment as above.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
```diff
@@ -147,18 +149,25 @@ public void ImportSourceFile(string path)
...
             }
             else if (ext != ComponentTypeExtensions.FormBinaryExtension)
             {
-                Import(path);
+                using(Import(path)){} //Nothing to do here, except properly disposing the wrapper returned from Import.
```
bclothier: This code smells. If `Import(path)` is itself atomic, then it should clean itself up, not spill over to the caller, no?

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
```diff
@@ -243,10 +253,21 @@ private static void OnComponentRenamed(VB.VBComponent vbComponent, string oldNam
...
                 {
-                    handler.Invoke(component, new ComponentRenamedEventArgs(project.ProjectId, project, new VBComponent(vbComponent), oldName));
+                    handler.Invoke(component,
+                        new ComponentRenamedEventArgs(project.ProjectId, project, new VBComponent(vbComponent),
+                            oldName));
+                }
+                else
+                {
+                    project.Dispose();
+                    component.Dispose();
```
bclothier: Does those `Dispose()` really belong in the `else`? I think they should be unconditional? Maybe wrap in a `try/finally` to guarantee cleanup?

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
```diff
@@ -281,11 +302,20 @@ private static void OnDispatch(EventHandler<ComponentEventArgs> dispatched, VB.V
...
+                    project = components.Parent;
+                }
                 if (project.Protection != ProjectProtection.Locked)
                 {
                     handler.Invoke(component, new ComponentEventArgs(project.ProjectId, project, component));
                 }
+                else
+                {
+                    project.Dispose();
+                    component.Dispose();
```
bclothier: Same concern.

Rubberduck.VBEEditor/ComManagement/ComSafe.cs
MDoerner: This is the standard way to implement a `ConcurrentHashSet`. The actual key in the dictionary is always the result of `equalityComparer.GetHash(key)`.

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
MDoerner: The lock guarantees that only ever one invocation of `Dispose` can proceed beyond the if statement. All others will return because `_isDisposed` is already true.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
MDoerner: `Import` returns the component it has imported. It cannot dispose that component as it would potentially make it unusable by the caller. It is just that the calling method here is only interested in the sidebar effects.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
MDoerner: They belong there because the handlers of the events should receive an indisposed component in the event arguments. Otherwise the event arguments may be unusable. However, in case we do not handle the event at all, we can dispose immediately and this is the only place that has a reference, except the COM safe now.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
MDoerner: As above.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
MDoerner: Can do that.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponents.cs
MDoerner: As above.

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
bclothier: I follow now. That said, I wonder if `_comSafe.TryRemove(this)` should be inside the `Remove()`? Also, how will we deal with exceptions that are risen in the section after the `lock` block? Because if the first call results in an exception, the `_isDisposed` is already set, so second and subsequent calls would fail to release this?

MDoerner: I do not see how the thread passing beyond the if statement could throw an exception before it reaches the `Release` method, which itself can really only throw on the COM release.

bclothier:
I should observe that with this pull, in addition to already mentioned issue about releasing too much the below. I am not sure if they are directly relevant to this PR or side issues. 

1) we are apparently leaking atoms....

`2018-01-04 20:18:47.0897;WARN;Rubberduck.Common.Hotkeys.Hotkey;Error calling UnregisterHotKey; the error was 1419; going to delete the atom anyway... The memory may leak.;
2018-01-04 20:18:47.0897;WARN;Rubberduck.Common.Hotkeys.Hotkey;Error calling DeleteGlobalAtom; the error was 6 and the id was 49379.;`

(See merged PR #3663 for details)

[1419](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681385%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) means the hotkey wasn't registered, and [6](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx) means the handler wasn't registered. 

2) we are getting various `MissingMethodException`:

`Exception thrown: 'System.MissingMethodException' in mscorlib.dll
No parameterless constructor defined for this object.` This occurs in
`var proto = Convert.ChangeType(Activator.CreateInstance(inspectionType), inspectionType);`

from `CodeInspectionSettings.cs`, in `GetSetting` method. 

3) The `Quack!` message now seems odd since it's no longer the last message:

```
2018-01-04 20:21:28.0597;INFO;Rubberduck._Extension;No exceptions were thrown.;
2018-01-04 20:21:28.0597;TRACE;Rubberduck._Extension;Unregistering AppDomain handlers....;
2018-01-04 20:21:28.0597;TRACE;Rubberduck._Extension;Done. Shutdown completed. Quack!;
2018-01-04 20:21:28.5357;TRACE;Rubberduck.UI._DockableWindowHost;IOleInPlaceObject::InPlaceDeactivate() called;
2018-01-04 20:21:28.5447;TRACE;Rubberduck.UI.CustomComWrappers.Wrapper_IOleClientSite;IOleInPlaceSite::OnInPlaceDeactivate() called;
2018-01-04 20:21:28.5447;TRACE;Rubberduck.UI._DockableWindowHost;IOleObject::DoVerb() called;
```

I think this is correct as Wayne did indicate that those `IOle**` shutdowns will occur  a little after so updating the `Shutdown` message might suffice.

MDoerner:
@bclothier I think point one and two are other issues. Maybe open actual issues for them.

Point three should be good now.

MDoerner:
Status update: 

- No more excess releases of RCWs.
- Occasionally, there is a random RaceOnRCW reported. (Trying to release an RCW while it is in use.) This seems to be connected to #3675. (I had one instance of it when I also got COM NREs for two commands, which both should not have been executing anyhting anymore. Instead, they called `FindControlByTag` on AppCommandBarBase`.)
- GC calls have been removed completely.
- One thread is exiting with code 1.
- There is still one exception crossing the managed/native boundary for each dockable window, (`DoVerb` with `iVerb = -6`) These get handled/swallowed by the Winforms controls.

bclothier:
For documentation ---

[IOleObject::DoVerb](https://msdn.microsoft.com/en-us/library/windows/desktop/ms694508%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) indicates that negative verbs are likely one of 7 common verbs -- the -6 represents OLEIVERB_DISCARDUNDOSTATE .

MDoerner:
No more exceptions on disconnect and shutdown.

MDoerner:
The thread exiting with code 1 can also be observed without RD loaded.

MDoerner:
I do not plan to add anything more to this PR. So feel free to review and point out anything that strikes you as odd.

Issue #3676, opened 2018-01-06
Hotkey leak


bclothier:
It was originally observed in PR #3670 and verified in the current `next` branch that we seem to be leaking hotkeys. Note that we recently modified the hotkey to be more compliant with Win32 API documentation in the PR #3663. 

We register hotkeys when we get a focus message and unregister them when we get a kill focus message via the subclassing of the tool windows. 

However, investigation shows that they are in fact registered, yet we are getting told that they weren't registered. Here's the debug output:

Note: error [#1491 is ERROR_HOTKEY_NOT_REGISTERED](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681385%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) and [#6 is ERROR_INVALID_HANDLE](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx).
```
Added handler for: Custom Popup 112045643 'Go to &implementation...' (tag: Project Window::RubberduckMenu::FindAllImplementationsCommandMenuItem, hashcode:16367560)
2018-01-06 10:32:57.6205;TRACE;Rubberduck.Common.Hotkeys.Hotkey;Hotkey for the associated command Rubberduck.UI.Command.Refactorings.CodePaneRefactorRenameCommand was registered with id 49265;
2018-01-06 10:32:57.6205;TRACE;Rubberduck.Common.Hotkeys.Hotkey;Hotkey for the associated command Rubberduck.UI.Command.Refactorings.RefactorEncapsulateFieldCommand was registered with id 49284;
2018-01-06 10:32:57.6205;TRACE;Rubberduck.Common.Hotkeys.Hotkey;Hotkey for the associated command Rubberduck.UI.Command.Refactorings.RefactorExtractMethodCommand was registered with id 49269;
2018-01-06 10:32:57.6205;TRACE;Rubberduck.Common.Hotkeys.Hotkey;Hotkey for the associated command Rubberduck.UI.Command.Refactorings.RefactorMoveCloserToUsageCommand was registered with id 49273;
2018-01-06 10:32:57.6205;TRACE;Rubberduck.Common.Hotkeys.Hotkey;Hotkey for the associated command Rubberduck.UI.Command.CodeExplorerCommand was registered with id 49274;
[...]

Issue #3677, opened 2018-01-06
Inspection Settings throws an exception


bclothier:
We are getting an exception in `CodeInspectionSettings.cs`, on method `GetSetting` on this line:

```
var proto = Convert.ChangeType(Activator.CreateInstance(inspectionType), inspectionType);
```

The exception is:
```
System.MissingMethodException occurred
  HResult=0x80131513
  Message=No parameterless constructor defined for this object.
  Source=mscorlib
  StackTrace:
   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean noCheck, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor, Boolean& bNeedSecurityCheck)
   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at Rubberduck.Settings.CodeInspectionSettings.GetSetting(Type inspectionType) in C:\GitHub\Rubberduck\Rubberduck\RetailCoder.VBE\Settings\CodeInspectionSettings.cs:line 55
```

In this particular case, the `MissingAnnotationArgumentInspection` inspection has this constructor:

```
public MissingAnnotationArgumentInspection(RubberduckParserState state)
            : base(state, CodeInspectionSeverity.Error) { }
```

Thus, we must pass in the state *and* the severity with the above. Those are not available in the calling code.

The exception is caught then swallowed by returning `null` for the settings. Thus, it isn't directly observed when running RD normally, however, it trips up the debugger which must be Shift+F11 to force it into the `catch` block.

I believe this may be a side-effect of PR #3534 which changed how we stored the settings.

bclothier:
Addendum - those inspections are what throws exceptions:

```
MissingAnnotationArgumentInspection
MissingAttributeInspection
BooleanAssignedInIfElseInspection
EmptyStringLiteralInspection
IllegalAnnotationInspection
ImplicitByRefModifierInspection
ModuleScopeDimKeywordInspection
MultilineParameterInspection
MultipleDeclarationsInspection
[...]

PR #3678, opened 2018-01-06
Minor tweaks

bclothier:
Add a missing `return` in a fail case for the `Hotkey.cs` and add `null` for RPS for the code inspection and also fix the `InspectionHelper` to provide name for the inspection so that it can locate in the settings correctly.


bclothier:
Note: the 3rd commit closes #3676 and 1st commit is a hack for #3677 (which requires a proper resolution one way or other).

PR #3682, opened 2018-01-07
Sync ExtractMethod branch w/ next

bclothier:
To get all other commits out of way


bclothier:
After discussing w/ @Vogel612  this isn't a solution - the `ExtractMethod` needs to be rebased instead.

PR #3683, opened 2018-01-08
Wrapping the designer

MDoerner:
This PR introduced a wrapper for the `UserForm` class in order to properly release the designer when querying the controls of a component. 

The only methods on the wrapper are those needed to get the controls. If ever the need arises to use the other methods, the interface and the implementation will have to be extended.

In addition to the wrapper, this PR contains some disposal of temporary instances of `ICodeModule`. This should cover all instances outside the main project and the tests.

### DIFF
--- a/RetailCoder.VBE/UI/CodeExplorer/Commands/AddComponentCommand.cs
+++ b/RetailCoder.VBE/UI/CodeExplorer/Commands/AddComponentCommand.cs
@@ -32,7 +32,7 @@ namespace Rubberduck.UI.CodeExplorer.Commands
         {
             using (var components = node != null
                 ? GetDeclaration(node).Project.VBComponents
-                : ComponentsColectionFromActiveProject())
+                : ComponentsCollectionFromActiveProject())
             {
                 var folderAnnotation = $"'@Folder(\"{GetFolder(node)}\")";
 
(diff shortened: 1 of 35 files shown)


RetailCoder.VBE/UI/CodeExplorer/Commands/AddComponentCommand.cs
```diff
@@ -30,14 +30,28 @@ public bool CanAddComponent(CodeExplorerItemViewModel parameter)
 
         public void AddComponent(CodeExplorerItemViewModel node, ComponentType type)
         {
-            var components = node != null
+            using (var components = node != null
                 ? GetDeclaration(node).Project.VBComponents
-                : _vbe.ActiveVBProject.VBComponents;
+                : ComponentsColectionFromActiveProject())
```
bclothier: Typo - probably should be `ComponentsCollectionFromActiveProject`

MDoerner:
Generally, the anything that is not a COM wrapper carries managed references to a wrapper and does not return a new one. Consequently, only wrappers coming from another wrapper should be disposed; the other ones might still be in use by someone else.

MDoerner:
Regarding the second point, outside of `Rubberduck. VBEditor` we only use wrapped COM objects, with the exception of the dockable window wrapper.

RetailCoder.VBE/UI/CodeExplorer/Commands/AddComponentCommand.cs
MDoerner: No typo; it is the Com collection `VBComponents` from the active project. 
This is only a separate function, because you cannot have anonymous functions in a ternary, at least AFAIK.

MDoerner: Sorry, was kind of blind reading the comment. Yes, the `l` was missing.

PR #3684, opened 2018-01-08
AssignByValueMakeLocalCopyQuickFix - corrects error when applied on a function 

BZngr:
The AssignByValueMakeLocalCopyQuickFix does not properly handle rewriting the AsTypeContext of a function.  This PR corrects the behavior.  

The current result when this QuickFix is applied to a Function:

Original code:
```vba
Private Function AddTwo(ByVal arg As Long) As Long
  arg = arg + 2
  AddTwo = arg
End Function
```

After the QuickFix makes a local copy of `arg` (Uncompilable):
```vba
Private Function AddTwo(ByVal arg As Long)
  Dim localArg As Long
  localArg = arg As Long  <-- Function's AsTypeContext is pushed to here
  localArg = localArg + 2
  AddTwo = localArg
End Function
```

The PR also enhances the QuickFix to retain any existing comment after the Sub/FunctionStmtContext and match the indentation format of the original code's first line.

### DIFF
--- a/Rubberduck.Inspections/QuickFixes/AssignedByValParameterMakeLocalCopyQuickFix.cs
+++ b/Rubberduck.Inspections/QuickFixes/AssignedByValParameterMakeLocalCopyQuickFix.cs
@@ -32,6 +32,9 @@ namespace Rubberduck.Inspections.QuickFixes
 
         public override void Fix(IInspectionResult result)
         {
+            Debug.Assert(result.Target.Context.Parent is ArgListContext);
+            Debug.Assert(null != ParserRuleContextHelper.GetChild<EndOfStatementContext>(result.Target.Context.Parent.Parent));
+
             var forbiddenNames = _parserState.DeclarationFinder.GetDeclarationsWithIdentifiersToAvoid(result.Target).Select(n => n.IdentifierName);
 
             var localIdentifier = PromptForLocalVariableName(result.Target, forbiddenNames.ToList());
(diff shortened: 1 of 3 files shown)


Rubberduck.Inspections/QuickFixes/AssignedByValParameterMakeLocalCopyQuickFix.cs
```diff
@@ -13,6 +13,8 @@
 using Rubberduck.Parsing.Inspections.Resources;
 using Rubberduck.Parsing.Rewriter;
 using Rubberduck.Parsing.VBA;
+using static Rubberduck.Parsing.Grammar.VBAParser;
+using System.Diagnostics;
```
Vogel612: not sure why that is included?

BZngr: `Rubberduck.Parsing.Grammar.VBAParser` supports use/compiling of references to `EndOfStatementContext`, `AgListContext`, and `AsTypeClauseContext` .  There's a Debug.Assert statement used (which could be removed if needed)  - so `System.Diagnostics`.  Is there a better, more preferred approach?

Vogel612: explicitly checking for your precondition.

BZngr: @Vogel612 - TBH I'm not sure I understand your last comment.  Made some changes consistent with what I could find on the web after searching for "checking preconditions".  Hopefully the latest change is something close to what you are recommending.

PR #3702, opened 2018-01-14
Minor corrections to uses of ITypeLib/ITypeInfo

WaynePhillipsEA:
My first official PR :)

Very minor... just seeing how all this git stuff works.

### DIFF
--- a/Rubberduck.Parsing/ComReflection/ComStruct.cs
+++ b/Rubberduck.Parsing/ComReflection/ComStruct.cs
@@ -27,14 +27,14 @@ namespace Rubberduck.Parsing.ComReflection
 
         private void GetFields(ITypeInfo info, TYPEATTR attrib)
         {
-            var names = new string[255];
+            var names = new string[1];
             for (var index = 0; index < attrib.cVars; index++)
             {
                 IntPtr varPtr;
                 info.GetVarDesc(index, out varPtr);
                 var desc = (VARDESC)Marshal.PtrToStructure(varPtr, typeof(VARDESC));
                 int length;
-                info.GetNames(desc.memid, names, 255, out length);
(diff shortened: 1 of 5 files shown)


Rubberduck.Parsing/ComReflection/ComStruct.cs
```diff
@@ -27,14 +27,14 @@ public ComStruct(ITypeLib typeLib, ITypeInfo info, TYPEATTR attrib, int index)
...
-            var names = new string[255];
+            var names = new string[1];
             for (var index = 0; index < attrib.cVars; index++)
             {
                 IntPtr varPtr;
                 info.GetVarDesc(index, out varPtr);
                 var desc = (VARDESC)Marshal.PtrToStructure(varPtr, typeof(VARDESC));
                 int length;
-                info.GetNames(desc.memid, names, 255, out length);
+                info.GetNames(desc.memid, names, 1, out length);
```
Vogel612: shouldn't 1 be `names.Length` here? (and in ComModule)?
Also why is there no corresponding change in ComEnumerationMember?

WaynePhillipsEA: Yep, I'll do that.   For ComEnumerationMember, the array was declared unnecessarily large, but the call to GetNames was correct.   I'll change it to names.Length there as well though.

bclothier:
For clarification - the calls here all assumes only one name. In case of `ComModule.cs`'s `GetComFields`, there's a loop where we call the `GetNames` -  could we have called it only once by finding out how many fields there are then pass in an array of that many elements without a loop?

WaynePhillipsEA:
@bclothier Unfortunately not, they each have there own member ID, so you have to call individually.

PR #3708, opened 2018-01-17
Rewite of ComWrapperEnumeration

WaynePhillipsEA:
Rewrite of ComWrapperEnumeration in order to offer control of the underlying IEnumVARIANT RCW, which is sometimes leaking past OnDisconnection due to being GC'd

The CLR automagically handles COM enumeration by custom marshalling IEnumVARIANT to a managed class (see EnumeratorToEnumVariantMarshaler) but as we need explicit control over all COM objects in RD, this is unacceptable.  We need to obtain the explicit IEnumVARIANT interface and ensure this RCW is destroyed in a timely fashion, using Marshal.ReleaseComObject.

The automatic custom marshalling of the enumeration getter method (DISPID_ENUM) prohibits access to the underlying IEnumVARIANT interface, and I can't find a way of turning off the automtic custom marshalling here.  To work around it, we must call the IDispatch:::Invoke method directly (instead of using CLRs normal late-bound method calling ability).

### DIFF
--- a/Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
@@ -1,39 +1,143 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
+using System.Runtime.InteropServices;
+using ComTypes = System.Runtime.InteropServices.ComTypes;
 
 namespace Rubberduck.VBEditor.SafeComWrappers
 {
+    // The CLR automagically handles COM enumeration by custom marshalling IEnumVARIANT to a managed class (see EnumeratorToEnumVariantMarshaler)
+    // But as we need explicit control over all COM objects in RD, this is unacceptable.  We need to obtain the explicit IEnumVARIANT interface
+    // and ensure this RCW is destroyed in a timely fashion, using Marshal.ReleaseComObject.
+    // The automatic custom marshalling of the enumeration getter method (DISPID_ENUM) prohibits access to the underlying IEnumVARIANT interface.
+    // To work around it, we must call the IDispatch:::Invoke method directly (instead of using CLRs normal late-bound method calling ability).
+
+    [Guid("00020400-0000-0000-C000-000000000046")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]        
+    interface IDispatch
+    {
+        [PreserveSig] int GetTypeInfoCount([Out] out uint pctinfo);
+        [PreserveSig] int GetTypeInfo([In] uint iTInfo, [In] uint lcid, [Out] out ComTypes.ITypeInfo pTypeInfo);
+        [PreserveSig] int GetIDsOfNames([In] ref Guid riid, [In] string[] rgszNames, [In] uint cNames, [In] uint lcid, [Out] out int[] rgDispId);
+ 
+        [PreserveSig]
+        int Invoke([In] int dispIdMember,
+            [In] ref Guid riid,
+            [In] uint lcid,
+            [In] uint dwFlags,
+            [In, Out] ref ComTypes.DISPPARAMS pDispParams,
+            [Out] out Object pVarResult,
+            [In, Out] ref ComTypes.EXCEPINFO pExcepInfo,
+            [Out] out uint pArgErr);
+    }
+
+    class IDispatchHelper
+    {
+        public enum StandardDispIds : int
+        {
+            DISPID_ENUM = -4
+        }
+
+        public enum InvokeKind : int
+        {
+            DISPATCH_METHOD = 1,
+            DISPATCH_PROPERTYGET = 2,
+            DISPATCH_PROPERTYPUT = 4,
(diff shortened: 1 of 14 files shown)


Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
```diff
@@ -1,39 +1,143 @@
...
 namespace Rubberduck.VBEditor.SafeComWrappers
 {
+    // The CLR automagically handles COM enumeration by custom marshalling IEnumVARIANT to a managed class (see EnumeratorToEnumVariantMarshaler)
+    // But as we need explicit control over all COM objects in RD, this is unacceptable.  We need to obtain the explicit IEnumVARIANT interface
+    // and ensure this RCW is destroyed in a timely fashion, using Marshal.ReleaseComObject.
+    // The automatic custom marshalling of the enumeration getter method (DISPID_ENUM) prohibits access to the underlying IEnumVARIANT interface.
+    // To work around it, we must call the IDispatch:::Invoke method directly (instead of using CLRs normal late-bound method calling ability).
+
+    [Guid("00020400-0000-0000-C000-000000000046")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
```
bclothier: Are we missing `ComImport` here?

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
```diff
@@ -1,39 +1,143 @@
...
+                // could expand this to handle DISP_E_EXCEPTION
+                throw new ArgumentException(string.Format("CallSimpleMethodNoArgs Error invoking DispId {0}:  COM error code is {1}", memberId, hr));
+            }
+
+            return pVarResult;
+        }
+    }
+
+    [Guid("00020404-0000-0000-C000-000000000046")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
```
bclothier: Same question RE: `ComImport` here

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
```diff
@@ -1,39 +1,143 @@
...
+            uint celtFetched;
+            int hr = _enumeratorRCW.Next(1, out currentItemRCW, out celtFetched);
+            // hr == S_FALSE (1) or S_OK (0), or <0 means error
 
-        object IEnumerator.Current => Current;
+            _currentItem = _itemWrapper.Invoke(currentItemRCW);     // creates a null wrapped reference even on end/error, just as a precaution
+            
+            if (hr < 0)
+            {
+                throw new ArgumentException(string.Format("Error invoking IEnumVARIANT::Next().  COM error code is {0}", hr));
```
bclothier: I would consider using [`Marshal. GetExceptionForHR`](https://msdn.microsoft.com/en-us/library/t8t3ykxc%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) instead here.

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
```diff
@@ -1,39 +1,143 @@
...
+            }
+           
+            return (celtFetched == 1);      // celtFetched == 0 at end
```
bclothier: This comment is a bit hard to understand  - could you perhaps elaborate for the benefits of those less savvy w/ COM? As it is, I'm left wondering if I'm supposed to set `celtFetched` = 0 somewhere but I don't see it.

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
WaynePhillipsEA: yup, good catch

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
WaynePhillipsEA: yup,same

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
WaynePhillipsEA: yeah, I'll clarify that one.

Rubberduck.VBEEditor/SafeComWrappers/ComWrapperEnumerator.cs
WaynePhillipsEA: sure, I'll sort that

WaynePhillipsEA:
Thanks @bclothier.  Yes, I've avoided changing the disposal responsibility of the returned RCWs, as I can see cases where the returned objects are cached/stored elsewhere, so implicit disposal would fail.

MDoerner:
Do these changes make it necessary to register the `VBEditor` project for COM interop?

WaynePhillipsEA:
No, these are just internal changes.

PR #3711, opened 2018-01-18
Refactor ParserRuleContextHelper to ParserRuleContextExtensions

BZngr:
This PR refactors the functions within ParserRuleContextHelper.cs (PRCHelpr) to ParserRuleContextExtensions.cs (PRCExtensions).

The PR is a follow-on to the suggestion made in comment(s) for #3684  where I had made use of PRCHelper.  In addition to porting the functions, I am proposing some modifications and renaming.

1. In the existing PRCExtensions, there were several overloaded versions of `GetProcedureSelection(...)`.  These overloaded functions may have been solving a problem that no longer exists.  None of the functions are referenced by the code base (except one that is referenced by a test).  I have removed these un-used functions as part of this PR.

I am also proposing some new method names. The terms "Parent" and "Child" are used in a few methods names where "Ancestor" and "Descendent" would be more precise/descriptive given the method's current behavior.

1. The overloaded methods `HasParent `are changed to `IsDescendentOf(...)`. 
 
2. Method `FindChildren `did not stop at the context's "children" list but rather walked the entire parse tree.  I have changed this method name to `GetDescendents(...)`. 
 
3. Method `GetParent `- which would imply (to me) a function that simply returns `context.Parent`, returned the the context itself (if it is the correct Type), or the first ancestor of the correct Type.  I've modified this function to only return the first qualifying Ancestor (IOW, a context cannot be it's own Ancestor/Parent).  I have changed the name of this method to `GetAncestor`.  The `DeclarationFinder `referenced `HasParent` and `GetParent`. It has been updated to use equivalent logic: e.g.,	`context is T || context.IsDescendentOf<T>()`


Retailcoder.VBE/UI/RubberduckUI.resx
```diff
@@ -381,6 +381,15 @@ Warning: All customized settings will be lost.  Your old file will be saved in '
...
   </data>
+  <data name="TodoMarkerBug" xml:space="preserve">
+    <value>BUG</value>
+  </data>
+  <data name="TodoMarkerNote" xml:space="preserve">
+    <value>NOTE</value>
+  </data>
+  <data name="TodoMarkerTodo" xml:space="preserve">
+    <value>TODO</value>
+  </data>
```
Vogel612: Warning: ghost resx reappeared

BZngr: Not sure how to do this...However, https://stackoverflow.com/questions/2047465/how-can-i-delete-a-file-from-git-repo looks promising.   I lack the experience and confidence with GIT to 'just do it'...Thoughts/comments/recommendations?

I definitely can see the ghost file if I generate a list of files in my local repository.  But, the file is not present/visible in File Explorer.

Vogel612: try using `git rm --cached` to prevent windows misinterpreting the given path and subsequently deleting the actual resx file.

BZngr:
Removed the ghost file "Retailcoder.VBE/UI/RubberduckUI.resx" using Git on my local repo.  Seems to have done the right thing.  Since the file deletion shows up in the diff here, I would think it means the ghost file is still in the master GitHub repository.  Maybe this PR when merged, will remove it.

PR #3724, opened 2018-01-25
It's 2018!

retailcoder:


rkapka:
Wouldn't it be better to have a `{0}` placeholder in the resource string and display it using `string.Format`? Is that not possible for some reason?

retailcoder:
Ha, indeed that would be the end-all-be-all fix! =)

On Jan 27, 2018 2:31 PM, "rkapka" <notifications@github.com> wrote:

Wouldn't it be better to have a {0} placeholder in the resource string and
display it using string.Format? Is that not possible for some reason?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/rubberduck-vba/Rubberduck/pull/3724#issuecomment-361009044>,
or mute the thread

PR #3744, opened 2018-02-01
Fix letter ranges

### LINKED ISSUE #3741: Parse Error related to antlr4.6 update in  .2809 prerelease build
Using the prerelease build .2809, this code generates a parse error:

```
Public Sub Foo()
    Dim a As Integer
    Dim b As Integer
    Dim c As Integer
    a = 1
    b = 2
    c = a - b  ' <<<<<<<< Parse Error, No alternative input 
    Debug.Print c
End Sub
```

MDoerner:
This PR closes #3741.

This PR returns matching letter ranges to the parser, using predicates. 
Moreover, it fixes the `IDENTIFIER` lexer rule, which broke with the Antlr update because escaped hyphens are somehow no longer recognized in character classes.


Rubberduck.Parsing/Grammar/VBALexer.g4
```diff
@@ -295,8 +295,7 @@ SINGLEQUOTE : '\'';
 UNDERSCORE : '_';
 WS : [ \t];
 GUIDLITERAL : '{' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '}';
-LETTER_RANGE : LETTER WS? MINUS WS? LETTER;
-IDENTIFIER :  ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\] ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; ]*;
+IDENTIFIER :  ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
```
bclothier: Hmm, we added a hyphen at the end of the subexpression `[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-]` but we did not remove one of the two original hyphens. Should we remove the `\-` from the `...^&*+\-+=;` part? Even if ANTLR is interpreting the hyphen as a character class `\-+`, this doesn't make sense because `+` is 0x2B while `\` is 0X5C.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -289,16 +291,17 @@ defType :
 // singleLetter must appear at the end to prevent premature bailout
 letterSpec : universalLetterRange | letterRange | singleLetter;
 
-// we're purpusefully not actually making sure this is an ASCII letter
-// that'd be too much of a hassle with the lexer to get working
-singleLetter : {_input.Lt(1).Text.Length == 1}? IDENTIFIER;
+singleLetter : {_input.Lt(1).Text.Length == 1 && Regex.Match(_input.Lt(1).Text, @"[a-zA-Z]").Success}? IDENTIFIER;
```
bclothier: I think this needs a explanatory comment why we are resorting to .NET's regex here, instead of using ANTLR's quirky regex.

RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -2136,6 +2135,26 @@ End Sub
...
+        {
+            const string code = @"
+Public Sub Foo()
+    Dim a As Long
+    Dim b As Long
+    Dim z As Long
+    a = 1
+    b = 2
+    z = a-b
+    b = a-z
```
bclothier: Note that VBA normally will pretty print such expression to `b = a - z`. Though the lexer rules does not permit whitespace, I think it'll be good to test the version with whitespace to ensure we don't ever match the pretty printed expression.

Rubberduck.Parsing/Grammar/VBAParser.g4
Vogel612: ANTLR regexes are for Lexer rules. We can't use them in the Parser. Also we can't use ANTLR Regexes in semantic predicates, because those need to be written in the target language...

Vogel612: The more interesting comment would be regarding why we're not using `Char.IsLetter`, but reading the docs of that method makes it clear :/

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -289,16 +291,17 @@ defType :
...
+
 // We make a separate universalLetterRange rule because it is treated specially in VBA. This makes it easy for users of the parser
 // to identify this case. Quoting MS VBAL:
 // "A <universal-letter-range> defines a single implicit declared type for every <IDENTIFIER> within 
 // a module, even those with a first character that would otherwise fall outside this range if it was 
 // interpreted as a <letter-range> from A-Z.""
-universalLetterRange : {_input.Lt(1).Text.StartsWith("A") && _input.Lt(1).Text.EndsWith("Z")}? LETTER_RANGE;
-letterRange : LETTER_RANGE;
+universalLetterRange : {_input.Lt(1).Text.Equals("A") && _input.Lt(3).Text.Equals("Z")}? IDENTIFIER MINUS IDENTIFIER;
```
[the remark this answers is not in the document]
bclothier: @Vogel612 , FWIW - the [spec](https://msdn.microsoft.com/en-us/library/ee177098.aspx?f=255&MSPPError=-2147217396) does not appear to make any allowance for whitespace. Though it is legal to enter in `DefBool a - c`, it get pretty printed to `DefBool A-C`. Interestingly, if one enters it malformed (e.g. `DefBoo A-C`), this get pretty printed to `Defboo A - c`. Importing even pretty prints, so I think we are fine to not handle the whitespace.

Issue #3753, opened 2018-02-05
Parse error using build .2827


daFreeMan:
Win7 64-bit, Office 2010 32-bit, RD .2827

I have clean compiling code, yet build .2827 is throwing a parse error:

```
2018-02-05 13:43:43.4370;ERROR-2.1.2.2827;Rubberduck.Parsing.VBA.ParseCoordinator;Unexpected exception thrown in parsing run. (thread 38).;System.Runtime.InteropServices.COMException (0x80004005): The expression you entered refers to an object that is closed or doesn't exist.
```

Full log attached.
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/1696366/RubberduckLog.txt)

Based on some additional testing:
* I get a clean parse when running .2825
* I get the error with NO forms open
* I get a clean parse with ALL forms open

Vogel612:
IIUC this is going to be related to the changes in how Rubberduck accesses the COM Api exposed by the VBE. Recently this has changed very heavily...

bclothier:
I found what I believe is a variant. Trying opening all forms in design as @daFreeMan originally reported but that did NOT fix my parse error.

In my log, I have this:

```2018-02-12 16:02:53.3594;ERROR-2.1.6617.23383;Rubberduck.Parsing.VBA.ParseCoordinator;Unexpected exception thrown in parsing run. (thread 17).;System.Runtime.InteropServices.COMException (0x80004005): Error HRESULT E_FAIL has been returned from a call to a COM component.
   at Microsoft.Vbe.Interop.VBComponentClass.get_Type()
   at Rubberduck.VBEditor.SafeComWrappers.VBA.VBComponent.get_Type() in C:\GitHub\Rubberduck\Rubberduck\Rubberduck.VBEEditor\SafeComWrappers\VBA\VBComponent.cs:line 16
   at Rubberduck.Navigation.CodeExplorer.CompareByType.Compare(CodeExplorerItemViewModel x, CodeExplorerItemViewModel y) in C:\GitHub\Rubberduck\Rubberduck\RetailCoder.VBE\Navigation\CodeExplorer\CodeExplorerItemViewModel.cs:line 96
   at System.Linq.EnumerableSorter`2.CompareKeys(Int32 index1, Int32 index2)
   at System.Linq.EnumerableSorter`1.QuickSort(Int32[] map, Int32 left, Int32 right)
[...]

PR #3764, opened 2018-02-14
Attempt to fix code modules and fix for projects not getting removed

MDoerner:
This PR tries to deal with issue #3753. Unfortunately, I cannot verify that it fixes the issue because I cannot reproduce the issue reliably myself.

The two attemps made here are the following:

1. I introduced a retry approach for accessing the `CodeModule` property of a `VBComponent`, which sometimes fails for some unknown reason.

2. I made `IsNewOrModified` always compute the content hash, even if the module is clearly new. This is an attempt to access the `CodeModule` property for the first time at the same stage of the parsing process as before the PR that made the issue appear.

### DIFF
--- a/Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
@@ -22,7 +22,7 @@ namespace Rubberduck.VBEditor.SafeComWrappers
             Target = target;
             _rewrapping = rewrapping;
 
-            if (!rewrapping && target != null)
+            if (!rewrapping && target != null && (this is IVBProject))
             {
                 _comSafe = ComSafeManager.GetCurrentComSafe();
                 _comSafe.Add(this);
--- a/Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponent.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponent.cs
@@ -1,7 +1,10 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
+using NLog;
 using Rubberduck.VBEditor.Extensions;
 using Rubberduck.VBEditor.SafeComWrappers.Abstract;
 using Rubberduck.VBEditor.SafeComWrappers.Office.Core.Abstract;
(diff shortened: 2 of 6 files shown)


Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponent.cs
```diff
@@ -11,14 +14,49 @@ namespace Rubberduck.VBEditor.SafeComWrappers.VBA
 {
     public class VBComponent : SafeComWrapper<VB.VBComponent>, IVBComponent
     {
+        private const int MaxRetryCount = 10;
+
+        private enum ComErrorCodes : uint
```
bclothier: Technically HRESULT are signed integers, with negative values indicating an error. I would use `int`, not `uint` here.

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
```diff
@@ -22,7 +22,7 @@ protected SafeComWrapper(T target, bool rewrapping = false)
             Target = target;
             _rewrapping = rewrapping;
 
-            if (!rewrapping && target != null)
+            if (!rewrapping && target != null && (this is IVBProject))
```
bclothier: Considering this only happens in Excel but not in Access, I wonder if we need to further restrict this to Excel only to avoid leaks in other host applications?

MDoerner: I do not know of any way to figure out the host application at that point.

Rubberduck.VBEEditor/SafeComWrappers/VBA/VBComponent.cs
MDoerner: I know that they are integers. However, I could not find a way to assign the literal if the enum type is `int`. I think we do the same for window messages.

rkapka: `unchecked((int)0x80004005)`

Rubberduck.VBEEditor/SafeComWrappers/SafeComWrapper.cs
MDoerner: Moreover, are you sure the issue does not exist in any other host but Excel?

bclothier: At this point I do know it does not occur in Access. I can test Word and PowerPoint later to ascribe. In case it matters, this was observed on Excel 2010.

MDoerner:
I got to the conclusion that some things in this PR are not a great idea.

I close this issue and will open another PR in the next days with an another attempt of a fix following a slightly different path.

PR #3766, opened 2018-02-15
Hotfix CodeCov invalid configuration

Vogel612:


Vogel612:
If this works correctly, we should see a different comment layout, mostly without the image at the top

Vogel612:
The coverage decrease is because we now actually ignore Coverage for RubberduckTests which decreases average coverage quite a bit

PR #3775, opened 2018-02-20
Fix for AttributeNode

### LINKED ISSUE #3772: Parse Error - I can not find why - where to look for the problem?
I got Parse Error - 2 refernces but no information what is the concrete error.
Can anyone help?

Thanks in advance
Frank

MDoerner:
Fixes #3772 

The simple most fix for the issue, which is contained in commit one of this PR, is a two character change. (`_values` --> `Values` in `HasValue`)

However, to make `AttributeNodes` less confusing and prone to throw exceptions in case the constructor taking a context is used, I redesigned the class a bit.

### DIFF
--- a/Rubberduck.Parsing/VBA/Attributes.cs
+++ b/Rubberduck.Parsing/VBA/Attributes.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


Vogel612:
I did approve these changes two days ago. Unless there's a dissenting review within the next hour, I'll just merge this ...

Issue #3675, opened 2018-01-06
Managed memory leak on disconnect


MDoerner:
RD seems to have a managed memory leak on disconnect.

This can be observed as follows:

1) Comment out the shutdown invokation in `UIDispatcher.Shutdown()`. (See #3645 for why you should do this.)
2) Add `GC.Collect(); GC.WaitForPendingFinalizers()` to the finally block of `ShutdownAddIn` in `Extension`. (This is nessesary since no finalizers will ever run until shutdown, provided you have enough memory.)
3) Open RD in the debugger.
4) Unload RD via the add-ins dialog.
5) Reload RD via the add-in dialog.
6) Close the host application.

In the debug output, you should be able to observe the following:

- There are messages that the finalizers for the `DockableToolwindowPresenter`s have run.
- All but the one for SC get finalized right on disconnect (unload and shutdown).
- Both SC presenters get finalized only on the final shutdown, and much later than anything else.
- On reload, there are exceptions for the three commands `OpenProjectPropertiesCommand`, `ImportCommand` and `ExportAllCommand` stating that an RCW has already been disconnected from its target.

I investigated a bit further by adding addig finalizers for logging to some suspects that could hold the SC presenter alive. It seems that the presenter is kept alive by two commands, the `CommitCommand` and the `UndoCommand`, which are both kept alive by the `CodeExplorerViewModel`, I could find out yet what is keeping this alive.

That the CE is kept alive means that everything it references, beit directly or indirectly, is kept alive as well; that is half of RD.

The exceptions on the three commands seem to be connected. All three are referenced by the CE view model and their `EvaluateCanExecute` methods contain the accesses to the stale RCWs. Somehow, they still seem to react to some change notifications.

bclothier:
Purely by happenstance, I noticed there were actually _two_ code explorer windows:
![image](https://user-images.githubusercontent.com/2367644/34728818-880434d6-f520-11e7-86ad-6628218573cc.png)

The useless code explorer window is initially invisible and must be programmatically made visible via the `Immediate` windows. Note that it's possibly easy to miss because a `For Each` will not iterate all `VBE.Window` in the `VBE.Windows` collection. A `For` must be used instead.

```
for i = 0 to application.VBE.Windows.Count -1: ?application.VBE.Windows(i).caption:next
Code Explorer
Code Explorer
Source Control
Code Metrics
Todo Explorer
Test Explorer
Code Inspections
Project - ItImpactTemplate

Properties
[...]

Issue #3645, opened 2017-12-26
UiDispatcher shutdown/restart problem causing Addin unload/reload to fail


WaynePhillipsEA:
As per title, there is a problem in the UiDispatcher that causes the weird crash when re-loading the RD addin from the VBE after unloading.

To demonstrate, we can remove the InvokeShutdown method call, and the unload/reload works without crashing (but with some RCW disconnection errors in the log which will need looking at)

bclothier:
I analyzed this a bit using the branch for the PR #4456 I can observe that:

1) the numbers of `Add` entries and `Removed` entries from the final trace of COM Safe do not always balance. At minimum, 3 will not have their `Removed` entries; the `IVBE`, the `IAddIn`, and the `IHostApplication`. 
2) We will get `RaceOnRCWCleanup` warnings in process of disposing the COM safe.
3) Even if we exempt those objects from disposal, we still get the warning on say, a `ICodePane`. It doesn't even have to be an active code pane.
4) Forcing the disposal in the UI thread does not help.

I'm inclined to suspect that to correctly dispose all the items, we need to ensure that there are no other threads running, and that they all have released their use of COM before we may proceed to dispose of COM Safe.

I have not seen any RCW disconnection errors, however.

MDoerner:
For context, please note that the call to `InvokeShutdown` is currently commented out in the `UiDispatcher`. This does not seem to be ideal but symptomatically fixes the original issue.

MDoerner:
Regarding, the `RaceOnRCWCleanup`, this cannot be caused by the `IVBE` or `IAddIn` because we still hold a reference to them when we dispose the COM safe in `Extension.ShutdownAddIn`. That these are still in the COM safe is to be expected because we use it call dispose. Excempting them from addition to the COM safe means that we will not dispose them.

bclothier:
FWIW, I just checked the `InvokeShutdown`. First, because the `UiDispatcher` has been since refactored, it no longer works as-is. That said, I made a static instance and attempted `InvokeShutdown` again to see if anything had changed since. It has not.

I'm not as confident that we will ever have `InvokeShutdown` re-implemented.

bclothier:
[...]

PR #3780, opened 2018-02-21
Pimp the UiDispatcher

MDoerner:
This PR adds the capability to the `UIDispatcher` to start tasks on the UI thread. With this capability we can now wait for an action to finish executing on the UI thread and get the return value if a `Func<T>` has been passed in.

Moreover, I commented out the shutdown code in the `UiDispatcher` in order to avoid the exception on reload we experience so far. (Ref #3645) This does avoid said exception but the state RD reloads in is not completely clean. This is probably connected to our managed memory leak at shutdown/unload. (Ref #3675)


RetailCoder.VBE/UI/Command/MenuItems/UiDispatcher.cs
```diff
@@ -55,18 +58,62 @@ public static void Invoke(Action action)
             }
         }
 
+        /// <summary>
+        /// Starts a task on the ui thread.
+        /// </summary>
+        /// <param name="action">The action that will be executed on the UI
+        /// thread.</param>
+        /// <param name="token">Optional cancellation token</param>
+        /// <param name="options">Oprional TaskCreationOptions</param>
```
Vogel612: typo: optional

WaynePhillipsEA:
referencing #3645, which this PR technically closes, but we may want to leave open until the managed memory leaks have been plugged, so that the shutdown code can then hopefully be reinstated.

bclothier:
I have a question about the `Initialize` process - it seems that we do this as a way to ensure that the `Initialize` is called on the UI thread and then we run `CheckInitialization` in all other methods. I'm wondering if it'd be more cleaner to simply name the UI thread within the `Extension` so we can always pull it out of the `Threads` collection in the ctor or perhaps store it in a static class, so that this can be always "initialized"?

MDoerner:
The concept .Net uses for synchronization is the `SynchronisationContext`, which is not identical to a thread. The Winforms `SynchronisationContext` corresponds exactly the UI thread, other types correspond to multiple threads.

In `Extension` only the default `SynchronisationContext` exists. You can see this because, by convention, `SynchronisationContext.Current` returns `null`. This default context corresponds to the current thread and the thread pool. So, we really do not want that context. Instead we call `Initialize` only at a point where the actual UI context has been generated.

rubberduck203:
Overall the code coverage has remained the same, but I noticed [ReferenceResolverRunnerBase dropped significantly](https://codecov.io/gh/rubberduck-vba/Rubberduck/pull/3780/diff?src=pr&el=tree#diff-UnViYmVyZHVjay5QYXJzaW5nL1ZCQS9SZWZlcmVuY2VSZXNvbHZlUnVubmVyQmFzZS5jcw==).

MDoerner:
@rubberduck203 That file is not even in the diff of this PR.

PR #3783, opened 2018-02-22
Compile before parse

### LINKED ISSUE #3732: Compile before parsing
With type information API provided in PR #3716 , we now can request for a silent compilation of the VBA project. Before we parse, we should now require a compilation, to save ourselves from parser errors.

Furthermore, we can compile on a per-component level, a option that has not been exposed for long time (probably since Office 2000?), which may enable us to support partial parsing. 

Bonus: maybe update code explorer to mark components as having compilation errors. 

Note: In chat there has been discussion of updating the parser to continue after an error (cf. cancelled PR #3632 ). Considerations are needed as to whether this means we can skip the requirement to compile. Note that certain compilation errors also prevent VBA code from running at all; but not all errors.

bclothier:
Closes #3732 

1) Enable ability to compile prior to parsing, implemented in `ReparseCommand`
2) Provide a new setting to opt-out of this. 

ATM, if a project (or projects) fails to compile, they will be prompted with a warning and an option to continue parsing anyway with default of `No`. 

It is hoped that it will help eliminate some parsing errors due to simply forgetting to compile the project.


RetailCoder.VBE/UI/RubberduckUI.resx
```diff
@@ -1780,4 +1780,13 @@ Would you like to import them to Rubberduck?</value>
   <data name="GeneralSettings_ModifierWarning" xml:space="preserve">
     <value>Modifier Warning</value>
   </data>
+  <data name="Command_Reparse_CannotCompile" xml:space="preserve">
+    <value>The following project(s) cannot be compiled, which wil likely will most likely result in parser errors. Continue anyway? {0}</value>
```
WaynePhillipsEA: typo ' which wil likely will most likely '

WaynePhillipsEA:
Looks good to me.  Thanks @bclothier!

RetailCoder.VBE/UI/Command/ReparseCommand.cs
```diff
@@ -26,6 +39,44 @@ protected override bool EvaluateCanExecute(object parameter)
...
+                                continue;
+                            }
+
+                            if (!VBETypeLibsAPI.CompileProject(project))
+                            {
+                                failedNames.Add(project.Name);
+                            }
+                        }
+                    }
+                }
```
Vogel612: I think it'd be nice to extract this into a separate method to keep `OnExecute` rather "high-level" Something like:

    If (_settings.CompileBeforeParse)
    {
        if (!AllProjectsCompile(out var failedNames)) 
        {
            // message box code
        }
    }

RetailCoder.VBE/UI/RubberduckUI.resx
bclothier: There are people who stutter as they speak. Then there's me who stutter as I type....

WaynePhillipsEA:
According to the IConnectionPoint::Advise documentation, https://msdn.microsoft.com/en-us/library/windows/desktop/ms678815(v=vs.85).aspx:

pdwCookie ... If the connection was not successfully established, this value is zero

Currently, the NotAdvising default const value for the cookie is set to -1, which is not correct.  It should be 0, as -1 could be a valid cookie value.

RetailCoder.VBE/UI/Command/ReparseCommand.cs
```diff
@@ -26,7 +39,55 @@ protected override bool EvaluateCanExecute(object parameter)
...
+                foreach (var project in projects)
+                {
+                    using (project)
+                    {
+                        if (project.Protection != ProjectProtection.Unprotected)
+                        {
+                            continue;
+                        }
+
+                        if (!VBETypeLibsAPI.CompileProject(project))
```
MDoerner: Please note that this makes the command untestable.

RetailCoder.VBE/UI/Command/ReparseCommand.cs
```diff
@@ -26,7 +39,55 @@ protected override bool EvaluateCanExecute(object parameter)
...
+
+            return failedNames.Any();
+        }
+
+        private bool PromptUserToContinue(List<string> failedNames)
+        {
+            var formattedList = string.Concat(Environment.NewLine, Environment.NewLine,
+                string.Join(Environment.NewLine, failedNames));
+            var msgbox = new MessageBox();
+            var result = msgbox.Show(
```
MDoerner: This too

bclothier:
Addressed the concerns RE: testability of the class. In my assessment, I don't think the `VBETypeLibAPI` class really needed to be static; it was only so because there were really no need to maintain an instance state AFAICS. However, since a static class can't have an interface and thus be mocked, I converted it and updated the `ReparseCommand` to use IoC accordingly.

WaynePhillipsEA:
yeah, TypeLib API need not be static, so above change seems fine.

PR #3791, opened 2018-02-25
Enforce ending NEWLINE on commentBody

### LINKED ISSUE #3789: Comment in UDT is ignored
With ver: 2.1.2.2886

If I create a UDT like:

```
Private Type tX
    foo As String
    bar As Long
    'foobar as shouldNotBeVisible
End Type
```
then in the **code explorer** `foobar` shows up the same as `foo` and `bar`.

Vogel612:
This requires an additional modification to endOfStatement to keep
consuming any "leading" whitespace for the following statement.
Fixes #3789

### DIFF
--- a/Rubberduck.Parsing/Grammar/VBAParser.g4
+++ b/Rubberduck.Parsing/Grammar/VBAParser.g4
@@ -502,8 +502,9 @@ subroutineName : identifier;
 // 5.2.3.3 User Defined Type Declarations
 publicTypeDeclaration : ((GLOBAL | PUBLIC) whiteSpace)? udtDeclaration;
 privateTypeDeclaration : PRIVATE whiteSpace udtDeclaration;
-udtDeclaration : TYPE whiteSpace untypedIdentifier endOfStatement udtMemberList endOfStatement END_TYPE;  
-udtMemberList : udtMember (endOfStatement udtMember)*; 
+// member list includes trailing endOfStatement
+udtDeclaration : TYPE whiteSpace untypedIdentifier endOfStatement udtMemberList END_TYPE;  
+udtMemberList : (udtMember endOfStatement)+; 
 udtMember : reservedNameMemberDeclaration | untypedNameMemberDeclaration;
 untypedNameMemberDeclaration : untypedIdentifier whiteSpace? optionalArrayClause;
 reservedNameMemberDeclaration : unrestrictedIdentifier whiteSpace asTypeClause;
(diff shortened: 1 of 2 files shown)


RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -440,7 +440,8 @@ public void TestModuleConfig()
         [Test]
         public void TestEmptyComment()
         {
-            string code = @"'";
+            string code = @"'
```
MDoerner: Why these test changes. According to the VBA specification, this is legal code.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -877,7 +878,8 @@ commentOrAnnotation :
 ;
 remComment : REM whiteSpace? commentBody;
 comment : SINGLEQUOTE commentBody;
-commentBody : (LINE_CONTINUATION | ~NEWLINE)*;
+// commentBody must terminate with a NEWLINE, see VBA Language spec, section 3.3.1 and 3789
```
MDoerner: The comment is not exactly true. In essence, the specification states that a comment has to end with the end of a logical line, which includes the end of the file.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -877,7 +878,8 @@ commentOrAnnotation :
...
+commentBody : (LINE_CONTINUATION | ~NEWLINE)* (NEWLINE | EOF);
```
MDoerner: Why exactly do we explicitly allow `LINE_CONTINUATION` here. It is included in `~NEWLINE`. (It is a token and thus no subtokens are visible to the parser.)

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -877,7 +878,8 @@ commentOrAnnotation :
...
 annotationList : SINGLEQUOTE (AT annotation whiteSpace?)+ (whiteSpace? COLON commentBody)?;
```
MDoerner: This is inconsistent now. If there is no trailing comment, the `NEWLINE` will not get consumed.

Vogel612: What do you think of the following solution?

``` 
annotationList : SINGLEQUOTE (AT annotation whiteSpace?)+ ((COLON commentBody?) | (NEWLINE | EOF));
``` 

Of note: I removed the redundant `whiteSpace?` before the `COLON` and the logical LineEnd is definitely consumed.  
It might be useful to define `(NEWLINE | EOF)` as `logicalLineEnd` to avoid repetition there...

PR #3798, opened 2018-02-27
Provide a method of pumping COM objects in a synchronous block

bclothier:
This is a result of experimentation from last weekend. This is currently unused because the PR does not include any actual cancellation of parsing/inspection which would likely need to use this. In the experimentation, it appears that it does help with processing the messages for other running threads. I chose to not include the cancellation implementation because there are too many other problems. I do hope that at least this will be useful when we do implement the synchronization for the cancellation in a subsequent PR. 

The understanding is that if one wants to block a thread, and there are other threads that needs COM access, the blocking thread must pump the messages for those threads, or all threads will be deadlocked. Because those COM objects are managed by VBA, and VBA most likely has its custom message pump, it become necessary to calls the `rtcDoEvents` method so that the other threads waiting on their COM marshaling can complete their operations while the thread requesting the `rtcDoEvents` can block synchronously.

It was originally a separate class but given that it has to run on UI thread and because it should only be used by parser & inspections (where there are multiple threads running), it seemed more logical to organize it with the `UiDispatcher` class in the `Rubberduck.Parsing` project.

Furthermore, this attempts to support both VBE7 and VBE6 clients so it has the scaffolding needed to determine the DLL to use and it will remember the DLL used for subsequent calls.


WaynePhillipsEA:
Thanks for this.  Though the description above is not quite to my liking.

More accurately put:
As a side-effect of COM STA marshaling, if one wants to block the main (i.e. UI) thread, and there are any other threads that are potentially calling methods on STA COM objects, the main thread MUST pump windows messages whilst blocking, else deadlocks may occur.

For absolute clarity; this should only be needed, and only be called inside the main thread, whilst blocking.   It does not need to be triggered at all by any other thread.

Typical use case for RD will be calling this in a loop whilst waiting for results from COM-enabled threads, in the response to a COM event on the main thread.

WaynePhillipsEA:
^ and for those reasons, I don't believe UiDispatcher is the correct place for this.

MDoerner:
I also think that this should not be part of the `UiDispatcher` because this new functionality is not really part of the specific responsibility of the `UiDispatcher`: delegate the execution of tasks to the UI thread.

bclothier:
Based on Wayne's feedback I was going to replace the `Invoke` with a guard clause to check that the thread calling the method is in fact the UI thread. Given that there are suggestions to move it out into another class, does that imply re-implementing the setup for the `SynchronizationContext` in the new class in order to have the guard clause?

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        private static void CheckContext()
+        {
+            if (UiContext == null)
+            {
+                throw new InvalidOperationException("UiSynchronizer is not initialized. Invoke Initialize() from UI thread first.");
+            }
+            
+            if (UiContext != SynchronizationContext.Current)
```
WaynePhillipsEA: can't we just check if SynchronizationContext.Current is not null, which would itself ensure we are in the main thread?  then you can remove UiContext and  Initialize() as well.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
```diff
@@ -0,0 +1,111 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Rubberduck.Parsing.UIContext
+{
+    public static class UiSynchronizer
```
WaynePhillipsEA: I suspect we might eventually want to call other exposed VBE dll functions, and the structure for detecting Vbe6/Vbe7 could be reused.  Perhaps consider renaming to VbeDllAccessor, or something along those lines.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
```diff
@@ -0,0 +1,111 @@
...
+            catch
+            {
+                try
+                {
+                    result = rtcDoEvents6();
+                    _version = DllVersion.Vbe6;
+                }
+                catch
+                {
+                    // we shouldn't be here.... Rubberduck is a VBA add-in, so how the heck could it have loaded without a VBE dll?!?
```
WaynePhillipsEA: I guess when VBE8.dll gets released in 2030...  though we'll worry about that when it happens :)

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
bclothier: AIUI, not just the UI thread can have one so it's not a guarantee that the `SynchronizationContext.Current` is in fact the UI thread; it could be anything else. See the question + the answer in the PR #3780. Furthermore, see this [blog post where it discusses the case where the `Current` is in fact not the UI thread](http://www.codewrecks.com/blog/index.php/2012/10/29/syncronizationcontext-current-is-not-null-still-not-be-able-to-access-ui-controls-in-wpf/).

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
bclothier: At this point, I'm gambling that those will be the only functions we need. If it's not, then it will definitely have to be split into its own class solely responsible for exposing VBE's exported functions.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
```diff
@@ -0,0 +1,111 @@
...
+                case DllVersion.Vbe7:
+                    return rtcDoEvents7();
+                case DllVersion.Vbe6:
+                    return rtcDoEvents6();
+                default:
+                    return DetermineVersionAndExecute();
+            }
+        }
+
+        private static int DetermineVersionAndExecute()
```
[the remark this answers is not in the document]
bclothier: My concern is that because it's a `static` class with a `static` ctor, it's hard to control when it gets initialized and exceptions from a static ctor are particularly troublesome to debug. It is also a micro-optimization because the `switch` will only execute the function after it has considered all other cases, which is better for more typical case where function is re-called subsequent times.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
WaynePhillipsEA: Ah yes, the issue is that there is a SynchronizationContext.SetSynchronizationContext, which I overlooked.  So indeed, we can't rely on it only being non-null for the main thread.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
WaynePhillipsEA: I'd bet the other way... come back in a few years and we'll find out :+1:

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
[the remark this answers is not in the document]
WaynePhillipsEA: my concern was more for if we were changing it into a VbeDllAccessor or whatever, then it would then make more sense to have a standard entry point for the DLL detection rather than having lots of similar DetermineVersionAndExecuteDoEvents, DetermineVersionAndExecuteXYZ... methods with repeated code.   But as you're not doing that... ignore me.

Rubberduck.Parsing/UIContext/UiSynchronizer.cs
WaynePhillipsEA: Perhaps it's just me, but the name 'UiSynchronizer' still doesn't sit well with me.  It appears the purpose of this class is to call the VBE exposed DoEvents function, with a guard to ensure we only call it from the main thread.   We're not really syncing anything?

MDoerner: My word. I think something in the direction of `ComMessagePumper` would be more appropriate.

bclothier: What I didn't mention was that I envisioned that in a sequent PR, the class would provide mechanism for actually synchronizing using something like `ReaderWriterLockSlim` to help manage the COM access. Should this be made into a dedicated class responsible for pumping message solely?

WaynePhillipsEA: @MDoerner what's with the 'my word' attitude?  Naming stuff is hard.   Important to get it right, for the sake of those who come after you.

WaynePhillipsEA: @bclothier ... I would.  up to you of course.

MDoerner: @WaynePhillipsEA I guess this is a language issue. Read it as _I fully agree._.

WaynePhillipsEA: @MDoerner in which case, I apologize. I read it like 'oh my word', which over here at least, and in context, would mean along the lines of 'why the f*ck are we even nitpicking at this'.

bclothier:
After listening to everyone's feedback, I decided to the following changes:

1) refactor the classes so that they hopefully now have a single responsibility
2) rename classes and methods to better reflect what they do. That include removing `DoEvents` which I never really liked since it says nothing. It is now `ComMessagePumper.PumpMessage`, which is a bit far way from the `DoEvents` it actually calls but I think describes the intention better.
3) Moved them into `VBEEditor` project since that seems to be better fit for what they do. 

I also removed `UiSynchronizer`; a form of it may or may not be put back in a subsequent PR.

Rubberduck.VBEEditor/ComManagement/ComMessagePumper.cs
```diff
@@ -0,0 +1,56 @@
...
+    public static class ComMessagePumper
+    {
+        private static SynchronizationContext UiContext { get; set; }
+        private static IVBERuntime Runtime { get; set; }
+
+        public static void Initialize()
+        {
+            if (UiContext == null)
+            {
+                UiContext = SynchronizationContext.Current;
```
Vogel612: that is a difficult assumption right here. Is there any way to guarantee that the current synchronization context is the UI Thread?

Rubberduck.VBEEditor/ComManagement/ComMessagePumper.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Threading;
+using Rubberduck.VBEditor.ComManagement.VBERuntime;
+
+namespace Rubberduck.VBEditor.ComManagement
+{
+    public static class ComMessagePumper
+    {
+        private static SynchronizationContext UiContext { get; set; }
+        private static IVBERuntime Runtime { get; set; }
```
Vogel612: I haven't seen this be set in the diff. I wonder whether it'd make sense to make this a Readonly and directly couple this to the `VBERuntimeAccessor`.

Rubberduck.VBEEditor/ComManagement/ComMessagePumper.cs
bclothier: ATM, it uses the same assumption as `UiDispatcher` does, having its `Initialize` called in both `ParserState.cs` and `App.cs`, which works well enough.

bclothier: Well, one way we could help make the code testable for both this and the `UiDispatcher` is to separate the task of creating the UI context out of the classes and instead have those classes call a method to check they are in the correct context.... ?

Rubberduck.VBEEditor/ComManagement/ComMessagePumper.cs
bclothier: Good catch. Fixed.

bclothier: Sorry, no. It's unfixed. :D What I didn't remember was that it's `static` and `readonly` can't be initialized in a non-static ctor.

PR #3821, opened 2018-03-03
Compile before parse fix

### LINKED ISSUE #3807: CompileProject not working properly if 'Compile On Demand' is set
So I couldn't stay away, as I found an issue with the TypeLib API...

The new CompileProject feature of the TypeLib API has an interesting quirk when 'Compile On Demand' is set inside the VBE options.

When 'Compile On Demand' is ON (the default), our CompileProject call seemingly only does a minimal compilation, meaning that procedure level source code is not compiled until the procedure is actually called into at runtime.   Seemingly, only code outside of procedures (e.g. member declarations and such) are being actually checked for compilation errors when this option is on.   As an example, the code below would not fail the CompileProject check when 'Compile On Demand' is ON, even if BlahBlahBlah is not defined:

`Option Explicit
Public Sub Foo()
   BlahBlahBlah
End Sub`

If you turn off 'Compile On Demand' then all is well, and our CompileProject test works as expected.

Slightly more concerning is that even though this is bad code, after calling CompileProject the 'Compile' option in the VBE Debug menu is then greyed out... as if a full compilation did occur.

We need to somehow enforce the 'Compile On Demand' option to OFF, at least whilst we call CompileProject.

bclothier:
Fixes #3807 

Introduces a new `VBESettings` class to provide some access to the VBE settings stored in the registry
Introduces a new `RegistryWrapper` class which aids testability for the registry operations; basically a stripped down version of the `Microsoft.Win32.Registry` static class.

Will now check that `Compile On Demand` is enabled during parsing and warn the user. Additionally, if the user enables the `Compile Before Parse` option via the general settings and the `Compile On Demand` is enabled, the user will be prompted whether they wish to update that. Unfortunately, it is necessary to restart to make the changes effective because in testing, VBE do not re-read the registry settings once it has loaded. As a consequence, a write to the registry will be inconsistent with what is shown in the Options dialog in the same session. For sake of keeping the implementation simple, I felt that was an acceptable trade-off.

Note also that `VBESettings` will attempt to write to all registry path, where they exist to keep the value consistent for different hosts. 

Furthermore, because of new dependencies added to the `GeneralSettingsViewModel` and `SettingsForm`, which required other consumers to know too much, I opted to use CW auto-magic factory for the class, to avoid forcing the other consumers to take on dependencies they do not even need.

### DIFF
--- a/RetailCoder.VBE/Rubberduck.csproj
+++ b/RetailCoder.VBE/Rubberduck.csproj
@@ -501,6 +501,8 @@
     <Compile Include="UnitTesting\Stubs\Beep.cs" />
     <Compile Include="UnitTesting\IStub.cs" />
     <Compile Include="UnitTesting\StubBase.cs" />
+    <Compile Include="VBERuntime\IVBESettings.cs" />
+    <Compile Include="VBERuntime\VBESettings.cs" />
     <Compile Include="VersionCheck\IVersionCheck.cs" />
     <Compile Include="UI\Command\MenuItems\CommandBars\AppCommandBarBase.cs" />
     <Compile Include="UI\Command\MenuItems\CommandBars\ContextSelectionLabelMenuItem.cs" />
--- a/RetailCoder.VBE/Rubberduck.csproj
+++ b/RetailCoder.VBE/Rubberduck.csproj
@@ -501,6 +501,7 @@
     <Compile Include="UnitTesting\Stubs\Beep.cs" />
     <Compile Include="UnitTesting\IStub.cs" />
     <Compile Include="UnitTesting\StubBase.cs" />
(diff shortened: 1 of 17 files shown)


bclothier:
Unfortunately, the `Compile On Demand` and related settings are not read by VBE again once it has been loaded. It is also conceivable that an user could change the setting in-session, which would be now inconsistent. The pie in sky solution would have been to be able to directly access the internal property but at this point, it is not an option. 

I'm open to any suggestions.

WaynePhillipsEA:
I think this is the only realistic solution at the moment.  Whilst obviously not ideal, it will work until I can find some time to look at alternatives.  

If we don't want to involve the user so much I guess we could just have the RD installer change the registry setting, but that seems a bit invasive.

PR #3823, opened 2018-03-04
Unify some duplicated codes and make them testable/mockable

bclothier:
This PR attempts to cut down on the duplications that was created as consequence of recent PRs, namely:

1) the VBE's version check

We introduced both `VBESettings` and `VBERuntime`; but they had their own version check logic, and were originally in two different projects. I have moved everything to `Rubberduck.VBEEditor` and separated out the version check logic.

2) the acquisition of the UI's `SynchronizationContext` used by `UiDispatcher` and `ComMessagePumper` into a new class, `UiContextProvider`.

This makes those 2 consuming classes mockable which will aid the testability of the dependent classes (though they themselves cannot be really tested). Furthermore, the `UiContextProvider` can be DI'd for any new consumers.


Rubberduck.VBEEditor/Utility/UiContext.cs
```diff
@@ -0,0 +1,56 @@
...
+                }
+            }
+        }
+
+        public static UiContextProvider Instance() => UiContextInstance;
+
+        public SynchronizationContext UiContext => Context;
+        public TaskScheduler UiTaskScheduler => TaskScheduler;
+
+        public bool CheckContext()
```
MDoerner: I would prefer a more descriptive name likem `CurrentContextIsUiContext` or `IsExecutingInUiContext`. The current name does not really convey what is checked.

PR #3824, opened 2018-03-06
Parse VBE version with invariant culture

MDoerner:
Fixes an issue with the `VBESettings` class when running with a culture that uses something else than `.` as decimal separator.


Vogel612:
There's a somewhat different fix for this very same problem [here](https://github.com/rubberduck-vba/Rubberduck/pull/3739/commits/e991c430988ab0d97d7b70f33f6b41896d110299), in PR #3739

MDoerner:
Seeing that the aother approach is actually more explicit about the truncation, I close this PR and live with the failing tests until PR #3739 got merged.

PR #3825, opened 2018-03-07
Fix for the Remove Parameter Refactoring

MDoerner:
This PR fixes three bugs in the remove parameter refactoring.

1) The refactoring was forgetting to call rewrite on the rewriters where the argument had been removed from a call site.
2) The refactoring did not recognize `whitespaceIndexExpr`, which is the grammar hack to deal with function calls with a line continuation and arbitrary further whitespace between the function name and the arguments.
3) The refactoring did not remove missing arguments from call sites.

For the third point, the resulting whitespace is not always ideal, but the VBE should do the rest.


Hosch250:
@Vogel612 https://github.com/MDoerner/Rubberduck/blob/57ebbca2869abb9d0d380959bdd578593d041f5f/RubberduckTests/Refactoring/RemoveParametersTests.cs#L171

Issue #3829, opened 2018-03-09
Keep Test Explorer Header visible as you scroll the list


IvenBach:
Currently scrolling down cuts the header off. It'd be nice to always have it visible.

IvenBach:
Works on my home machine.

I made sure to get a fresh pull from `next` and it doesn't work on my work machine. Can anyone else confirm if it works on theirs?

PR #3835, opened 2018-03-12
Freeze column header for vertical scrolling.

IvenBach:
Address Issue #3829 

F5 tested & Unit Tested.


IvenBach:
I think this should take care of the Retailcoder.VBE rename and get my PR current.

PR #3845, opened 2018-03-17
Fixed inspections and quick fixes registration in Castle Windsor

rkapka:
Apparently some PR broke the IoC registration for inspections and quick fixes. Only `ParseTreeInspections` were registered correctly. Other inspections and all quick fixes were not registered at all or registered having service type of `Object`. This PR's purpose is to restore the functionality.

### DIFF
--- a/Rubberduck.Main/Root/RubberduckIoCInstaller.cs
+++ b/Rubberduck.Main/Root/RubberduckIoCInstaller.cs
@@ -244,8 +244,7 @@ namespace Rubberduck.Root
                 container.Register(Classes.FromAssembly(assembly)
                     .IncludeNonPublicTypes()
                     .Where(type => type.IsBasedOn(typeof(IQuickFix)) && type.NotDisabledExperimental(_initialSettings))
-                    .WithService.Base()
-                    .WithService.Self()
+                    .WithService.Select(new[] {typeof(IQuickFix)})
                     .LifestyleSingleton());
             }
         }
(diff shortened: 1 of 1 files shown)


rkapka:
I'm not familiar with CW, so this fix might not be the best approach. I don't know why `Base` and `Self` services were not working.

Rubberduck.Main/Root/RubberduckIoCInstaller.cs
```diff
@@ -244,8 +244,7 @@ private void RegisterQuickFixes(IWindsorContainer container, Assembly[] assembli
                 container.Register(Classes.FromAssembly(assembly)
                     .IncludeNonPublicTypes()
                     .Where(type => type.IsBasedOn(typeof(IQuickFix)) && type.NotDisabledExperimental(_initialSettings))
-                    .WithService.Base()
-                    .WithService.Self()
```
MDoerner: The `Self()` should work and is important at the places where specific quickfixes get injected.

Rubberduck.Main/Root/RubberduckIoCInstaller.cs
```diff
@@ -269,7 +268,6 @@ private void RegisterParseTreeInspections(IWindsorContainer container, Assembly[
                 container.Register(Classes.FromAssembly(assembly)
                     .IncludeNonPublicTypes()
                     .Where(type => type.IsBasedOn(typeof(IParseTreeInspection)) && type.NotDisabledExperimental(_initialSettings))
-                    .WithService.Base()
                     .WithService.Select(new[] { typeof(IInspection) })
```
MDoerner: This has to be extended by `typeof(IParseTreeInspection)` to account for the removed `Base()`.

MDoerner:
I just looked further in the CW documentation. There was no link in the part about registration by convention but in the part about registering component-by-component.

[Conditional component registration](https://github.com/castleproject/Windsor/blob/master/docs/conditional-component-registration.md)

After reading this, I think we should go back to `BasedOn` combined with `If` or `Unless` to deal with the experimental types.

PR #3857, opened 2018-03-24
Add IFakes Registration name, fixes RunAllTestsCommand

Vogel612:


Rubberduck.Main/Root/RubberduckIoCInstaller.cs
```diff
@@ -159,6 +159,7 @@ private void RegisterUnitTestingComSide(IWindsorContainer container)
                 .LifestyleSingleton());
             container.Register(Component.For<IFakes>()
                 .ImplementedBy<FakesProvider>()
+                .Named("Fakes")
```
MDoerner: This will solve the problem at hand, but a real fix would be to rename `GetFakes` in `IFakesFactory` to something not starting with `Get`.

PR #3869, opened 2018-03-29
LocalDebugRegistryWriter now registers keys for x64 OS.

NelsonVides:
But it's not yet ready, registry redirections are not being printed back to the file. Please refer to the FIXME.

### DIFF
--- a/Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
+++ b/Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
@@ -13,6 +13,7 @@ namespace Rubberduck.Deployment.Writers
     {
         public string Write(IOrderedEnumerable<RegistryEntry> entries)
         {
+            //System.Diagnostics.Debugger.Launch(); //uncomment if need to debug
             var sb = new StringBuilder("Windows Registry Editor Version 5.00" + Environment.NewLine + Environment.NewLine);
             var distinctKeys = new List<string>();
             
@@ -30,14 +31,16 @@ namespace Rubberduck.Deployment.Writers
                     throw new InvalidOperationException("Unexpected registry entry: " + entry.Key);
                 }
 
-                var value = ReplacePlaceholder(entry.Value, entry.Bitness);
-
-                var key = Registry.CurrentUser.CreateSubKey(entry.Key);
-                if (!(string.IsNullOrWhiteSpace(entry.Name) && string.IsNullOrWhiteSpace(value)))
+                if (Environment.Is64BitOperatingSystem)
                 {
-                    key.SetValue(entry.Name, value);
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32));
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry64));
                 }
-
+                else 
+                {
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32));
+                }
+                
                 if (!distinctKeys.Contains(entry.Key))
                 {
                     distinctKeys.Add(entry.Key);
@@ -50,7 +53,19 @@ namespace Rubberduck.Deployment.Writers
                 sb.AppendLine("[-HKEY_CURRENT_USER\\" + key + "]" + Environment.NewLine);
             }
 
-            return sb.ToString();
+            return sb.ToString(); //FIXME: this is not printing reg redirections.
+        }
+
+        private void MakeRegistryEntries(RegistryEntry entry, RegistryKey hKey) 
+        {
+            var key = hKey.CreateSubKey(entry.Key);
+
+            var value = ReplacePlaceholder(entry.Value, entry.Bitness);
+
+            if (!(string.IsNullOrWhiteSpace(entry.Name) && string.IsNullOrWhiteSpace(value)))
+            {
+                key.SetValue(entry.Name, value);
+            }
         }
 
         //Cache the string so we call the AssemblyDirectory only once


Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
```diff
@@ -30,14 +31,16 @@ public string Write(IOrderedEnumerable<RegistryEntry> entries)
...
                 {
-                    key.SetValue(entry.Name, value);
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32));
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry64));
                 }
-
+                else 
+                {
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32));
+                }
```
Vogel612: Can be simplified: Both branches use `MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32))`. Accordingly we can simplify to:

```
MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry32));
if (Environment.Is64BitOperatingSystem)
{
    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry64));
}
```

NelsonVides: Pretty obvious, easy fix. Will be in the next push 👍

bclothier: What I'm not 100% sure about is whether there'll be an error on a 32-bit only system. I would prefer that for the `else` branch there be no specific view requested.

NelsonVides: @bclothier not sure what do you mean here.... `OpenBaseKey` requires a view, but I guess maybe we can just say take `RegistryView.Default` (?)

bclothier: No, just pass in a `Registry.CurrentUser` without a call to the `OpenBaseKey` at all in the `else`. That makes it implicit.

Rubberduck.Deployment/BuildRegistryScript.ps1
```diff
@@ -95,9 +95,17 @@ try
...
 					$datetime = Get-Date;
-					& reg.exe import $regFile;
+					if ([Environment]::Is64BitOperatingSystem)
+					{
+						& reg.exe import $regFile /reg:32;
+						& reg.exe import $regFile /reg:64;
+					}
+					else 
+					{
+						& reg.exe import $regFile /reg:32;
```
bclothier: I would not pass an explicit `/reg:32` flag here. Just remove it and let it be implicit in this case.

Rubberduck.Deployment/BuildRegistryScript.ps1
```diff
@@ -95,9 +95,17 @@ try
...
+						& reg.exe import $regFile /reg:64;
+					}
+					else 
+					{
+						& reg.exe import $regFile /reg:32;					
+					}
 					& reg.exe import ($dir + "\RubberduckAddinRegistry.reg");
-					Move-Item -Path $regFile -Destination ($regFile + ".imported_" + $datetime.ToUniversalTime().ToString("yyyyMMddhhmmss") + ".txt" )
+					Move-Item -Path $regFile -Destination ($regFile + ".imported_" + $datetime.ToUniversalTime().ToString("yyyyMMddHHmmss") + ".txt" )
```
NelsonVides: Small thing, hope you don't mind. Having hours set to am/pm was disordering the files if ordered by name. Also, European habit, I'm so used to my 24h timing.

Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
```diff
@@ -30,14 +31,12 @@ public string Write(IOrderedEnumerable<RegistryEntry> entries)
                     throw new InvalidOperationException("Unexpected registry entry: " + entry.Key);
                 }
 
-                var value = ReplacePlaceholder(entry.Value, entry.Bitness);
-
-                var key = Registry.CurrentUser.CreateSubKey(entry.Key);
-                if (!(string.IsNullOrWhiteSpace(entry.Name) && string.IsNullOrWhiteSpace(value)))
+                MakeRegistryEntries(entry, Registry.CurrentUser);
```
bclothier: The concern is that this presumes that the powershell script will run in 32-bit mode all the time. That might not be the case because if the script is run in 64-bit Powershell, the keys end up being written to same path twice. 

It would be more appropriate, IMO, to only explicitly write to both 64-bit and 32-bit if the environment is a 64-bit and otherwise use implicit view. (e.g. reverse the condition and statements)

Rubberduck.Deployment/BuildRegistryScript.ps1
bclothier: Oversight on my part. Good catch - yes I meant to use 24 hours clock.

Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
NelsonVides: Writing twice the same key, yes, now that you mention it that will happen. But also, this powershell is always running within VS build, which is only 32bits (there's no 64bits VS), so I guess we can assume it will always run on 32 (unless Microsoft surprises us with a 64bits VS) (?)
Anyway, let's put it back to the else branch, I also don't like writing to the same key twice.

bclothier: That is true, but I don't like writing code with assumptions that may be one day wrong next year, and especially because it can be very easy to miss the change. 

```
Belts++;
Suspenders++;
```

Rubberduck.Deployment/Writers/LocalDebugRegistryWriter.cs
```diff
@@ -30,12 +31,13 @@ public string Write(IOrderedEnumerable<RegistryEntry> entries)
...
-                var key = Registry.CurrentUser.CreateSubKey(entry.Key);
-                if (!(string.IsNullOrWhiteSpace(entry.Name) && string.IsNullOrWhiteSpace(value)))
+                if (Environment.Is64BitOperatingSystem)
+                {
+                    MakeRegistryEntries(entry, RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry64));
+                }
+                else 
                 {
-                    key.SetValue(entry.Name, value);
+                    MakeRegistryEntries(entry, Registry.CurrentUser);
```
bclothier: Not quite; we want to write to *both* 64-bit and 32-bit view on a 64-bit system. Because this is now in a `else` branch, it means only one or other are written to. Put it outside the `else` so that the 32-bit is always written to.

NelsonVides: I don't get it. Outside the else is where it was before. I think you mean something like, in the else then we call `MakeRegistryEntries` with implicit parameter, in the if, we call if twice with both RegistryViews explicitly. Right?

PR #3875, opened 2018-04-01
Correct filtering behaviour in CodeInspection Settings

Vogel612:
Additionally add i18n for the filtering feature and slightly change the UI


Vogel612:
I'm collapsing this into #3876

PR #3888, opened 2018-04-06
Fix unit testing

### LINKED ISSUE #3809: Unit testing issues
In unit test modules, when the VBE 'Compile On Demand' option is OFF (and probably when ON in some cases), and there is a compile time error that makes ModuleInitialize uncallable, it will throw an error in the TypeLib API ExecuteCode which is being thrown right back to the RunAllTestsCommand.   No visual feedback is given as to the problem, and the remaining test results are not updated.

I believe this code in the UnitTesting.TestEngine needs exception handling to catch and update the test results:

     Run(module.Key.FindModuleInitializeMethods(_state));

Additionally, it appears that the 'Compile Before Parse' option isn't effective for the automatic parse that is triggered from running tests.

bclothier:
Closes #3809 

The unit tests now will run after a parse, rather than during which helps avoid errors due to running tests and parsing concurrently. 

Updated the test engine to listen to the event. I had to add a flag to indicate whether to run the tests.

I also fixed the issue with assembly because `Assembly.GetEntryAssembly` was returning null, and the code made wrong assumption about the name of the library. We avoid this by doing a typelib registry lookup instead. That neatly solves the problem without needing to discover the `Main` at all and more importantly, give us an opportunity to warn of a problem with early binding --- if we can't locate the typelib, then early binding is not going to work at all.

### DIFF
--- a/Rubberduck.Core/UnitTesting/TestEngine.cs
+++ b/Rubberduck.Core/UnitTesting/TestEngine.cs
@@ -52,6 +52,13 @@ namespace Rubberduck.UnitTesting
                     RunInternal(_tests);
                 });
             }
+
+            if (_testRequested && (e.State == ParserState.ResolverError || 
+                                   e.State == ParserState.Error ||
+                                   e.State == ParserState.UnexpectedError))
+            {
+                _testRequested = false;
+            }
         }
 
         public TestExplorerModel Model { get; }
--- a/Rubberduck.Core/UnitTesting/TestEngine.cs
+++ b/Rubberduck.Core/UnitTesting/TestEngine.cs
@@ -6,6 +6,7 @@ using System.Runtime.InteropServices;
 using NLog;
 using Rubberduck.Parsing.Annotations;
 using Rubberduck.Parsing.Symbols;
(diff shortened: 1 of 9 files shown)


Rubberduck.Core/UnitTesting/TestEngine.cs
```diff
@@ -21,17 +22,43 @@ public class TestEngine : ITestEngine
...
+                _testRequested = false;
+                _uiDispatcher.InvokeAsync(() =>
+                {
+                    RunInternal(_tests);
+                });
+            }
+
+            if (_testRequested && (e.State == ParserState.ResolverError || 
+                                   e.State == ParserState.Error ||
+                                   e.State == ParserState.UnexpectedError))
```
Vogel612: we should be able to unify that check into a single `e.State.IsError()`. That can be used across the codebase for consistent handling of all error states, even those we introduce in the future...

bclothier: Good idea. I made it `e.IsError`, though, since `e.State` is an enum.

PR #3923, opened 2018-04-11
Bring CommandBarButton events under control

bclothier:
The PR applies the `SafeEventedComWrapper` to the `CommandBarButton` which supports a single `Click` event.

Note that this created a diamond problem because `CommandBarButton` derives from `CommandBarBase` which in turns derives from `SafeComWrapper` and we need to derive from `SafeEventedComWrapper` instead. For now, I've chosen to (with apologies) copy-pasta the code and implement `SafeEventedWrapper`. A better long-term fix would be to have a private `CommandBarBase` field to pass all non-overriden methods to. 

The real focus of the PR, however, is to make the event management of the `Click` event deterministic and can be controlled via the `VBEEvents` indirectly. 

The button creation is done via the `CommandBarButtonFactory`, which originally was a static class but is now DI'd. The factory will maintain a list of buttons and listen to the `VBEEvents.EventTerminated` event to detach all buttons' events in a timely manner. Note that it does not dispose the buttons as we can't assume that it's appropriate to do so but we do ensure that the events are detached, making buttons useless. 

Note that this does not fix the unclean exit issue.


WaynePhillipsEA:
Haven't had chance to review the code, but does this PR also ensure that the events are unregistered before the objects in question are deleted at shutdown?  (with the Delete method)

bclothier:
It will call `DetachEvents` as part of the `Delete` method ([ref](https://github.com/rubberduck-vba/Rubberduck/pull/3923/files#diff-00356a00afa912bc19f3b58beacb1ff9R336)). 

Note that `DetachEvents` is also called on the `Dispose()` and also in response to `VBEEvents.EventsTerminated` event via the factory.

bclothier:
With the commit, there should be no exceptions thrown as result of tearing down the command bars which was introduced by my prior commits. I also had to add caching for the `GetHashCode` because the calculation depends on `Target.Parent` which cause an exception at shutdown so that i can track buttons even after they've been torn down since we don't necessarily want to dispose of them when we are detaching the events. 

Despite not throwing any exceptions, the exit hanging remains, however.

WaynePhillipsEA:
If accessing the Parent was causing an access violation, that's a red flag... something is not right.   I would guess the order of destruction (... it suggests child elements not being deleted and released before their parent objects get the same treatment)

WaynePhillipsEA:
^ can probably ignore, as it looks like the Parent member was simply being accessed after the deletion for the purpose of the hashcode to delete from the collection (so any member access would have failed there, not just Parent).

WaynePhillipsEA:
If this gets merged, I will grab a new log at the weekend to ensure this fixes the leaking commandbars.

PR #3930, opened 2018-04-13
Inject a concrete factory for the FakesProvider

MDoerner:
This fixed part of the shutdown issue for me. However, there are two caveats.

1. I had to reset my settings in order to get a clean exit. So, there is probably something wrong with the new default settings loading, which did not work for me anyways. (I could not find the Case not reachable inspection in the inspections settings.)
2. After changing the inspections severity for one inspection in the settings window, I got an eternally hanging processes on shutdown in the debugger; no parse required. With this I mean that I had to kill VS to make it respond again to anything.  

The two points above require further investigation, but I think this PR is a step in the right direction.

### DIFF
--- a/Rubberduck.Core/UnitTesting/TestEngine.cs
+++ b/Rubberduck.Core/UnitTesting/TestEngine.cs
@@ -107,51 +107,44 @@ namespace Rubberduck.UnitTesting
                                 && test.Declaration.QualifiedName.QualifiedModuleName.ComponentName == capturedModule.Key.ComponentName);
 
                 var fakes = _fakesFactory.Create();
-                try
+                Run(module.Key.FindModuleInitializeMethods(_state));
+                foreach (var test in moduleTestMethods)
                 {
-                    Run(module.Key.FindModuleInitializeMethods(_state));
-                    foreach (var test in moduleTestMethods)
+                    // no need to run setup/teardown for ignored tests
+                    if (test.Declaration.Annotations.Any(a => a.AnnotationType == AnnotationType.IgnoreTest))
                     {
-                        // no need to run setup/teardown for ignored tests
-                        if (test.Declaration.Annotations.Any(a => a.AnnotationType == AnnotationType.IgnoreTest))
-                        {
-                            test.UpdateResult(TestOutcome.Ignored);
-                            OnTestCompleted();
-                            continue;
-                        }
-
-                        var stopwatch = new Stopwatch();
-                        stopwatch.Start();
-
-                        try
-                        {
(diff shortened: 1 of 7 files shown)


Rubberduck.Core/UnitTesting/TestEngine.cs
```diff
@@ -107,51 +107,44 @@ private void RunInternal(IEnumerable<TestMethod> tests)
                                 && test.Declaration.QualifiedName.QualifiedModuleName.ComponentName == capturedModule.Key.ComponentName);
 
                 var fakes = _fakesFactory.Create();
-                try
+                Run(module.Key.FindModuleInitializeMethods(_state));
```
bclothier: Shouldn't all `Run` methods be inside a `try` block? Or is it not needed here if we are simply updating the ignored tests' result?

Rubberduck.Core/UnitTesting/TestEngine.cs
```diff
@@ -107,51 +107,44 @@ private void RunInternal(IEnumerable<TestMethod> tests)
...
+                        fakes.StopTest();
+                    }
+
+                    stopwatch.Stop();
+                    test.Result.SetDuration(stopwatch.ElapsedMilliseconds);
+
+                    OnTestCompleted();
+                    Model.AddExecutedTest(test);
                 }
+                Run(module.Key.FindModuleCleanupMethods(_state));
```
bclothier: Same comment RE: `Run` inside a `try` block.

Rubberduck.Core/UnitTesting/TestEngine.cs
bclothier: Actually, I also realize this is inside the `RunInteral` and we are calling `Run`. I do not think that is right and I should have caught that earlier. I think all `Run` methods within the `RunInternal` needs to be updated to avoid going through the `UiDispatcher.InvokeAsync` again?

MDoerner:
Regarding not disposing of `fakes`, please note that `IFakes` does not extend `IDisposable`. So, there is nothing to dispose.

bclothier:
But `FakeBase` class which implements `IFake` is based on `StubBase` which implements `IDisposable`?

Rubberduck.Core/UnitTesting/TestEngine.cs
MDoerner: The code is rather confusing, but the calls to `Run` inside `RunInternal` use the overload taking `IEnumerable<Declaration>`, which you find below `RunInternal`. They are not going through the `UiDispatcher`.

PR #3935, opened 2018-04-15
VB6 IDE Support

### LINKED ISSUE #298: VB6 IDE support
Is this add-in supposed to work in VB6 IDE? 

I think ANTLR parser has great potential for really useful features in the IDE. RegExp find&replace within routines with {Name} placeholder comes to mind.

mansellan:
Ran all tests locally and F5'd in Excel and VB6.

My goals for this PR were (in order):

1. Support VB6 (obvs!)
2. Segregate COM wrapper implementations into new projects (Rubberduck.VBEditor.VBA and Rubberduck.VBEditor.VB6) and clean up filesystem accordingly.
3. Eliminate all COM interop references from all projects except the two new ones above.
4. Reduce namespace proliferation under Rubberduck.VBEditor.SafeComWrappers
5. Minimise disturbance to any other projects

Actions taken:

> Removed all VB and Office interop dependencies from all projects except Rubberduck.VBEditor.VBA and Rubberduck.VBEditor.VB6

> Moved initial resolution of VBE and AddIn wrappers (which happens before IoC registration) into Rubberduck.Core.RootComWrapperFactory
>> Allows both API and Add-in entry points to share
>> Decouples Rubberduck.Main from interop assemblies

> Removed CommandBarButtonFactory and CommandBarPopupFactory - buttons are now added through methods on their collection wrappers

> Rationalised namespaces for SCWs down to:
>> Rubberduck.VBEditor.SafeComWrappers
>> Rubberduck.VBEditor.SafeComWrappers.Abstract
>> Rubberduck.VBEditor.SafeComWrappers.Office
>> Rubberduck.VBEditor.SafeComWrappers.Office.Abstract

And (in VBA project only)
>> Rubberduck.VBEditor.SafeComWrappers.VBA
>> Rubberduck.VBEditor.SafeComWrappers.Office12

And (in VB6 project only)
>> Rubberduck.VBEditor.SafeComWrappers.VB6
>> Rubberduck.VBEditor.SafeComWrappers.Office8

> Dropped IVBComponents.AddMTDesigner, ICommandBar.IsPriorityDropped
>> Undocumented, unused and unwanted.

> Changed ReleaseDockableHosts into an instance method to remove interop dependency


Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -0,0 +1,34 @@
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+using VB = Microsoft.Vbe.Interop.VB6;
+
+namespace Rubberduck.VBEditor.SafeComWrappers.VB6
```
bclothier: The namespace doesn't match the folder -- folder is `VB`. Shouldn't it be `VB6`?

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControls.cs
```diff
@@ -0,0 +1,47 @@
+using System.Collections;
+using System.Collections.Generic;
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+using VB = Microsoft.Vbe.Interop.VB6;
+
+namespace Rubberduck.VBEditor.SafeComWrappers.VB6
```
bclothier: Ditto RE: folder/namespace

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBE.cs
```diff
@@ -167,6 +120,71 @@ public int ProjectsCount
...
+            var sameName = c.Name == name;
+            if (!sameName)
+            {
+                c.Dispose();
+            }
+            return sameName;
+        }
+
+
+        public IHostApplication HostApplication()
```
bclothier: if we're getting rid of `IHostApplication`, why is this still here?

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBForm.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+using VB = Microsoft.Vbe.Interop.VB6;
+
+namespace Rubberduck.VBEditor.SafeComWrappers.VB6
```
bclothier: Ditto RE: namespace/folder

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -13,39 +13,54 @@ public VBProject(VB.VBProject target, bool rewrapping = false)
...
         }
 
-        public EnvironmentMode Mode => throw new NotImplementedException();
+        public EnvironmentMode Mode
+        {
+            get
+            {
+                if (IsWrappingNullReference) {return 0; }
+
+                //// Yes, this is a hack. Go ask Microsoft why they created a Mode enum and then didn't use it.
```
bclothier: Can't we use the function suggested in #298 ? [ref to comment](https://github.com/rubberduck-vba/Rubberduck/issues/298#issuecomment-338431949)

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProjects.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using Rubberduck.VBEditor.Events;
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+using VB = Microsoft.Vbe.Interop.VB6;
+
+namespace Rubberduck.VBEditor.SafeComWrappers.VB6
```
bclothier: Ditto RE: namespace/folder

Rubberduck.VBEditor.VBA/SafeComWrappers/Office/CommandBarPopup.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+using MSO = Microsoft.Office.Core;
+
+namespace Rubberduck.VBEditor.SafeComWrappers.Office12
```
bclothier: Ditto RE: namespace/folder

Rubberduck.VBEditor.VBA/SafeComWrappers/Office/CommandBars.cs
```diff
@@ -4,14 +4,13 @@
 using System.Linq;
 using System.Runtime.InteropServices;
 using Rubberduck.VBEditor.SafeComWrappers.Abstract;
-using Rubberduck.VBEditor.SafeComWrappers.MSForms;
-using Rubberduck.VBEditor.SafeComWrappers.Office.Core.Abstract;
+using MSO = Microsoft.Office.Core;
 
-namespace Rubberduck.VBEditor.SafeComWrappers.Office.Core
+namespace Rubberduck.VBEditor.SafeComWrappers.Office12
```
bclothier: folder/namespace

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBComponents.cs
```diff
@@ -183,22 +178,19 @@ public void RemoveSafely(IVBComponent component)
         }
 
         #region Events
-        
-        private delegate void ItemAddedDelegate(VB.VBComponent vbComponent);
```
bclothier: Nice catch!

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/Windows.cs
```diff
@@ -35,7 +35,7 @@ public ToolWindowInfo CreateToolWindow(IAddIn addInInst, string progId, string c
             return new ToolWindowInfo(new Window(window), control);
         }
 
-        public static void ReleaseDockableHosts()
+        public void ReleaseDockableHosts()
```
bclothier: Can you elaborate the reasoning why this was made into an instance method?

Rubberduck.API/API/VBA/ParserState.cs
```diff
@@ -61,14 +63,16 @@ public ParserState()
             _dispatcher = new UiDispatcher(UiContextProvider.Instance());
         }
 
-        public void Initialize(Microsoft.Vbe.Interop.VBE vbe)
+        // vbe is the com coclass interface from the interop assembly.
+        // There is no shared interface between VBA and VB6 types, hence object.
+        public void Initialize(object vbe)
```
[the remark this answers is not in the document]
MDoerner: We do not expose `IVBE` to COM interop. So the COM side, which is using this interface, has no idea what `IVBE` is.

Rubberduck.Core/Common/KeyHookEventArgs.cs
```diff
@@ -9,13 +9,13 @@ namespace Rubberduck.Common
     /// </summary>
     public class KeyHookEventArgs : EventArgs
     {
-        public KeyHookEventArgs(Keys key, VBComponent component)
+        public KeyHookEventArgs(Keys key, IVBComponent component)
```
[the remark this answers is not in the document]
MDoerner: This might actually help with the shutdown problems. It looks as if this has been using an RCW directly.

Rubberduck.VBEEditor/Extensions/IDEExtensions.cs
```diff
@@ -2,144 +2,12 @@
...
-            {"WINPROJ.EXE", typeof(ProjectApp)},
-            {"MSPUB.EXE", typeof(PublisherApp)},
-            {"VISIO.EXE", typeof(VisioApp)},
-            {"ACAD.EXE", typeof(AutoCADApp)},
-            {"CORELDRW.EXE", typeof(CorelDRAWApp)},
-            {"SLDWORKS.EXE", typeof(SolidWorksApp)},
-        };
-
-        /// <summary> Returns the type of Office Application that is hosting the VBE. </summary>
-        public static IHostApplication HostApplication(this IVBE vbe)
```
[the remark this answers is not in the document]
mansellan: Perhaps. I just moved what we had for now. Although we don't need IHostApplication for test invocation any more, IIRC we have previously talked of having host-specifics for things such as custom document types. So it makes sense to keep these until those decisions are finalised. If they are to go, we should raise a new issue.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
mansellan: Interesting point. There is a related discussion in Framework Design Guidelines (Cwalina and Abrams) on namespaces and folders. I don't have my copy to hand atm, but IIRC it makes the point that namespaces are primarily for type discoverability, whereas folders are for file organisation. They need not follow each other exactly, and typically the file structure would be more subdivided than the namespace.

Happy to take comments though!

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
mansellan: Busted! I was being lazy... will take a look.

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/Windows.cs
mansellan: Used in the shutdown process in Extension.cs in Rubberduck.Main. As a static method, this needed a hard reference to the (VBA) SCW assembly. As an interface method, it can be exposed on the IWindow interface, removing the hard dependency.

Rubberduck.Parsing/Symbols/Declaration.cs
```diff
@@ -575,7 +575,7 @@ public override int GetHashCode()
             {
                 var hash = 17;
                 hash = hash * 23 + QualifiedName.QualifiedModuleName.GetHashCode();
-                hash = hash * 23 + IdentifierName.GetHashCode();
+                hash = hash * 23 + IdentifierName?.GetHashCode() ?? 0;  // VB6 has no IdentifierName
```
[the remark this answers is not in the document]
mansellan: I'll take another look to see if an analog is available

retailcoder:
Some very good stuff in here that I'd like to merge asap! PR is still WIP, right?

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -127,7 +126,7 @@ private List<Action> CollectionConstructionActions(IReadOnlyList<Declaration> de
                 );
             actions.Add(() => 
                 _declarationsByName = declarations
-                    .GroupBy(declaration => declaration.IdentifierName.ToLowerInvariant())
+                    .GroupBy(declaration => declaration.IdentifierName?.ToLowerInvariant() ?? string.Empty) // VB6 has no IdentifierName
```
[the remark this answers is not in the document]
retailcoder: @MDoerner isn't `VBA` the identifier we use for the VBA Standard Library "project"? In that case then shouldn't the VB6 standard library "just work", off whatever identifier the type library defines? (can't remember, is that `VB`?)

Rubberduck.Main/Extension.cs
```diff
@@ -251,7 +239,7 @@ private void ShutdownAddIn()
                 VBENativeServices.UnhookEvents();
 
                 _logger.Log(LogLevel.Trace, "Releasing dockable hosts...");
-                Windows.ReleaseDockableHosts();
+                _vbe.Windows.ReleaseDockableHosts();
```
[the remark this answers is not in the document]
bclothier: Wouldn't this be actually two using; one for `_vbe` then for the `Windows`?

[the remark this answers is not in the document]
retailcoder: @bclothier we quite probably don't want to dispose the `_vbe` field right there & then.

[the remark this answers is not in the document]
bclothier: You're right. I was thinking it was injected but that's not the case here. #Readfail.

Rubberduck.VBEEditor/WindowsApi/ChildWindowFinder.cs
```diff
@@ -2,11 +2,11 @@
 
 namespace Rubberduck.VBEditor.WindowsApi
 {
-    internal class ChildWindowFinder
+    public class ChildWindowFinder
```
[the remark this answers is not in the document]
mansellan: Seems to be used by the ForceFocus method in the CodePane SCW. Change is so as to allow use from both VB6 and VBA SCW assemblies.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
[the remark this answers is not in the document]
mansellan: Hmm... I did this back in the day and forgot why. Just tried removal, got some funky OLE fails on parse on an empty form. QMN has this:

image

Guessing it should be solved somewhere upstream, but not sure where...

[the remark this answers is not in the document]
MDoerner: This is probably a problem with the `ReferencedDeclarationsCollector`. Can you tell me the type of the declaration, either the com type or the `DeclarationType`? I will have a look what is going on here.

Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
```diff
@@ -0,0 +1,81 @@
...
+                    return;
+                }
+
+                if (_cookie != NotAdvising)
+                {
+                    return;
+                }
+
+                // Call QueryInterface for IConnectionPointContainer
+                var icpc = (IConnectionPointContainer) _eventSource;    // TODO - HELP!!! This is gonna leak like a sieve :-(
```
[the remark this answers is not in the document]
mansellan: Not sure how to go about that - this is the RCW type being passed (in this case, an instance implementing Microsoft.Vbe.VB6.CommandBarEvents). Any pointers?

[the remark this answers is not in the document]
mansellan: (excuse the pun!)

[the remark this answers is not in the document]
bclothier: Right - know what, we can simplify there. We don't need to wrap this. We get this in the ctor, so we can dispose it ourselves. Just override the `Dispose` (note all SCW implement `IDisposable`), call `Marshal.ReleaseComObject(_eventsource)` then call `base.Dispose()`.

Rubberduck.VBEEditor/Extensions/IDEExtensions.cs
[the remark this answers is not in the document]
mansellan: Added directives to suppress R# warnings.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
mansellan: Discussed and agreed in SE - added R# directives to suppress inspection warnings.

Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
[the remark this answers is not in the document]
mansellan: thanks :-) will do.

Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
```diff
@@ -23,6 +23,7 @@ protected SafeRedirectedEventedComWrapper(TSource target, TEventSource eventSour
         protected override void Dispose(bool disposing)
         {
             DetachEvents();
+            Marshal.ReleaseComObject(_eventSource);
```
bclothier: I don't think that's the right place. It should be inside the `DetachEvents`. Otherwise, you end up leaking when multiple `AttachEvents`/`DetachEvents` calls are made to the evented wrapper. The `Dispose` will call the `DetachEvents` so there should be no leak that way.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -202,5 +197,9 @@ public string ProjectDisplayName
                 }
             }
         }
+
+        private const string DllName = "vba6.dll";
```
bclothier: Two comments:

1) I prefer to keep externs separated from the implementation but that can be dealt with in a separate PR.

2) Are we sure VB6 actually uses this function from VB**A**6 DLL? I would have thought it'd be `msvbvm60.dll` or something similar.

Rubberduck.VBEditor.VB6/SafeComWrappers/Office/CommandBarButton.cs
```diff
@@ -0,0 +1,343 @@
...
+    {
+        private readonly CommandBarControl _control;
+        private readonly IVBE _vbe;
+        private static readonly Logger _logger = LogManager.GetCurrentClassLogger();
+        public const bool AddCommandBarControlsTemporarily = false;        
+
+        // Command bar click event is sourced from VBE.Events.CommandBarEvents[index]
+        // where index is the command bar button COM object.
+        public CommandBarButton(MSO.CommandBarButton target, IVBE vbe, bool rewrapping = false) 
+            : base(target, ((VB.VBE)vbe.HardReference).Events.CommandBarEvents[target], rewrapping)
```
MDoerner: This will certainly leak RCWs. You cannot chain member accesses on COM interop objects without leaking all intermediate objects. This is why we wrap the result of each member call in the COM wrappers.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/LinkedWindows.cs
```diff
@@ -18,28 +19,38 @@ public LinkedWindows(VB.LinkedWindows target, bool rewrapping = false)
...
+                return;
+            }
+
             Target.Add(((Window)window).Target);
         }
 
         IEnumerator IEnumerable.GetEnumerator()
         {
-            return Target.GetEnumerator();
+            return IsWrappingNullReference ? new List<IEnumerable>().GetEnumerator() : Target.GetEnumerator();
```
MDoerner: Same comment about enumerator.
Btw, `Properties` below is doing it right.

Rubberduck.VBEditor.VB6/SafeComWrappers/Office/CommandBarControls.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return IsWrappingNullReference
+                ? new List<ICommandBarControl>().GetEnumerator()
+                : ((IEnumerable<ICommandBarControl>) this).GetEnumerator();
+        }
+
+        //public override void Release(bool final = false)
```
MDoerner: Can be removed.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBComponents.cs
```diff
@@ -0,0 +1,255 @@
...
+                    }
+                    catch
+                    {
+                        component = Import(path);
+                    }
+
+                    var codeString =
+                        File.ReadAllText(path,
+                            Encoding
+                                .Default); //The VBE uses the current ANSI codepage from the windows settings to export and import.
```
MDoerner: The line break here is strange.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class VBControl : SafeComWrapper<VB.VBControl>, IControl
+    {
+        public VBControl(VB.VBControl target, bool rewrapping = false) 
+            : base(target, rewrapping)
+        {
+        }
+
+        public string Name
+        {
+            get => IsWrappingNullReference ? string.Empty : Target.Properties.Item("Name").Value.ToString();
```
MDoerner: chained COM access

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -0,0 +1,35 @@
...
+            set { if (!IsWrappingNullReference) Target.Properties.Item("Name").Value = value; }
```
MDoerner: dito

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBE.cs
```diff
@@ -167,6 +121,71 @@ public int ProjectsCount
             }
         }
 
+        public static void SetSelection(IVBProject vbProject, Selection selection, string name)
+        {
+            using (var components = vbProject.VBComponents)
+            {
+                using (var component = components.SingleOrDefault(c => ComponentHasName(c, name))) 
+                {
+            if (component == null || component.IsWrappingNullReference)
```
MDoerner: indentation is off

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBE.cs
```diff
@@ -167,6 +121,71 @@ public int ProjectsCount
...
+            {
+                return;
+            }
+
+                    using (var module = component.CodeModule)
+                    {
+                        using (var pane = module.CodePane)
+                        {
+            pane.Selection = selection;
```
MDoerner: dito

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -13,39 +15,47 @@ public VBProject(VB.VBProject target, bool rewrapping = false)
...
-            set => Target.Name = value;
+            set { if (!IsWrappingNullReference) Target.Name = value; }
         }
 
-        public EnvironmentMode Mode => throw new NotImplementedException();
+        public EnvironmentMode Mode
+        {
+            get
+            {
+                if (IsWrappingNullReference) {return 0; }
```
MDoerner: As a general rule, we prefer to have the scope on separate lines so that the indentation matches the scope.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/Windows.cs
```diff
@@ -0,0 +1,75 @@
...
+        public int Count => IsWrappingNullReference ? 0 : Target.Count;
+
+        public IVBE VBE => new VBE(IsWrappingNullReference ? null : Target.VBE);
+
+        public IApplication Parent => new Application(null); // VB6 has no Application
+
+        public IWindow this[object index] => new Window(IsWrappingNullReference ? null : Target.Item(index));
+
+
+        private static readonly Dictionary<VB.Window, object> _dockableHosts = new Dictionary<VB.Window, object>();
```
MDoerner: As a note for further PRs, we should see whether we can make that a dictionary over `IWindow`, maybe using the `ReferenceEqualityComparer` to ignore any custom implementations of `GetHashCode` and `Equals`.
The same applies to the VBA version.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/Windows.cs
```diff
@@ -0,0 +1,75 @@
...
+                item.Key.Close();
+                dynamic host = item.Value;
+                host.Release();
+            }
+            _dockableHosts.Clear();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return IsWrappingNullReference ? new List<IEnumerable>().GetEnumerator() : Target.GetEnumerator();
```
MDoerner: non-generic enumerator again

Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
```diff
@@ -0,0 +1,84 @@
...
+                }
+
+                if (_cookie != NotAdvising)
+                {
+                    _icp.Unadvise(_cookie);
+                    _cookie = NotAdvising;
+                }
+
+                Marshal.ReleaseComObject(_icp);
+                Marshal.ReleaseComObject(_eventSource);
```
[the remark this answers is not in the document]
bclothier: That would be my bad since I was one who suggested to do that and I should have known better. I was thinking it was used only once but reading the code again, Max is correct - `AttachEvents` gets used, so it would be premature. The version prior (where we dispose of it) would be acceptable. Sorry!

Rubberduck.VBEditor.VB6/SafeComWrappers/Office/CommandBars.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            DeleteExistingCommandBar(name);
+            return new CommandBar(IsWrappingNullReference ? null : Target.Add(name, position, Temporary: true), _vbe);
+        }
+
+        private void DeleteExistingCommandBar(string name)
+        {
+            try
+            {
+                var existing = Target.Cast<MSO.CommandBar>().FirstOrDefault(bar => bar.Name == name);
```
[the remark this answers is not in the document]
mansellan: Would this work?

`this.FirstOrDefault(bar => bar.Name == name)?.Delete();`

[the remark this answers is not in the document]
retailcoder: I would avoid LINQ code in the wrappers.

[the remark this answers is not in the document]
mansellan: ok nw - but is the underlying idea of using the (wrapped) enumerable itself, and deleting through the child wrapper sound?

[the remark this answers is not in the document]
mansellan: i.e.

    private void DeleteExistingCommandBar(string name)
        {
            foreach (var commandBar in (IEnumerable<ICommandBar>) this)
            {
                if (commandBar.Name == name)
                {
                    commandBar.Delete();
                }
            }
     }
`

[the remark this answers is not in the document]
MDoerner: Now, you just have to add disposal for the wrappers you create in the loop.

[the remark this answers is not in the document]
mansellan: #TIL ;-)

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -11,10 +12,54 @@ public VBControl(VB.VBControl target, bool rewrapping = false)
...
+                    return string.Empty;
+                }
+
+                using (var properties = this.Properties)
+                {
+                    foreach (var property in properties)
+                    {
+                        using (property)
+                        {
+                            if (property.Name == "Name")
```
bclothier: @MDoerner is it necessary to do a `foreach` to locate a specific thing here? I think it should not leak if one does  this way:
```
using (var properties = this.Properties;
using (var property = properties.Item("Name");
{
  return property.value as string;
}
```

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -11,10 +12,54 @@ public VBControl(VB.VBControl target, bool rewrapping = false)
...
+                            {
+                                return (string)property.Value;
```
bclothier: This line may throw exception. If possible, I'd use `return property.Value as string` which will simply return `null` when it cannot be cast to string. Otherwise, you need to wrap a `try`/`catch` to guard for potential exception.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
```diff
@@ -11,10 +12,54 @@ public VBControl(VB.VBControl target, bool rewrapping = false)
...
+                            }
+                        }
+                    }
+                }
+                // Should never happen - all VB controls are required to be named.
+                return null;
```
bclothier: Add a `Debug.Assert` here?

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBControl.cs
MDoerner: This seems reasonable, but it will require a try catch block because the properties collection will throw when asked for a non-existent key.

Rubberduck.VBEditor.VB6/SafeComWrappers/Office/CommandBarButton.cs
mansellan: Raised #3949

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/Windows.cs
mansellan: #3957

Hosch250:
@MDoerner Think you could confirm that your concerns were addressed? Might as well get this in and start testing it if so.

PR #3945, opened 2018-04-18
Parsing vb forms

MDoerner:
This PR changes the parser and lexer to enable us to parse VB6 forms.

If there are more constructs that have to be added, we can add them once we find an example we cannot parse.

### DIFF
--- a/Rubberduck.Parsing/Grammar/VBAParser.g4
+++ b/Rubberduck.Parsing/Grammar/VBAParser.g4
@@ -30,7 +30,7 @@ module :
     moduleAttributes
     moduleHeader?
     moduleAttributes
-	moduleConfigReferences?
+    moduleConfigReferences?
     moduleConfig?
     moduleAttributes
     moduleDeclarations
--- a/Rubberduck.Parsing/Grammar/VBAParser.g4
+++ b/Rubberduck.Parsing/Grammar/VBAParser.g4
@@ -44,12 +44,18 @@ moduleHeader : VERSION whiteSpace numberLiteral whiteSpace? CLASS? endOfStatemen
 
 moduleConfig :
     BEGIN (whiteSpace (GUIDLITERAL | expression) whiteSpace unrestrictedIdentifier whiteSpace?)? endOfStatement
-    (moduleConfig | moduleConfigElement)+
+        (moduleConfig | moduleConfigProperty | moduleConfigElement)+
     END endOfStatement
 ;
 
+moduleConfigProperty :
+    BEGINPROPERTY whiteSpace unrestrictedIdentifier (whiteSpace GUIDLITERAL)? endOfStatement
+        (moduleConfigProperty | moduleConfigElement)+
+    ENDPROPERTY endOfStatement
+;
+
 moduleConfigElement :
-    unrestrictedIdentifier whiteSpace* EQ whiteSpace* expression (COLON numberLiteral)? endOfStatement
+    unrestrictedIdentifier whiteSpace? EQ whiteSpace? expression (COLON numberLiteral)? endOfStatement
 ;
 
 moduleAttributes : (attributeStmt endOfStatement)*;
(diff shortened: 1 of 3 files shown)


Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -30,6 +30,7 @@ module :
     moduleAttributes
     moduleHeader?
     moduleAttributes
+	moduleConfigReferences?
```
bclothier: tab/spaces?

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -687,6 +698,7 @@ keyword :
      | ARRAY
      | ATTRIBUTE
      | BEGIN
+	 | BEGINPROPERTY
```
bclothier: tab/spaces?

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -713,6 +725,7 @@ keyword :
      | DOEVENTS
      | DOUBLE
      | END
+	 | ENDPROPERTY
```
bclothier: tab/spaces?

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -42,14 +43,24 @@ module :
 
 moduleHeader : VERSION whiteSpace numberLiteral whiteSpace? CLASS? endOfStatement;
 
+moduleConfigReferences: moduleConfigReferenceElement+;
+
+moduleConfigReferenceElement: unrestrictedIdentifier whiteSpace? EQ whiteSpace? STRINGLITERAL whiteSpace? SEMICOLON whiteSpace? STRINGLITERAL endOfStatement;
```
bclothier: I think I'd rather have literal `OBJECT` than `unrestrictedIdentifier` for the LHS part. This has potential to hide bugs or something worse if we allow for a loose interpretation.

MDoerner:
We will have to kind of parse it anyway. Simply removing the initial part completely is not an option as anything not in the token stream is lost on rewrite. Accordingly, we would destroy the form with every refactoring.

Moreover, to deal with it like with conditional compilation directives also requires to parse it in the precompilation parser in order to know which to know which tokens to hide.

In addition, we already had rules in the grammar for module configurations, which had not been sufisticated enough to deal with VBA forms. However, they might be necessary for some exported VBA code. I would rather not like to risk breaking existing functionality here. 

All in all, simply parsing the forms seems to me to be the easiest and most robust solution to the problem.

Rubberduck.Parsing/Grammar/VBAParser.g4
MDoerner: Oh, I hate that the grammar uses spaces. Will correct it.

WaynePhillipsEA:
With the absence of the specification, I respectfully disagree.  Removing the vb forms text and restoring it before any rewrite seems far simpler than trying to parse something that we don't actually want to have to parse, and don't have the spec for.

Rubberduck.Parsing/Grammar/VBAParser.g4
Vogel612: While we're on that: @retailcoder also introduced a tabs vs spaces mismatch in the PR to exclude certain identifiers from linelabels

MDoerner: I will go through the entire file and replace all the tabs with the usual number of spaces; there are a lot of mismatches.

bclothier:
@WaynePhillipsEA  

It may help to consider that parsing it does not mean actually consuming it. ATM, we don't really do much with that section. So even if there were some inaccuracy or mismatch with the unknown specs, parser would merely have wrong tokens assigned. Because we don't consume the result, it hardly matter and we don't need to manually work around the rewriting API for that section of text. 

For our use case, it's enough to handle the top-level and block statements so we know where the section starts/ends, then for the rest of process, we basically just ignore that block and not analyze those tokens within the RD codebase.

PR #3948, opened 2018-04-21
Remove the restrictions on the admin elevation

### LINKED ISSUE #3921: Rubberduck doesn't load v2.2.0.3086
After updating to the V2.2.0 release Rubberduck no longer loads at all.

I can see that the new version is installed in Add\Remove programs. 

No Add-Ins appear in my VBE Add-In Manager.

There is no log because it never gets that far. The installer appears to be silently failing to do something.
(Where would I find the installer log?)

bclothier:
Closes #3921 

I made some changes to the `RegisterAddIn()` function and in testing, the bug that was previously observed no longer exhibits and I am able to install as an admin and thus register the IDE for myself. However, I would appreciate others testing to verify it's not "just my computer" symptoms.

Pending: Update the messaging around admin not being able to register as they should no longer apply. Will remove after tests.

### DIFF
--- a/Rubberduck.Deployment/InnoSetup/Installer Build Script.iss
+++ b/Rubberduck.Deployment/InnoSetup/Installer Build Script.iss
@@ -514,18 +514,19 @@ end;
 
 ///<remarks>
 ///Called after successfully installing, including via the elevated installer
-///to register the VBE addin. Should be never run under elevated context
-///or the registration may not work as expected.
+///to register the VBE addin.
 ///</remarks>
 procedure RegisterAddin();
 begin
-  if not IsElevated() then
-  begin
-    RegisterAddinForIDE(HKCU32, 'Software\Microsoft\VBA\VBE\6.0\Addins', '{#AddinProgId}');
-
     if IsWin64() then 
+    begin
+      RegisterAddinForIDE(HKCU32, 'Software\Microsoft\VBA\VBE\6.0\Addins', '{#AddinProgId}');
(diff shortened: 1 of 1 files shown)


PeterMTaylor:
To provide some feedback on my PC. 
	Rubberduck version 2.2.6672.28001 loading:
	Operating System: Microsoft Windows NT 6.2.9200.0 x64
	Host Product: Microsoft Office 2010 x86
	Host Version: 14.0.7197.5000
	Host Executable: EXCEL.EXE; 
Able to install RD 2.2 green light release under as Admin. Window Defender Smartscreen complains so I allow this to run.  Doesn't show any RD until I execute the Repair VBE batch file and the RD 2.2 works.

bclothier:
That's expected for the green release 2.2. Just for reference, the installer with this PR changes can be downloaded here: https://ci.appveyor.com/api/buildjobs/drbpn8my5ng2k4d7/artifacts/Rubberduck.Deployment%2FInnoSetup%2FInstallers%2FRubberduck.Setup.2.2.0.3137-pre.exe

PeterMTaylor:
Thanks, @bclothier. 
Have uninstalled green release 2.2 firstly and installed the ci linked 2.2.0.3137-pre without Run as Admin. Installs fine after expected Windows Defender SmartScreen allowing to run. The "Only You" option is enabled (nice!) so I selected this. Then prompts for wished to register RD, I left the default tick and complete the install. Works the moment I start the cDataSet test I had.
Version 2.2.6685.5179
OS: Microsoft Windows NT 6.2.9200.0, x64
Host Product: Microsoft Office 2010 x86
Host Version: 14.0.7197.5000
Host Executable: EXCEL.EXE

PR #3958, opened 2018-04-24
Fix ExecuteCode/unit tests to work with VB6

WaynePhillipsEA:
Background: the IVBETypeInfo interface IID changed between VB6 and VBA, which means ExecuteCode/unit tests weren't working with the new VB6 support.

This PR completely removes the dependency on the IVBETypeInfo interface, moving the GetStdModAccessor member over to the IVBEComponent interface (which hasn't changed since at least the VB6 days), where it is implemented as well.


WaynePhillipsEA:
closes #3953

bclothier:
The change seems good since we already require a standard module (e.g. `IVBComponent`) anyway. 

Just to confirm - if we lose the `IVBETypeInfo` do we lose any extra metadata we wouldn't have had via the normal `ITypeInfo`?

WaynePhillipsEA:
No we don't lose anything.  IVBETypeInfo was just being used for the GetStdModAccessor member, which I've now located on IVBEComponent.  There shouldn't be any side effects.

PR #3976, opened 2018-04-30
Introducing Rubberduck.Resources

retailcoder:
This will likely conflict with #3975 a bit.

ref. #3862


bclothier:
I do think this is a case of choosing your own poisons. That said, I can offer one reason to centralize the resources --- it eliminate the guesswork of "do I need to share it?". A changing of resource is now only a namespace realignment; no project references needs to be changed, and though it's not strictly necessary, there is something to be said for having all i18n stuff in one place. 

The bad is that of course, it can turn into a kitchen sink and we so totally don't want it. Also when we do change from `Rubberduck.Core` to `Rubberduck.UI`, `Rubberduck.Resources` could seem like hair splitting because 90%+ of resources are for UI.

So yeah, pick one or other and be happy/sad.

retailcoder:
@bclothier okay then I'll pick the "let's put everything we can under `Rubberduck.Resources`"; the DLL itself might not need to be updated anytime soon, but if we add Italian and Spanish translations then we should be able to do that without needing to recompile the regex assistant.

Thoughts?

Rubberduck.Core/Settings/HotkeySetting.cs
```diff
@@ -29,14 +29,7 @@ public class HotkeySetting
...
 
-        public override string ToString()
-        {
-            return string.Format("{0}{1}{2}{3}",
-                HasCtrlModifier ? KeyModifierCtrl : string.Empty,
-                HasShiftModifier ? KeyModifierShift : string.Empty,
-                HasAltModifier ? KeyModifierAlt : string.Empty,
-                Key1);
-        }
+        public override string ToString() => $"{(HasCtrlModifier ? KeyModifierCtrl : string.Empty)}{(HasShiftModifier ? KeyModifierShift : string.Empty)}{(HasAltModifier ? KeyModifierAlt : string.Empty)}{Key1}";
```
bclothier: what's up with this? seems less readable?

Rubberduck.Main/Rubberduck.Main.csproj
```diff
@@ -141,7 +141,8 @@
     <Optimize>false</Optimize>
   </PropertyGroup>
   <PropertyGroup>
-    <ApplicationIcon>Ducky.ico</ApplicationIcon>
+    <ApplicationIcon>
+    </ApplicationIcon>
```
bclothier: Is that intentional?

Rubberduck.VBEditor.VBA/Rubberduck.VBEditor.VBA.csproj
```diff
@@ -112,7 +112,7 @@
       <EmbedInteropTypes>False</EmbedInteropTypes>
     </Reference>
     <Reference Include="stdole, Version=7.0.3300.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
```
bclothier: Do we want that? I thought we should be embedding to avoid problems w/ versioning?

retailcoder: Hmm I toggled that in response to a build warning.

Rubberduck.Main/Rubberduck.Main.csproj
retailcoder: Nope, WIP :smiley:

Rubberduck.Core/Settings/HotkeySetting.cs
retailcoder: Yeah, agreed. will revert.

retailcoder: Done.

retailcoder:
This PR is pretty much impossible to review, but these two screenshots summarize pretty much everything that's happening (`Meta.resx` was renamed to `InspectionInfo.resx`):

image

image

Some cleanup still needs to happen with `Resources.resx` (contains a number of icons and images) and `RubberduckUI.resx` needs to be sliced.

"About" resource keys in `RubberduckUI` have already been extracted into `AboutUI.resx`, but I'll stop it here before hitting 1000 affected files, and continue the slicing of `RubberduckUI.resx` in another PR - I'll make one PR per feature group, to limit the scope.

Thoughs on whether it's a good idea to merge this now and continue later, or if more changes should be included here?

bclothier:
My vote would be to merge it and try to keep subsequent reorganization PR limited in the scope. It looks better on the history and helps avoid hiding any potential mistakes.

PR #3985, opened 2018-05-03
Adjust the IoC installer

MDoerner:
This PR includes a few tweaks around the IoC installer.

1. I completely removed the string-based loading of assemblies; only assemblies directly or indirectly referenced by `Rubberduck.Main` are now considered for the registration conventions.
2. After digging a bit through CW, I found out that probably using `If` on `BasedOn` registrations is the best way to handle the check against experimental types.
3.  I extended the experimental type check to the commands and commandMenuItems and also restricted the menu items loaded based on the experimental attribute.
4. I tweaked a condition to probably achieve better performance using the convention.
5. I removed all static modifiers to reduce the possibility that anything sticks to a static context in th IoC installer.
6. I made extract method and regex search and replace experimental in non-debug builds instead of only not showing the corresponding menu items.


Rubberduck.Main/Root/RubberduckIoCInstaller.cs
```diff
@@ -748,53 +754,29 @@ private void RegisterCustomDeclarationLoadersToParser(IWindsorContainer containe
...
-                catch (Exception)
-                {
-                    // can we log yet?
-                }
-            }
-            */
-
-            return assemblies;
+            var referencedAssemblies = assembly.GetReferencedAssemblies()
+                                                .Where(asmbly => filterPredicate(asmbly))
```
Vogel612: this restricts the paths which we take through the dependency graph. That way we only consider Reference paths that contain only `Rubberduck.*`  assemblies.

This implies that transitive Rubberduck References that have an assembly with a different name in between do not get considered. 
That seems fair, considering we want to lock down what we consider for injection, might be useful to make note of that in a comment or XML-Doc

MDoerner:
I decided to introduce a new `DisabledAttribute` in order to allow to exclude types from the conventions in the IoC installer without making them experimental, which would  allow to reactivate them in the settings once everything works as intended.

PR #3991, opened 2018-05-05
Fixed null parameter names when collecting reference VB6.OLB

### LINKED ISSUE #3951: Parsing a VBForm results in Unexpected Error
See https://github.com/rubberduck-vba/Rubberduck/pull/3935#discussion_r181759991

Currently, parsing an empty VBForm fails with Unexpected Error. This is caused by two null reference exceptions:

https://github.com/rubberduck-vba/Rubberduck/blob/14743dc32ef02938b569b841fe94e8d250217b72/Rubberduck.Parsing/Symbols/Declaration.cs#L578

image

https://github.com/rubberduck-vba/Rubberduck/blob/14743dc32ef02938b569b841fe94e8d250217b72/Rubberduck.Parsing/Symbols/DeclarationFinder.cs#L130

image

The root cause appears to be collection of the VB6.OLB reference.

mansellan:
Closes #3951 

Root cause was that this line:
https://github.com/rubberduck-vba/Rubberduck/blob/f43b2b172c120d143750e9a7f8ca5feb06515b4d/Rubberduck.Parsing/ComReflection/ComMember.cs#L88

was returning null values for all array values, but a non-zero count. From this I infer that not all parameter names in the type library are named.

This PR also adds an Assert in the parameter name, and names the background threads which are collecting the references. Happy to remove these if not wanted, but I think they could be helpful for debugging in future.

### DIFF
--- a/Rubberduck.Parsing/ComReflection/ComMember.cs
+++ b/Rubberduck.Parsing/ComReflection/ComMember.cs
@@ -91,7 +91,7 @@ namespace Rubberduck.Parsing.ComReflection
             {
                 var paramPtr = new IntPtr(funcDesc.lprgelemdescParam.ToInt64() + Marshal.SizeOf(typeof(ELEMDESC)) * index);
                 var elemDesc = (ELEMDESC)Marshal.PtrToStructure(paramPtr, typeof(ELEMDESC));
-                var param = new ComParameter(elemDesc, info, names[index + 1]);
+                var param = new ComParameter(elemDesc, info, names[index + 1] ?? string.Empty);
                 Parameters.Add(param);
             }
             if (Parameters.Any() && funcDesc.cParamsOpt == -1)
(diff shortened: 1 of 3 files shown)


bclothier:
According to this documentation [ITypeInfo::GetNames](https://msdn.microsoft.com/en-us/library/windows/desktop/ms221711(v=vs.85).aspx), it is indeed legal and should be expected that parameters may be unnamed for some cases, in particular:

> For property put and put reference functions, the right side of the assignment is unnamed. 

Whether the VB6.OLB actually falls in this case, I don't know. 

Incidentally that kinds of explains why you see VBE putting in `RHS` when implementing an interface with a property let/set; it simply doesn't exist. So in this case, we can take lead from VBE and use `RHS` as the default name (though I do wonder what'd happen w/ multiple parameters...)

If we want to resolve this for the general case, we might need to provide some default name like `parameterX` where `X` corresponds to its ordinal position.

MDoerner:
I would prefer `XunamedParameter` as the artificial name because that cannot coliseum with anything: it is not a legal name.

bclothier:
If we do go w/ the illegal name approach, it might be necessary to check that no downstream code will try to use it in the COM side of things.

PR #3992, opened 2018-05-06
Added missing ComponentType icons or VB6

### LINKED ISSUE #3950: Code Explorer needs additional icons for VB6 DeclarationTypes
Currently, parsing a VBForm causes an exception here: https://github.com/rubberduck-vba/Rubberduck/blob/14743dc32ef02938b569b841fe94e8d250217b72/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs#L44

as the declaration type for VBForms is not in the dictionary. We should identify the missing declaration types and create icons for them. It may be worthwhile to have a default too (or at least guarding with ContainsKey), to avoid blowing up CE if anything unexpected is found.

mansellan:
Closes #3950 

This PR adds the 'missing' icons for the VB6-specific [ComponentType](https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.VBEEditor/SafeComWrappers/VB/Enums/ComponentType.cs)s (enum values 4 through 10 inclusive). Icons sourced mostly from Fugue, with the exception of MDIForm which was adapted from the existing ProjectForm custom icon (originally from SharpDevelop).

image

The only concern I have is that I had to change the type base of the DeclarationType enum from int to long. This *should* be safe as long as we're not casting it to int anywhere.


bclothier:
Nice! Just a question - what's up with that blank name for the resource file?

mansellan:
Seems to be just how VB6 rolls.... Only one resource file allowed per project, and it doesn't have a Name property. Note though that the PE has file names in brackets, which CE doesn't (and hasn't needed to ).

bclothier:
I see what you mean, yeah. That said, it does seem weird to see a blank field and I twitch thinking there was some kind of an error. It might be necessary to provide a fake name so people don't wonder why there's a blank entry?

mansellan:
Yep good plan, lemme take a look.

bclothier:
and while we're at it, we might as well ensure we have appropriate handling for unknown/unexpected types:

https://github.com/rubberduck-vba/Rubberduck/pull/3992/files#diff-e68df6e7c2f0732d3f095af3d0ab4d63R44

could be made safer, with no icon as the fallback.

mansellan:
image

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs
```diff
@@ -50,7 +50,9 @@ public CodeExplorerComponentViewModel(CodeExplorerItemViewModel parent, Declarat
                                 .Select(item => new CodeExplorerMemberViewModel(this, item, grouping)))
                                 .ToList<CodeExplorerItemViewModel>();
 
-            _name = Declaration.IdentifierName;
+            _name = DeclarationType == DeclarationType.ResFile && string.IsNullOrEmpty(Declaration.IdentifierName) 
+                ? "Resources"
```
rkapka: Can this be hard-coded in English?

mansellan: eek sorry!

rkapka: I forgive you ;-)

Rubberduck.Core/Rubberduck.Core.csproj
```diff
@@ -996,6 +996,12 @@
     <Resource Include="Resources\plus-circle.png" />
   </ItemGroup>
   <ItemGroup>
+    <Content Include="Resources\document-block.png" />
+    <Content Include="Resources\document-globe.png" />
+    <Content Include="Resources\pencil-ruler.png" />
+    <Content Include="Resources\status-offline.png" />
+    <Content Include="Resources\ui-scroll-pane-form.png" />
+    <Content Include="Resources\ui-tab-content.png" />
```
retailcoder: Shouldn't these be `<Resource Include="Resources\..." />` tags?

PR #4006, opened 2018-05-13
Splice Off Rubberduck.Refactorings from Rubberduck.Core

Vogel612:
This PR also adresses `IMessageBox` leaking types from the System.Windows.Forms namespace by pushing the different configurations for MessageBoxButtons and MessageBoxIcons into appropriately named methods.

Furthermore this introduces the `Rubberduck.Interaction` project, which provides interfaces for interacting with the user that are not inside the `Rubberduck.Core` project. This allows Projects that are to be consumed by Rubberduck.Core to interact with the user without creating a Cyclic dependency.

On that note this also fixes the Model of RemoveParameters and ExtractInterface being intermingled with ViewModel concerns (`IsRemoved` and `IsSelected` respectively). 
This broke the RemoveParametersRefactoring (16 failing tests).

While the "104 changed files" sounds like a lot, the vast majority of them are just renamed and have changes in the low single digits to accomodate the moved classes.


Rubberduck.API/Rubberduck.API.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Rubberduck.API</RootNamespace>
     <AssemblyName>Rubberduck.API</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
bclothier: Just as FYI - #3975 will bump the version to 4.6, so this will conflict.

Rubberduck.Core/App.cs
```diff
@@ -155,7 +154,7 @@ private void LoadConfig()
             catch (CultureNotFoundException exception)
             {
                 Logger.Error(exception, "Error Setting Culture for Rubberduck");
-                _messageBox.Show(exception.Message, "Rubberduck", MessageBoxButtons.OK, MessageBoxIcon.Error);
+                _messageBox.NotifyError(exception.Message, "Rubberduck");
```
bclothier: I am a little ambivalent about the `Caption` property - in other examples, we seem to be passing hard-coded literal `"Rubberduck"` - seems to me that it should be made the default, or substituted with a resource key so that we maintain the default caption in only one place. OTOH, having same captions that says nothing more than just the name of program seems to be a waste of the titlebar? I'd have preferred an explicit summary of the choice (titles like `Confirm required`, `Data missing`, that kind of stuff).

Rubberduck.Core/UI/CodeExplorer/Commands/RemoveCommand.cs
```diff
@@ -55,15 +56,14 @@ protected override bool EvaluateCanExecute(object parameter)
         protected override void OnExecute(object parameter)
         {
             var message = string.Format("Do you want to export '{0}' before removing?", ((CodeExplorerComponentViewModel)parameter).Name);
```
bclothier: Shouldn't this be a resource key?

Rubberduck.Core/UI/CodeExplorer/Commands/RemoveCommand.cs
```diff
@@ -55,15 +56,14 @@ protected override bool EvaluateCanExecute(object parameter)
...
-            var result = _messageBox.Show(message, "Rubberduck Export Prompt", MessageBoxButtons.YesNoCancel,
-                    MessageBoxIcon.Exclamation, MessageBoxDefaultButton.Button1);
+            var result = _messageBox.Confirm(message, "Rubberduck Export Prompt", true);
```
bclothier: FWIW, I'd have used `Confirm export` as the caption but that's just me.

Rubberduck.Core/UI/Refactorings/RemoveParameters/RemoveParametersPresenter.cs
```diff
@@ -1,16 +1,11 @@
-﻿using System.Windows.Forms;
+﻿using Rubberduck.Interaction;
 using Rubberduck.Resources;
-using Rubberduck.UI;
-using Rubberduck.UI.Refactorings;
-using Rubberduck.UI.Refactorings.RemoveParameters;
+using Rubberduck.Refactorings.RemoveParameters;
+using System.Windows.Forms;
```
bclothier: Should this reference be here still?

Rubberduck.Core/UI/Refactorings/ReorderParameters/ReorderParametersPresenter.cs
```diff
@@ -1,18 +1,14 @@
 ﻿using System.Collections.ObjectModel;
 using System.Linq;
-using System.Windows.Forms;
 using Rubberduck.Resources;
-using Rubberduck.UI;
+using Rubberduck.Interaction;
 using Rubberduck.UI.Refactorings;
 using Rubberduck.UI.Refactorings.ReorderParameters;
+using System.Windows.Forms;
```
bclothier: Same question RE: the reference

Rubberduck.Inspections/Rubberduck.Inspections.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Rubberduck.Inspections</RootNamespace>
     <AssemblyName>Rubberduck.Inspections</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
bclothier: Signpost RE: framework version

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Rubberduck.Interaction
+{
+    public interface IMessageBox
+    {
+        void Notify(string text);
+        void NotifyError(string text, string caption);
+        void NotifyWarn(string text, string caption);
+        bool Question(string text, string caption);
+        bool Prompt(string text, string caption);
```
bclothier: It's not clear what the intended usage of `Question` and `Prompt` are - they sounds the same to me. 

It's also not clear how `Notify` differs from `NotifyError`/`NotifyWarn`.

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,75 @@
...
+        bool ConfirmYN(string text, string caption);
+        bool ConfirmYN(string text, string caption, bool suggestion);
+        bool? Confirm(string text, string caption, bool? suggestion);
```
bclothier: `Confirm` isn't a Yes/No confirmation? It's best to be explicit - call it `ConfirmYesNoCancel` or something like that.

Rubberduck.Main/Rubberduck.Main.csproj
```diff
@@ -8,12 +8,13 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Rubberduck</RootNamespace>
     <AssemblyName>Rubberduck</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
bclothier: Signpost RE: framework version

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public interface IMessageBox
+    {
+        void Notify(string text);
+        void NotifyError(string text, string caption);
+        void NotifyWarn(string text, string caption);
+        bool Question(string text, string caption);
+        bool Prompt(string text, string caption);
+        bool ConfirmYN(string text, string caption);
+        bool ConfirmYN(string text, string caption, bool suggestion);
```
bclothier: I think I'd rather that it was spelled out `ConfirmYesNo` - this has no change on the intellisense selection and make it more readable.

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,75 @@
...
+            Forms.MessageBox.Show(text, caption, Forms.MessageBoxButtons.OK, Forms.MessageBoxIcon.Exclamation);
+        }
+
+        public bool Question(string text, string caption)
+        {
+            return Forms.MessageBox.Show(text, caption, Forms.MessageBoxButtons.YesNo, Forms.MessageBoxIcon.Question) == Forms.DialogResult.Yes;
+        }
+
+        // FIXME check whether to merge with Question
+        public bool Prompt(string text, string caption)
```
bclothier: Yes definitely merge those into one. They do the same thing (NB: `Question` and `Prompt` likely came from two different uses of the old Windows enumerations, which is why we have 2 enumerations for same icons (c.f. [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms645505(v=vs.85).aspx) )

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            return Forms.MessageBox.Show(text, caption, Forms.MessageBoxButtons.YesNo, Forms.MessageBoxIcon.Exclamation) == Forms.DialogResult.Yes;
+        }
+
+        public bool ConfirmYN(string text, string caption, bool suggestion)
+        {
+            return Forms.MessageBox.Show(text, caption, Forms.MessageBoxButtons.YesNo, Forms.MessageBoxIcon.Exclamation, suggestion ? Forms.MessageBoxDefaultButton.Button1 : Forms.MessageBoxDefaultButton.Button2) == Forms.DialogResult.Yes;
+        }
+
+        public bool? Confirm(string text, string caption, bool? suggestion)
```
bclothier: FWIW, instead of a `bool?` it seems better to wrap the `DialogResult` so that we can have more readable code like `var response = Confirm(...);` => `if(response = DialogResult.Cancel) ....`

Rubberduck.Core/Rubberduck.Core.csproj
```diff
@@ -9,12 +9,13 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Rubberduck</RootNamespace>
     <AssemblyName>Rubberduck.Core</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
bclothier: Signpost RE: framework version

Rubberduck.Interaction/IMessageBox.cs
retailcoder: @bclothier meh. `if _msgbox.Confirm(...) {...}` is just as expressive IMO; would you know whether to check for `DialogResult.Yes` or `DialogResult.OK` if you weren't providing arguments telling you what the buttons are going to be? A Boolean makes it much clearer what the possible returned values are, provided that the caller is testing for the positive result, i.e. assuming that "not no" means "yes" would be a mistake... might be worth noting in the XML-doc summary 😉

bclothier: This works for the `bool` where there's only 2 results but with `bool?`, there's now 3 possible results. What does `null` means then? As it is, I'd have to look at the source code to see that `null` meant the user cancelled out and not some silly third button like `Abort`. In this case, I'd just have a custom `enum` allowing only `Yes`, `No`, `Cancel` -- no need to deal with `OK` or `Abort` or `null`.

Rubberduck.Core/App.cs
Vogel612: In this case I'd not want to make this into a resource string, since that's where stuff is blowing up when we couldn't set the culture for resources :/ As such a resource key doesn't make terribly much sense here. I will add a comment to that effect, though.

Rubberduck.Core/UI/Refactorings/RemoveParameters/RemoveParametersPresenter.cs
Vogel612: the View itself is implemented as an `IRefactoringDialog` which piggy-backs off of Windows Forms' `DialogResult`

Rubberduck.Core/UI/Refactorings/ReorderParameters/ReorderParametersPresenter.cs
Vogel612: Same reason as in RemoveParametersPresenter

Rubberduck.Interaction/IMessageBox.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using Forms = System.Windows.Forms;
+
+namespace Rubberduck.Interaction
+{
+    public interface IMessageBox
+    {
+        void Message(string text);
```
mansellan: Should this have a caption parameter?

Vogel612: I don't see much of a benefit, honestly. And I am somewhat hard-pressed to come up with a good default, soo ...

mansellan: I mention it only be because all the other methods have a caption, this was the odd one out.

Rubberduck.Interaction/IMessageBox.cs
Vogel612: As mentioned in chat: this was just the quick-and-dirty fix to get the Forms API out of the IMessageBox interface. It's not really at the core of this PR. If you want a proper fix for it, I'd suggest opening a follow-up issue and assigning that to me.

Vogel612:
I've extracted the remaining concerns into follow-up issues and assigned them to me. Considering that there have been two tentative approvals and the respective suggestions have either been incorporated or moved into a follow up, I'm merging this PR

PR #4007, opened 2018-05-14
VB6 ProjectId saved to .vbp file

### LINKED ISSUE #3990: VB6 should store project ID in the project file
In VBA, we store the RD Project ID in the VBProject.HelpFile property. 

This is because in VBA:
1) Custom VBProject properties are not available
2) Help files are extremely uncommon (due to complexity of authoring the required format)

For VB6:
1) is not true - the VBProject object has ReadProperty and WriteProperty which allow custom properties to be written into the project file
2) will be less true - for large VB6 projects, Help files are often essential despite the complexity barrier

mansellan:
Closes #3990

### DIFF
--- a/Rubberduck.Core/UI/Command/Refactorings/ProjectExplorerRefactorRenameCommand.cs
+++ b/Rubberduck.Core/UI/Command/Refactorings/ProjectExplorerRefactorRenameCommand.cs
@@ -53,7 +53,7 @@ namespace Rubberduck.UI.Command.Refactorings
             
             return _state.AllUserDeclarations.SingleOrDefault(
                     t => t.IdentifierName == Vbe.SelectedVBComponent.Name &&
-                            t.ProjectId == Vbe.ActiveVBProject.HelpFile &&
+                            t.ProjectId == Vbe.ActiveVBProject.ProjectId &&
                             new[]
                                 {
                                     DeclarationType.ClassModule,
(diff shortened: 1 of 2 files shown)


Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -19,7 +20,21 @@ public VBProject(VB.VBProject target, bool rewrapping = false)
...
+        public string ProjectId
+        {
+            get
+            {
+                try
+                {
+                    return IsWrappingNullReference ? string.Empty : Target.ReadProperty("Rubberduck", "ProjectId");
+                }
+                catch (COMException)
+                {
```
mansellan: Note, there doesn't appear to be a clean way to check for property existence, hence catching ComEx.

Rubberduck.Core/UI/Command/Refactorings/ProjectExplorerRefactorRenameCommand.cs
```diff
@@ -53,7 +53,7 @@ private Declaration GetTarget()
             
             return _state.AllUserDeclarations.SingleOrDefault(
                     t => t.IdentifierName == Vbe.SelectedVBComponent.Name &&
-                            t.ProjectId == Vbe.ActiveVBProject.HelpFile &&
+                            t.ProjectId == Vbe.ActiveVBProject.ProjectId &&
```
mansellan: Found this along the way - should be going via ProjectId indirection.

PR #4022, opened 2018-05-20
Stop removing new line when removing an expression with a line number

### LINKED ISSUE #3974: Move closer refactoring leaves line number behind
If a declaration statement has a line number preceding it, the line number will be left behind when moving the declaration. This probably makes sense because the number may not be in sequence if it was moved too (although that seems to be valid VBA). However, since the following line is moved up, that line inherits the line number and if it already has one, we get a line with two numbers as per the following example:

Before refactoring:
```vb
Sub SimpleLineNumber()
1   Dim a As Long
2   Dim b As Long
3   a = 1
4   b = 1
End Sub
```

After refactor on line 1:
```vb
Sub SimpleLineNumber()
1   2   Dim b As Long
Dim a As Long
3   a = 1
4   b = 1
End Sub
```

tommy9:
Fixes #3974 

Add special case where a variable declaration has a line number. Normally the declaration and following new line (end of statement) are removed which helps avoid a blank line being left behind. However, with a line number on the declaration the brings the code from the following line to the line with a line number. This can cause invalid code e.g. when the following line has a line number itself.

Incidentally, I can see unreachable code in the file VariableRewriterInfoFinder. The condition `if (blockStmtIndex == containingBlock.ChildCount)` doesn't make sense because it is comparing the (zero-based) index of an item to the number of items and it must always be less?

### DIFF
--- a/Rubberduck.Parsing/Rewriter/RewriterInfo/VariableRewriterInfoFinder.cs
+++ b/Rubberduck.Parsing/Rewriter/RewriterInfo/VariableRewriterInfoFinder.cs
@@ -65,7 +65,7 @@ namespace Rubberduck.Parsing.Rewriter.RewriterInfo
                 int stopIndex = variables.Stop.TokenIndex + 1; // also remove trailing newlines?
                 
                 var containingBlock = (VBAParser.BlockContext)mainBlockStmt.Parent.Parent;
-                var blockStmtIndex = containingBlock.children.IndexOf(mainBlockStmt.Parent);
+                var blockStmtIndex = containingBlock.children.IndexOf(blockStmt);
                 // a few things can happen here
                 if (blockStmtIndex == containingBlock.ChildCount)
                 {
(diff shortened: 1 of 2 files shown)


Rubberduck.Parsing/Rewriter/RewriterInfo/VariableRewriterInfoFinder.cs
```diff
@@ -89,7 +90,16 @@ private static RewriterInfo GetRewriterInfo(VBAParser.VariableSubStmtContext var
...
-                            stopIndex = containingBlock.GetChild<ParserRuleContext>(blockStmtIndex + 2).Start.TokenIndex - 1;
+                            if (blockStmt.statementLabelDefinition() != null)
+                            {
+                                // special case where line has statement label or line number
+                                // don't want to remove trailing newline or can have label/number apply to next line where it isn't necessarily valid
+                                stopIndex = eol.Stop.TokenIndex - 1;
+                            }
+                            else
+                            {
+                                stopIndex = containingBlock.GetChild<ParserRuleContext>(blockStmtIndex + 2).Start.TokenIndex - 1;
```
Vogel612: 2 is a magic number here that should be explained.

Rubberduck.Parsing/Rewriter/RewriterInfo/VariableRewriterInfoFinder.cs
```diff
@@ -89,7 +90,16 @@ private static RewriterInfo GetRewriterInfo(VBAParser.VariableSubStmtContext var
                         }
                         else
                         {
-                            stopIndex = containingBlock.GetChild<ParserRuleContext>(blockStmtIndex + 2).Start.TokenIndex - 1;
+                            if (blockStmt.statementLabelDefinition() != null)
```
Vogel612: please collapse this into an `else if`, that saves one level of indentation :)

Vogel612:
:+1: aside from two things that should be easily fixed. Thanks for your contribution :)

tommy9:
Thanks for the review. Definitely looks cleaner with your suggestions.

PR #4068, opened 2018-06-09
Fix evil for loop

### LINKED ISSUE #4066: Parser Error reported on certain For Next loop structures
Write a nested `For Next` in the form

    For x = a To b
    For y = c To d
        ' ...
    Next y, x

Or

    For x = a To b
    ' some code
    For y = c To d
        ' ...
    Next y, x

Both of which are valid syntax

throws a Parser Error on the `Next y, x`

> extraneous input ',' expecting {<EOF>, ':', REM, NEWLINE, ''', WS, LINE_CONTINUATION}

[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2082276/RubberduckLog.txt)

Version 2.2.6672.28001
Host Version: 16.0.9330.2087
Host Executable: EXCEL.EXE

MDoerner:
This PR closes #4066 (I see the indenter issue as a separate issue.)

In addition to amending the grammar to cope with combined `Next` statements for `For` and `For Each` loops based on the idea by @Vogel612 (see the referenced issue), this PR removes the `DeploymentItemAttribute`. Instead of using this to make the test engine copy test files into the test environment the test runner runs in, the tests now reference the original test files. As a consequence, one does no longer run into authorization problems when not running VS as admin and the ReSharper test runner can be used again.


RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -1484,6 +1484,126 @@ Sub Test()
...
+            string code = @"
+Sub Test()
+    For n = 1 To 10
+        For m = 1 To 20
+            a = m + n
+        Next m _
+    , n%
+End Sub";
+            var parseResult = Parse(code);
+            AssertTree(parseResult.Item1, parseResult.Item2, "//forNextStmt", matches => matches.Count == 2);
```
bclothier: I realize the PR is already merged but I want to comment that all tests seems to verify that there are correct counts of the nested for blocks. Shouldn't we have a test to verify that the `m` is the inner and `n` is the outer, too?

RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -1484,6 +1484,126 @@ Sub Test()
...
+        b = n
+        For Each c In foo
+            For m = 1 To 20
+                For Each d In bar
+                    a = m + n + c + d
+                        For k = 0 To 100
+                            t = a + k
+    Next k, d, m,_
+            c, _
+            n
```
bclothier: Same comment but for the mixed cases (verifying that the `c` and `d` are really assigned to the `ForNext` while the `n, m, k`, `For`.)

PR #4090, opened 2018-06-14
VB6 - get project and component events working

### LINKED ISSUE #4046: VB6 project events not working
Several project-level events (Project load \ unload etc) not functional in VB6.

mansellan:
Closes #4046 

Calling EbMode doesn't seem to work, so for now I've returned to querying the main window caption. Inelegant but seems reliable.

### DIFF
--- a/Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
+++ b/Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
@@ -62,7 +62,7 @@ namespace Rubberduck.VBEditor.SafeComWrappers.VB6
                 using (var vbe = VBE)
                 using (var mainWindow = vbe.MainWindow)
                 {
-                    var caption = mainWindow.Caption;
+                    var caption = mainWindow.Caption.ToLowerInvariant();
 
                     if (caption.EndsWith("[design]"))
                     {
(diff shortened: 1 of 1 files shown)


Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -58,14 +59,30 @@ public EnvironmentMode Mode
...
             {
-                if (IsWrappingNullReference)
+                using (var vbe = VBE)
+                using (var mainWindow = vbe.MainWindow)
                 {
-                    return 0;
-                }
+                    var caption = mainWindow.Caption;
+
+                    if (caption.EndsWith("[design]"))
```
bclothier: IDK if that is different for VB6 or if I'm not remembering but I think it's usually proper cased (`[Design]`)? The comparison is case sensitive. It might be better to make the comparison case-insensitive so even if they managed to get it to go `[DeSiGn]` it'll still work? 

(please don't ask me how they'd manage to achieve that!)

mansellan: Good point, will do. Hoping it's not locale-sensitive...

mansellan:
Won't work, international editions have the caption text localised. Closing PR.

PR #4111, opened 2018-06-18
VB6 - Add missing VBForm grammar cases

### LINKED ISSUE #3995: VB6 VBForm parse errors
In testing RD against real-world VB6 forms, I'm finding additional parse failures with form headers.

As I find them, I'll update this issue so they can be tackled collectively.

mansellan:
Closes #3995.

Note that a couple of changes have been made to the lexer:

1. BARE_HEX_LITERAL added after IDENTIFIER
2. Braces have been added to the excluded character set of IDENTIFIER

Point 1 should be safe, as there should be no other rules that allow a numeric followed by a-f (the only case that won't be captured by IDENTIFIER)

I *believe* point 2 is safe, as braces do not appear to be legal anywhere within identifiers in the VBE.


Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -62,7 +62,7 @@ moduleConfigProperty :
 ;
 
 moduleConfigElement :
-    unrestrictedIdentifier whiteSpace? EQ whiteSpace? expression (COLON numberLiteral)? endOfStatement
+    unrestrictedIdentifier(DOT unrestrictedIdentifier)* whiteSpace? EQ whiteSpace? (DOLLAR | PERCENT | PLUS POW | PLUS | POW)? L_BRACE? expression R_BRACE? (COLON unrestrictedIdentifier)? endOfStatement
```
MDoerner: I am rather sure this not what you want. First of all please note that `unrestrictedIdentifier` is always an identifier, which excludes all literals. Since the examples to which I adapted the grammar did have a `numberLiteral` at that position, I do not see how that option can be removed. In addition, I believe you always want to have matching braces,  right? In that case, use an alternative to either have no or both braces.

Moreover, there is a space missing in front of the first opening parenthesis.

MDoerner: In addition, I think the shortcut part should go into its own rule used here. Moreover, it seems more appropriate to use an `lExpression` instead of hand rolling the member access expression at the start.

mansellan: Yep I figured `unrestrictedIdentifier` was probably not right, but I need something a little more permissive than `numberLiteral` as it needs to accept not only digits but hex digits ([0-9A-H]). I wasn't sure about exposing a HEX_DIGIT from the lexer as I didn't want it matching anything except in this very narrow case.

Not sure how to proceed.

MDoerner: Please note that the hex digit also is not covered by an `unrestrictedIdentifier`.

mansellan:
Still gotta do the tests, but it seems to parse most of the cases when I kicked the tyres on it. Adding BARE_HEX_LITERAL to the lexer didn't work, as it would fail when the hex started A-F (picked it up as an IDENTIFIER). So I just allowed (numberLiteral | IDENTIFIER). A little too permissive, but that shouldn't matter.

The key was excluding braces from IDENTIFIER. I checked in the VBE and they are not legal, so I think this is a safe (and indeed, desirable) change.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -62,7 +62,12 @@ moduleConfigProperty :
 ;
 
 moduleConfigElement :
-    unrestrictedIdentifier whiteSpace? EQ whiteSpace? expression (COLON numberLiteral)? endOfStatement
+    lExpression whiteSpace? EQ whiteSpace? (shortcut | DOLLAR? expression) (COLON (numberLiteral | IDENTIFIER))? endOfStatement
```
MDoerner: Did you test this on a case where the hex value starts with a number and contains a letter?

mansellan:
@MDoerner Excellent points on the unit tests, it's something I need to get into the habit of doing. Being reflective for a second, I think that my negative past experience with TDD has taught me some very bad habits, which I should correct. Specifically, I race to the code out of sheer impatience. I need to remember the old maxim that sometimes: "Slower is quicker."

Thanks for reminding me of this. Next sync will have proper test coverage.

Rubberduck.Parsing/Grammar/VBALexer.g4
```diff
@@ -302,6 +302,7 @@ WS : [ \t];
 GUIDLITERAL : '{' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '}';
 IDENTIFIER :  ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
 LINE_CONTINUATION : [ \t]* UNDERSCORE [ \t]* '\r'? '\n';
+BARE_HEX_LITERAL : [0-9a-fA-F]+;
```
Vogel612: I **really** don't like this lexer rule. A lot of it's matching space is already covered by `IDENTIFIER` (which is terrible). It  can't ever match hex literals starting with a letter, because those are already matched by the `IDENTIFIER` token.

Are we 100% sure we need this rule? And if yes, can we document what we need it for?

Rubberduck.Parsing/Grammar/VBALexer.g4
```diff
@@ -298,8 +300,9 @@ SINGLEQUOTE : '\'';
 UNDERSCORE : '_';
 WS : [ \t];
 GUIDLITERAL : '{' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '}';
-IDENTIFIER :  ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
+IDENTIFIER :  ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
 LINE_CONTINUATION : [ \t]* UNDERSCORE [ \t]* '\r'? '\n';
+BARE_HEX_LITERAL : [0-9a-fA-F]+;
```
Vogel612: I really don't like this lexer rule. A lot of it's matching space is already covered by IDENTIFIER (which is terrible). It can't ever match hex literals starting with a letter, because those are already matched by the IDENTIFIER token.

Are we 100% sure we need this rule? And if yes, can we document what we need it for?

MDoerner: This rule is necessary for the VB6 Form declarations. These use hex numbers without a format specifier after the `OleObjectBlob`, which is totally stupid in a language that has hex literals, btw. There is no other rule matching a number followed by a letter.

Unfortunately, legal hex numbers both have an overlap with decimal numbers and identifiers. None of the two other possibilities must be matchedby this rule. Otherwise this special case handling for VB forms would cause problems in ordinary VBA. (Remember that what has been matched in the lexer cannot be undone in the parser.)

mansellan: Could make it perhaps more targeted with `[0-9][0-9a-fA-F]*;`, that would at least eliminate the overlap and (perhaps) make the intent more explicit ? The name might need tweaking if so though.

ThunderFrame: Am I missing something? Hex literals can start with A-F, not just 0-9, given a sufficient offset size into the FRX.

ThunderFrame: Couldn't the leading colon be used as a way of parsing the offsetIdentifier?

    [\:]whitespace?[0-9a-fA-F]

MDoerner: Since the lexer generally prefers longer matches, this would tokenize `bar : Add` as an identifier followed by an offset specification. 

Moreover, `whiteSpace` is not a thing in the the lexer. It is a parser rule.

BTW, we have the fragment `DIGIT` for `[0-9].

To explain why the suggestion to put the digit first: This removes the overlap with `IDENTIFIER`, which is currently preferred over `BARE_HEX_LITERAL`, as it should be, because it matches at least the same length and comes first in the lexer rules.
Offsets that do start with a letter are handled in the parser by also allowing `unrestrictedIdentifier`. (This could be simplified to `IDENTIFIER`, which is actually cleaner, but never came to 
my mind.) Hex numbers looking like decimal numbers are also not matched by `BARE_HEX_LITERAL` since they are already tokenized as `DECIMALLITERAL`. So we also allow `numberLiteral` for the offset. 

My suggestion would be to leave the rule as it is and add a comment what it is for and that numbers starting with a letter or only consisting of digits will be matched by other rules.

mansellan: @ThunderFrame just curious - have you ever seen an example with whitespace between the colon and the number? I haven't, and wonder if that would be legal.

If not, then the suggested lexer rule would become:

    RESOURCE_OFFSET : ':' [0-9a-zA-Z]+;

which makes the grammar:

    moduleConfigElement :
    (unrestrictedIdentifier | lExpression) whiteSpace? EQ whiteSpace? (shortcut | DOLLAR? expression RESOURCE_OFFSET?)? endOfStatement

This looks much cleaner to my eye. The only caveat is that `RESOURCE_OFFSET` would have to be defined before `COLON`, which I think should be fine.

Just testing this now - it has revealed a pre-existing issue with the Mid statement on this line:

https://github.com/rubberduck-vba/Rubberduck/blob/e2468806b7da88f20ca28a4ea0f08edb92cacd78/Rubberduck.Parsing/Grammar/VBAParser.g4#L467

IIUC, `ASSIGN` here is not correct, it should be `EQ`.

MDoerner: This lexer rule is equally disrupting. The whitespace after a statement separator is completely optional. Try what happens to `bar:Add` with this rule present.

MDoerner: Regarding the `Mid`, that should be an `EQ` indeed. Please fix that in another commit.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -62,7 +62,16 @@ moduleConfigProperty :
 ;
 
 moduleConfigElement :
-    unrestrictedIdentifier whiteSpace? EQ whiteSpace? expression (COLON numberLiteral)? endOfStatement
+    (unrestrictedIdentifier | lExpression) whiteSpace? EQ whiteSpace? (shortcut | resource | expression) endOfStatement
```
mansellan: I've just noticed that the final `| expression` is kinda redundant here, as it's captured by the resource rule. But it does no functional harm and helps to describe intent, so should it remain?

Rubberduck.Parsing/Grammar/VBALexer.g4
```diff
@@ -298,8 +300,10 @@ SINGLEQUOTE : '\'';
 UNDERSCORE : '_';
 WS : [ \t];
 GUIDLITERAL : '{' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '-' [0-9A-F]+ '}';
-IDENTIFIER :  ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\]()\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
+IDENTIFIER :  ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; 0-9-/\\-] ~[[\](){}\r\n\t.,'"|!@#$%^&*\-+:=; -]*;
```
bclothier: I don't think we should be modifying the `IDENTIFIER` here. This is not a legal identifier in the VBA code. Could we use a separate lexer rule for VB6 forms to allow the use of curly braces?

mansellan: The lists specify characters that are *not* legal in VBA identifiers - `~[...]`, which includes braces.

Issue #3962, opened 2018-04-24
Unreachable case inspection does not perform correct arthmetics for all types


MDoerner:
The unreachable case inspection does not perform correct arithmetic calculations for all types and does not always determine the correct type of arithmetic expressions.

Example:
```vba
Public Sub Something()
    Dim bar As Double
    bar = 22
    Select Case bar
        Case 4
            Debug.Print "First"
        Case "2" + "2"
            Debug.Print "Second"
    End Select
End Sub
```
Output: `Second"
String addition is concatenation.

Inspection results:
![stringadditionfalsepositive](https://user-images.githubusercontent.com/22868956/39189468-99ab816c-47d2-11e8-86bd-1806c2d1e9a2.jpg)

The underlying problems are that the `ParseTreeExpressionEvaluator` assumes that all mathematical operations can be performed using `double` arithmetics and that the declared type of an expression only depends on the input types.

The actual behaviour in VBA can be foun in section 5.6.9 of the VBA specification (see wiki for a link). There you can see that the return type and the type in which the calculations are performed depend on the input types and the operator.

Issue #3969, opened 2018-04-26
Unify expression evaluation back-ends for Unreachable case and the preprocessor


MDoerner:
We currently have two separate back-ends trying to evaluate literals and expressions in VBA code, one used by the preprocessor and one used by the unreachable case inspection. Both do their job reasonably well, but both also have their shortcomings. (See  #3961, #3962, #3963, #3966, #3967.)

My suggestion in this issue is to decide on one of the approaches and then to move that into its own namespace in `Rubberduck.Parsing`. Then this one approach should be fixed to agree with what the VBA specification dictates and adapted to be suitable to serve both clients, the preprocessor and the unreachable case inspection. 

This would allow to maintain only one back-end and not having to fix both. Moreover, this one back-end could be used by other features to determine the type and, if possible, the value of expressions as well. One application coming to mind is the determination of the type of constants without `As Type` statement.

bclothier:
I completely agree. However, IMO, it's large enough to warrant its own namespace, perhaps `Rubberduck.Parsing.Expression` or something like that. The mistake was that it was tied to predecessor and now there's another tied to the USCI when in fact it ought to be more general and thus be useful for other scenarios like one you mention.

MDoerner:
Although I appreciate the work that has gone into the back-end for the Unreachable Case inspection, I would vote to base the unified approach primarily on the one used by the preprocessor. My main reasons for this preference are the following: 

1. The preprocesor actually uses different types implementing a common interface to cover the different VBA types instead of relying on string representations.
2. The preprocessor actually already covers more literal formats and understands that logical operators are bit-wise operators.

Regarding the the comment about its own namespace, that is actually what I have proposed.

MDoerner:
I think there is more than just the tests we can reuse from the UCI back-end. E.g. we could use the logic from the `ParseTreeValue` to provide `IValue` objects from string representations of literals and combinations of a string literal and a declared type, respectively.

bclothier:
Just to record for posterity, the new unified expression engine should ideally make use of the functions imported from the `oleaut32` DLL. The functions are described [here](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/api/oleauto/). The engine should invoke those functions to get the same behavior as we observe with VBA for parsing, conversion and casting of the expressions and its types.

PR #4127, opened 2018-06-22
Unreachable Case Inspection:  follow-up

### LINKED ISSUE #4119: Unreachable case inspection does not understand enums
Consider
```
Option Explicit

Public Enum LogLevel
    TraceLevel = 0
    DebugLevel
    InfoLevel
    WarnLevel
    ErrorLevel
    FatalLevel
End Enum

Public Function LogLevelToString(ByVal level As LogLevel) As String

    Select Case level

        Case LogLevel.DebugLevel
            ' Reachable (as per inspection results)
            LogLevelToString = "DEBUG"

        Case LogLevel.ErrorLevel
            ' Unreachable (as per inspection results)
            LogLevelToString = "ERROR"

        Case LogLevel.FatalLevel
            ' Unreachable (as per inspection results)
            LogLevelToString = "FATAL"

        Case LogLevel.InfoLevel
            ' Unreachable (as per inspection results)
            LogLevelToString = "INFO"

        Case LogLevel.TraceLevel
            ' Reachable (as per inspection results)
            LogLevelToString = "TRACE"

        Case LogLevel.WarnLevel
            ' Unreachable (as per inspection results)
            LogLevelToString = "WARNING"

    End Select

End Function
```
As per my testing, every single case is reachable (i.e. produces the correct output string. Yet RD assumes most cases to be unreachable. (First case has value 1, and RD only deems the one case with value 0 as reachable. Every case with a higher value is considered unreachable.)

Version 2.2.6738.40126
Host Product: Microsoft Office 2010 x86
Host Version: 14.0.7197.5000
Host Executable: WINWORD.EXE
(switched to Word because there are no add-ins or other modules producing inspection noise)

BZngr:
This is a refactored version of the UnreachableCaseInspection.  This version addresses many of the unresolved comments from the initial PR as well as some improvements/corrections:

1. Refactored tests into multiple files - IMO, the single file had become unwieldy.  Also allows better identification of unit tests versus functional tests.
2. Simplified the inspections processing to a single pass - prior design essentially evaluated each RangeClause context a couple times.
3. Adds `Like` expression support.
4. Adds `&` operator support
5. Enhanced/Corrected `Eqv` and `Imp` operator support
6. Flags value range clauses like 'x To y' where x > y.  These RangeClauses are unreachable due to invalid order...but VBE still compiles and executes without errors.
7. Strengthened the handling of RangeClauses containing variables
8. Respects `Option Compare Binary/Text` setting.
9. Resolves #4119
10. Supports String addition (identified as part of in #3962)


Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilterFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+        };
+
+        public static IExpressionFilter Create(string typeName)
+        {
+            if (IntegralNumberExtents.Keys.Contains(typeName))
+            {
+                var integralNumberFilter = new ExpressionFilterIntegral(StringValueConverter.TryConvertString);
+                integralNumberFilter.SetExtents(IntegralNumberExtents[typeName].typeMin, IntegralNumberExtents[typeName].typeMax);
+                return integralNumberFilter;
+            }
```
bclothier: This stuff probably can be done in the derived class's constructor, simplifying the factory and remove the need to set the extents.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeVisitorResults.cs
```diff
@@ -32,6 +33,23 @@ public IParseTreeValue GetValue(ParserRuleContext context)
...
+                    if (child is ParserRuleContext prCtxt  && Contains(prCtxt))
+                    {
+                        results.Add(prCtxt);
+                    }
+                }
+            }
+            return results;
+        }
+
+
```
bclothier: extra whitespace

Rubberduck.Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
```diff
@@ -0,0 +1,653 @@
+﻿using Rubberduck.Parsing.Grammar;
```
bclothier: I believe all files in `Rubberduck.Inspections/Concrete` are near-duplicates of same files in `Rubberduck.StaticCodeAnalysis/Inspections/Concrete`. There are also some differences. Please fix.

BZngr: Oh wow...my bad...really bad....thank you for spotting this.  When I did the merge a while back to the new hierarchy, the initial new files were left behind in the old folder structure...and I missed it - obviously.  Amazing what you no longer see after staring at something for months.

Rubberduck.Parsing/ParserRuleContextExtensions.cs
```diff
@@ -173,6 +174,25 @@ private static TContext GetAncestor_Recursive<TContext>(ParserRuleContext contex
             return opCtxt != null;
         }
 
+        /// <summary>
+        /// Determines if the context's module declares or defaults to 
+        /// Option Compare Binary 
+        /// </summary>
+        public static bool IsOptionCompareBinary(this ParserRuleContext context)
```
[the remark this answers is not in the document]
BZngr: When I wrote this function I was looking at information like this: https://msdn.microsoft.com/en-us/vba/language-reference-vba/articles/option-compare-statement  <- this looks like Option Compare Binary (case sensitive) is the default even for Access.  Am I mis-interpreting this?

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilterFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+            public static long LONGMIN = Int32.MinValue; //- 2147486648;
+            public static long LONGMAX = Int32.MaxValue; //2147486647
+            public static long INTEGERMIN = Int16.MinValue; //- 32768;
+            public static long INTEGERMAX = Int16.MaxValue; //32767
+            public static long BYTEMIN = byte.MinValue;  //0
+            public static long BYTEMAX = byte.MaxValue;    //255
+            public static decimal CURRENCYMIN = -922337203685477.5808M;
+            public static decimal CURRENCYMAX = 922337203685477.5807M;
+            public static double SINGLEMIN = float.MinValue; // -3402823E38;
+            public static double SINGLEMAX = float.MaxValue;  //3402823E38;
```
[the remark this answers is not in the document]
BZngr: I'm treating `double` as essentially limitless since IIRC, when I attempted to enter a number greater than the max value for a `double`, the VBE intervenes.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilterFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        private static Dictionary<string, (long typeMin, long typeMax)> IntegralNumberExtents = new Dictionary<string, (long typeMin, long typeMax)>()
```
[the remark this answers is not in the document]
BZngr: I agree and I tried to do that...the difficulty I ran into, is that the ExpressionFilter is a generic class.  The factory was the only opportunity/place I could assign 'typed' values.  Within the generic classes, I constantly ran up against compile errors like "cannot convert float to type 'T'".  I guess I could experiment with statements like `if(typeof(T) is double){}` within the ExpressionFilter class.

Rubberduck.Parsing/ParserRuleContextExtensions.cs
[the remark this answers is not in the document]
MDoerner: It is the default, but Access will always add `Option Compare Database` to every new module and class.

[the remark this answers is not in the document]
ThunderFrame: Not a grammar issue, but when testing `Compare` be aware that the Immediate window has its own `Compare` setting which becomes the default when used in native functions inside the immediate window, regardless of what the setting is in the activr project/module. IIRC there was an SO question/discussion in chat about it.

BZngr:
Duplicate files and old folder structure are now removed.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilterFactory.cs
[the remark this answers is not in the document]
bclothier: I was thinking that the Min/Max would be made `abstract` in the base, and the derived classes must provide an implementation, then you could hard-code the constant in each derived class.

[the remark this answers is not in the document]
MDoerner: The problem is that the data type of the min and max depend on the type.

Rubberduck.Parsing/ParserRuleContextExtensions.cs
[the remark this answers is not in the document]
bclothier: The issue is that `Option Compare Text` and `Option Compare Database` will give different results. AFAIK, the `Option Compare Text` is simply case insensitive and never ever is accent insensitive. With `Option Compare Database`, it may or may not be accent-sensitive, depending on the collating order of the file. This is likely beyond the PR's scope and is also a host-specific thing. I'm sure that `Option Compare Database` is not legal in any other VBA hosts. Nonetheless, because this is going to end up as an expression engine, so it'll be necessary to handle this edge case. We would also need to review VBA specs to see if `Option Compare Text` will likewise behave differently with a non-English host.

MDoerner:
Thank you very much for the overview. Reading it and having a very short look at the code, three things immediately came to my mind:

1. Are you aware that there has already been an expression evaluation engine in the preprocessor? It seems to implement the derivation a bit closer to the VBA specification, but also does not fully get it right.
2. The `ParseTreeExoressionEvaluator` seems to treat all arithmetic operators as operators between `double` variables. This is problematic because VBA has data types with a higher precision, namely `LongLong` and `Decimal` (declared type is `Currency`).
3. Relational operators seems to always compare the two sides as strings, which is not quite correct when comparing large double values (exponent notation). 

BTW, I like that the different kinds of filter logic now have their own subtype.

BZngr:
@MDoerner  Thank you for taking the time to _work_ through the overview. 

Regarding 1. Looks like I am guilty of reinventing the wheel a bit (...OK...a lot) here.  I looked for inspections and other support code that did what I needed/wanted to do....but missed this body of code completely.  Stopped looking and never went back.  Ugh.
Regarding 2. Yes, the currently implemented math for VBA Currency would lose some accuracy.  I'll revisit the handling of the `Currency` type.
Regarding 3. There are two sets of Relational operators math, the `String` version is used when comparing `Strings`.  Otherwise, the operands are compared as `Double`...which is _better_, but brings us back to comment (2). ;-) 

Thanks for pointing out the existence of the expression evaluation code - I certainly wish I had discovered it months ago before I stopped looking.  

I'm thinking there are a couple paths forward:
1. Close this PR and refactor/rework the inspection to leverage the expression evaluation engine used by the preprocessor.  I'm hoping that doing so would completely eliminate the `ParseTreeValue` and `ParseTreeExpressionEvaluator` classes (I'll focus on all the interesting things I learned along the way as I hit the 'Delete' button :-\\).

2. Proceed with this current PR (leaving the PR's expression evaluation code largely as-is) only because this version of the inspection still represents an improvement (IMO) to the version currently in the code base.  Then, an UnreachableCaseInspection 3.0 PR would eventually arrive some time in the future that  leverages the expression evaluation engine used by the preprocessor. (Again...I'll focus on all the interesting things I learned along the way as I hit the 'Delete' button :-\\).  

I'm good with either \#1 or \#2....or \#3 if there is one.

bclothier:
IMO, the merging of expression evaluation engine should happen -- there's an issue for it already (#3969 ) but this is a large project in itself and thus must be its own PR. For that reason, I prefer the 2nd option. As noted in the linked issue, there are some difference and we need to merge the best parts from each engine. I'd rather use the improved version as the base for the merging.

MDoerner:
I very much prefer option 2. Using a combined evaluation engine would be nice. However, this PR improves more parts of the inspection. In particular, the range and filter handling gets improved, which will not be part of the expression evaluation engine.

Moreover, the expression evaluation from the preprocessor will also need some improvements because some operations depend both on the value type and the declared type; the latter is missing atm.

I will have a look at the current PR in the coming days.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeValueVisitor.cs
```diff
@@ -87,31 +97,37 @@ private bool ContextHasResult(ParserRuleContext context)
...
-            else if  (IsUnaryLogicalContext(parserRuleContext) || IsUnaryMathContext(parserRuleContext))
+            switch (parserRuleContext)
             {
-                VisitUnaryOpEvaluationContext(parserRuleContext);
+                case VBAParser.LExprContext lExpr:
+                    Visit(lExpr);
+                    return;
+                case VBAParser.LiteralExprContext litExpr:
+                    Visit(litExpr);
+                    return;
```
bclothier: I believe that you can get the same behavior by calling the `base.Visit()`. Maybe call it in the `default` case where none of the other conditions are met, and remove those trivial cases.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeValueVisitor.cs
```diff
@@ -268,31 +294,58 @@ private void GetContextValue(ParserRuleContext context, out string declaredTypeN
...
-            if (identifierReferences.Count() == 1)
+            if (IdRefRetriever != null)
             {
-                idRef = identifierReferences.First();
-                return true;
+                (bool success, IdentifierReference idReference) = IdRefRetriever(context);
+                idRef = idReference;
+                return success;
+            }
+            else if (State != null) //State is set to non-null for testing
```
bclothier: I don't understand why the state is needed only for the testing. Can we avoid having different code path between the production and testing?

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeValueVisitor.cs
BZngr: The overloaded Visit(...) calls like the one identified above, have logic unique to their VBAParser type (or group of types).  I did some experimenting and have yet to find any way to reuse/leverage the `base.Visit()` (by which I think you mean `IParseTreeVisitor<IParseTreeVisitorResults> Visit(IParserTree)` - if you are referring to different `Visit()` , please let me know).

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeValueVisitor.cs
BZngr: Agreed - removed.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
```diff
@@ -0,0 +1,628 @@
...
+            }
+
+            if (!ComparablePredicateFilters.ContainsKey(variable))
+            {
+                ComparablePredicateFilters.Add(variable, ExpressionFilterFactory.Create(variableTypeName));
+                ComparablePredicateFiltersInverse.Add(variable, ExpressionFilterFactory.Create(variableTypeName));
+            }
+        }
+
+        public override bool Equals(object obj)
```
bclothier: Seems redundant; we have a override already defined on the line 47.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
```diff
@@ -0,0 +1,628 @@
...
+                && ComparablePredicates.SetEquals(filter.ComparablePredicates)
+                && LikePredicates.SetEquals(filter.LikePredicates)
+                && this[VariableClauseTypes.Range].SetEquals(filter[VariableClauseTypes.Range])
+                && this[VariableClauseTypes.Value].SetEquals(filter[VariableClauseTypes.Value])
+                && this[VariableClauseTypes.Predicate].SetEquals(filter[VariableClauseTypes.Predicate])
+                && this[VariableClauseTypes.Is].SetEquals(filter[VariableClauseTypes.Is])
+                && Limits.Equals(filter.Limits);
+        }
+
+        public override int GetHashCode()
```
bclothier: Ditto: line 46.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
```diff
@@ -0,0 +1,628 @@
...
+        {
+            return _hashCode;
+        }
+
+        public override string ToString()
```
bclothier: Ditto: line 45

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
```diff
@@ -0,0 +1,628 @@
...
+        {
+            var rangesToRemove = Ranges.Where(rg => Limits.CoversRange(rg));
+            foreach(var range in rangesToRemove)
+            {
+                Ranges.Remove(range);
+            }
+
+        }
+
+        protected void RemoveRangesCoveredByRange((T Start, T End) range)
```
bclothier: Given that we use the `(T Start, T End)` a lot and not just in private, I think it's better to define them as a proper `struct` or `class` object and reference to that instead. 

Furthermore, that would allow you to separate out the operations on the `(T Start, T End)` as its own methods, apart from this class, which would make it more readable and understandable.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
BZngr: The overrides on lines 45 - 47 are for the private `PredicateValueExpression` struct used to store info about comparable predicates (e.g., `x < 5`).  The overrides identified here are for `ExpressionFilter<T>`.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeExpressionEvaluator.cs
```diff
@@ -61,32 +66,79 @@ public class ParseTreeExpressionEvaluator : IParseTreeExpressionEvaluator
...
+            {
+                if (opSymbol.Equals(_ampersand))
+                {
+                    var concatResult = $"{Concat(LHS.ValueText, RHS.ValueText)}";
+                    return _valueFactory.Create(concatResult, Tokens.String);
+                }
+
+                if (opSymbol.Contains(Tokens.Like))
+                {
+                    if (LHS.ParsesToConstantValue && RHS.ParsesToConstantValue)
```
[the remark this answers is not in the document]
BZngr: Good points.  The combination of `foo Like "*"` and `Not foo Like "*"` needs to result in setting values True and False respectively in the `ExpressionFilter<T>`.  Any subsequent relational operator cases ( `Like` or otherwise) would then be flagged as unreachable.  (Note: if the `Select Case Context` is of a type other than `Boolean`, then `Case Else` is still in-play).  I have this fix ready to go on my local machine.

What I realize is missing, is the use case where the `SelectExpressionContext` resolves to a constant value (as in `Select Case True`).  In that use case, the inspection could determine which `CaseClause` (if any) are equal, and flag the rest as unreachable.  Currently, the inspection does not consider the `SelectExpressionContext` value, only its Type.  The constant `SelectExpressionContext` use case is probably rare except for `Select Case True/False` - but handling the `Boolean` case would handle the rest.

As to `foo Like "?*"` making `"foo Like a*"` unreachable - I agree.  I did not wade into the task of evaluating Like patterns for superset/subset effects.  That said, Like patterns that have a variable LHS are retained (indexed by the LHS variable).  The associated Like patterns are evaluated/filtered for equality (finds duplicates) before storing.  The 'TODO' for adding superset/subset evaluations would go there.  I will poke at this enhancement a little...on the surface it looks difficult.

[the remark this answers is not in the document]
BZngr: Updated per above comments.  
1.  Handles use case where a `SelectExpressionContext` resolves to a constant value.  
2.  Set up for `Like` pattern filtering.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ExpressionFilter.cs
BZngr: Replaced `Tuple <T,T>(T Start, T End)` for ranges as RangeOfValues struct....I agree, the change does make the code more readable.

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ParseTreeExpressionEvaluator.cs
```diff
@@ -61,32 +66,79 @@ public class ParseTreeExpressionEvaluator : IParseTreeExpressionEvaluator
...
             Tokens.Byte,
             Tokens.Boolean,
+            Tokens.String
         };
 
-        public ParseTreeExpressionEvaluator(IParseTreeValueFactory valueFactory)
+        public ParseTreeExpressionEvaluator(IParseTreeValueFactory valueFactory, bool isOptionCompareBinary = true)
         {
             _valueFactory = valueFactory;
+            _isOptionCompareBinary = isOptionCompareBinary;
```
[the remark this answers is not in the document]
bclothier: So, if we review the documentation for the particular overload of `string.Compare` ([here](https://msdn.microsoft.com/en-us/library/hyxc48dt(v=vs.110).aspx)), we see a problem - it will consider `Animal` and `ani-mal` equivalent, which is not how VBA works. 

I think we should be using [this overload](https://msdn.microsoft.com/en-us/library/cc190529(v=vs.110).aspx) instead. There, you can pass in both the `CultureInfo` and `CompareOption` to better correspond to the locale that is used by the VBA project. For this PR, you only need to support `CompareOptions.Ordinal` and `CompareOptions.OrdinalIgnoreCase`, while make it easy to update the code to pass in a different `CultureInfo`/`CompareOption` when we do support locale-specific comparison in future. 

Does that help?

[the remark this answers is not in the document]
BZngr: @bclothier thanks.  Yes that helped a lot.  I've updated the string compare portions to:

        `private bool AreEqual(string lhs, string rhs)
        {
            var compareOptions = _isOptionCompareBinary ? 
                StringComparison.CurrentCulture | StringComparison.Ordinal 
                : StringComparison.CurrentCulture | StringComparison.OrdinalIgnoreCase;
            return String.Equals(lhs, rhs, compareOptions);
        }

        private bool IsLessThan(string lhs, string rhs)
        {
            var compareOptions = _isOptionCompareBinary ? CompareOptions.None : CompareOptions.IgnoreCase;
            return String.Compare(lhs, rhs, CultureInfo.CurrentCulture, compareOptions) < 0;
        }

        private bool IsGreaterThan(string lhs, string rhs)
        {
            var compareOptions = _isOptionCompareBinary ? CompareOptions.None : CompareOptions.IgnoreCase;
            return String.Compare(lhs, rhs, CultureInfo.CurrentCulture, compareOptions) > 0;
        }`

Hopefully I've interpreted the documentation correctly.

BZngr:
The latest update/sync addresses the comments to date including some general improvements:
1. `Date` type is now supported (decorates the `DateValue` type from the `Preprocessor` to add `IComparable<T>` support for `ExpressionFilter `ops).
2. Improved the result `Type` determination of all operators and their operands.  The Unary and Binary OperatorDeclaredType combinations from the VBA Language spec are encapsulated within a (new) `OperatorDeclaredTypeProvider` class. 
3.  Refactored/Renamed operator 'groups' (Arithmetic, Relational, Logical) to conform with VBA Language spec groups and nomenclature.

RubberduckCodeAnalysis/RubberduckCodeAnalysis.csproj
```diff
@@ -26,10 +26,11 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <OutputPath>bin\Release\</OutputPath>
+    <OutputPath>bin\Debug\</OutputPath>
```
bclothier: Please revert this change. That is why the build is failing. The DLL must be compiled into Release, regardless of whether we build a debug or a release.

BZngr:
The latest sync refactors `ParseTreeValue `to move type/value evaluation code to dedicated classes and make construction of the `ParseTreeValue `instances more explicit from the start.  (i.e., a `ParseTreeValue `created from an actual value calls a `CreateConstant` ctor rather than a generic ctor.  Previously, the generic ctor was the only option and it had to determine if the parameters represented a constant value.
This latest version also attempts to make explicit the Operator Declared Types, EffectiveValueTypes and the LetCoercions that convert from one value type to another.  Hopefully, these changes move the inspection far closer to compliance with the VBA Language Spec than before.

BZngr:
Sorry this stayed in WIP status for so long.  Introducing the `LetCoerce` class (a good thing I think) resulted in a significant re-evaluation of how errors/invalid tokens were handled.  The result was an approach that supports parsing operations to throw exceptions rather than hiding them all behind the 'TryParse' type functions.  Somewhat unexpected, the changes allowed the inspection to now flag compilable runtime errors (Mismatch and Overflow) that are lurking within the code (within Select Case statements).

Rubberduck.CodeAnalysis/Inspections/Concrete/UnreachableCaseInspection/ComparableDateValue.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            catch (InputMismatchException)
+            {
+                return false;
+            }
+            catch (Exception e)
+            {
+                //even though a SyntaxErrorException/InputMismatchException is thrown, 
+                //this catch-all block seems to be needed(?)
+                return false;
```
retailcoder: Why not just return false given *any* exception then?

PR #4129, opened 2018-06-23
Grammar - fix MID statement

mansellan:
Closes #4111 

There were a few issues:

1. Was using `ASSIGN` instead of `EQ`
2. Was using `lExpression` instead of `expression`
3. Was not referenced from `mainBlockStmt`
4. No tests

### DIFF
--- a/RubberduckTests/Grammar/VBAParserTests.cs
+++ b/RubberduckTests/Grammar/VBAParserTests.cs
@@ -2873,7 +2873,118 @@ End Sub
             var parseResult = Parse(code);
             AssertTree(parseResult.Item1, parseResult.Item2, "//midStatement", matches => matches.Count == 1);
         }
-        
+
+        [Category("Parser")]
+        [Test]
+        public void MidDollarStatement()
+        {
+            const string code = @"
+Public Sub Test()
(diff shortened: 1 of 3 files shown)


RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -2859,6 +2859,21 @@ End Type
             AssertTree(parseResult.Item1, parseResult.Item2, "//commentOrAnnotation", matches => matches.Count == 1);
         }
 
+        [Category("Parser")]
+        [Test]
+        public void MidStatement()
```
bclothier: IMO, we need negative tests, too to assert that an expression that contains a `Mid()` function does not get parsed as a mid statement. We also need equivalent tests for `Mid$()` statement/function. 

For example, we definitely don't want something like `If Mid$("foo", 1) = "f" Then` to be treated as a `Mid` statement. The tests need to validate this.

mansellan:
image

bclothier:
I wonder if it's necessary to have additional unit tests to prove that it's not an undeclared variable to directly test the effect of the hack that was removed?

PR #4131, opened 2018-06-23
Add fakes for Now, Time and Date functions

tommy9:
Adds fakes for the Now, Time and Date functions. Issue #2891 tracks the addition of each item.

Minor issue that integration test for passthrough of Time function could fail for 1 second out of every 24 hours (86 400 seconds). Not considered important enough to add complex logic to avoid this but let me know if this is needed.

### DIFF
--- a/RubberduckTests/IntegrationTests/FakeTests.bas
+++ b/RubberduckTests/IntegrationTests/FakeTests.bas
@@ -272,3 +272,34 @@ TestFail:
     Assert.Fail "Test raised an error: #" & Err.Number & " - " & Err.Description
 End Sub
 
+'@TestMethod
+Public Sub TimeFakeWorks()
+    On Error GoTo TestFail
+
+    With Fakes.Time
+        .Returns #9:00:00 AM#
+        Assert.IsTrue Time = #9:00:00 AM#
+        .Verify.Once
+    End With
+TestExit:
+    Exit Sub
+TestFail:
+    Assert.Fail "Test raised an error: #" & Err.Number & " - " & Err.Description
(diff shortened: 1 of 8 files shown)


RubberduckTests/IntegrationTests/FakeTests.bas
```diff
@@ -238,3 +238,100 @@ TestExit:
...
+End Sub
+
+'@TestMethod
+Public Sub TimeFakePassThroughWorks()
+    On Error GoTo TestFail
+
+    With Fakes.Time
+        .Returns #9:00:00 AM#
+        .Passthrough = True
+        Assert.IsTrue Time <> #9:00:00 AM#
```
Vogel612: this unit-test only works if it's not actuall 9AM. that's probably fine, though, soo... not much to do there

RubberduckTests/IntegrationTests/FakeTests.bas
```diff
@@ -238,3 +238,100 @@ TestExit:
...
+    Assert.Fail "Test raised an error: #" & Err.Number & " - " & Err.Description
+End Sub
+
+
+'@TestMethod
+Public Sub DateFakePassThroughWorks()
+    On Error GoTo TestFail
+
+    With Fakes.Date
+        .Returns #9:00:00 AM#
```
Vogel612: this doesn't match the checked date. IIUC this should be `.Returns #1/1/1993#` to test the Passthrough

tommy9: Yep, will fix

PR #4175, opened 2018-07-12
Add user defined compilation arguments to the preprocessor

WaynePhillipsEA:
This PR pushes user defined compilation arguments (from the VBE) into the preprocessor.

TODO:  read the arguments once at the start of parsing, rather than have each parse thread grab its own version (as it requires a context switch to the main thread, which is relatively slow).

TODO:  before starting to parse, check if the arguments have changed, and force a parse of all modules in order to avoid caching issues.


Rubberduck.Parsing/Preprocessing/VBAPreprocessor.cs
```diff
@@ -16,17 +21,31 @@ public VBAPreprocessor(double vbaVersion)
...
             var expr = evaluator.Visit(tree);
             var processedTokens = expr.Evaluate(); //This does the actual preprocessing of the token stream as a side effect.
             tokenStream.Reset();
         }
+
+        public Dictionary<string, short> GetUserDefinedCompilationArguments(IVBProject project)
+        {
+            // use the TypeLib API to grab the user defined compilation arguments.  must be obtained on the main thread.
+            var providerInst = UiContextProvider.Instance();
+            var task = (new UiDispatcher(providerInst)).StartTask(delegate () {
```
MDoerner: Why new up a a `UiDispatcher` here? Wouldn't it be better to constructor inject an `ICompilationArgumentsProvider` into the preprocessor into which we constructor inject the `UiDispatcher`?

Rubberduck.Parsing/VBA/ComponentParseTask.cs
```diff
@@ -181,7 +181,7 @@ private CommonTokenStream RewriteAndPreprocess(CancellationToken cancellationTok
  
             var tokenStreamProvider = new SimpleVBAModuleTokenStreamProvider();
             var tokens = tokenStreamProvider.Tokens(code);
-            _preprocessor.PreprocessTokenStream(_module.Name, tokens, new PreprocessorExceptionErrorListener(_module.ComponentName, ParsePass.CodePanePass), cancellationToken);
+            _preprocessor.PreprocessTokenStream(component.ParentProject, _module.Name, tokens, new PreprocessorExceptionErrorListener(_module.ComponentName, ParsePass.CodePanePass), cancellationToken);
```
MDoerner: This seems to leak an `IVBProject`.

Rubberduck.Parsing/VBA/AttributeParser.cs
```diff
@@ -53,7 +53,7 @@ public AttributeParser(ISourceCodeHandler sourceCodeHandler, Func<IVBAPreprocess
             var tokens = tokenStreamProvider.Tokens(code);
             var preprocessor = _preprocessorFactory();
             var preprocessorErrorListener = new PreprocessorExceptionErrorListener(module.ComponentName, ParsePass.AttributesPass); 
-            preprocessor.PreprocessTokenStream(module.ComponentName, tokens, preprocessorErrorListener, cancellationToken);
+            preprocessor.PreprocessTokenStream(component.ParentProject, module.ComponentName, tokens, preprocessorErrorListener, cancellationToken);
```
MDoerner: This seems to leak an `IVBProject`.

Rubberduck.Parsing/Preprocessing/VBAPreprocessor.cs
```diff
@@ -16,17 +21,34 @@ public VBAPreprocessor(double vbaVersion)
...
         }
+
+        public Dictionary<string, short> GetUserDefinedCompilationArguments(IVBProject project)
+        {
+            // for the mocks, just return an empty dictionary for now
+            if (project == null) return new Dictionary<string, short>();
+
+            // use the TypeLib API to grab the user defined compilation arguments.  must be obtained on the main thread.
+            var providerInst = UiContextProvider.Instance();
+            var task = (new UiDispatcher(providerInst)).StartTask(delegate () {
```
MDoerner: Why new up an `UiDispatcher` here? Wouldn't it be cleaner to constructor inject an `ICompilationArgumentsProvider` into which we constructor inject the `IUiDispatcher`?

WaynePhillipsEA:
Thanks @MDoerner.  With regards to the step before parsing:  I fully agree, and I said as much in the description of this WIP PR.

Rubberduck.Parsing/VBA/ComponentParseTask.cs
WaynePhillipsEA: That's a good catch.  I wrongly assumed, from looking at previous code using ParentProject that this was cached (i.e. owned and disposed by the component).

Looking at the RD codebase, there are many accesses to ParentProject without proper disposal.  Mostly from the TypeLib API, where I made the same error, but a few pre-date that.

Rubberduck.Parsing/Preprocessing/VBAPreprocessor.cs
WaynePhillipsEA: Unfortunately, I don't do DI.  Any help here would be appreciated.

WaynePhillipsEA:
as per chat, I'll leave it to @MDoerner to refactor the code to grab the argument values only once before parse when he does his larger scope PR for the preprocessor.

bclothier:
I think this is a ore-existing leak that also needs to be plugged?

https://github.com/rubberduck-vba/Rubberduck/pull/4175/files#diff-aae74206ac3d07c402d95c4dac6ae8edR39

WaynePhillipsEA:
@bclothier they looks like they are cached?  I can't see that they are getting new wrappers, so they should be OK.

WaynePhillipsEA:
^ whilst I was in here, I thought I may as well correct some errors with the handling of builtin constants (VBE7, VBE6, Win64 etc).

PR #4207, opened 2018-07-18
Test for anonymous object access in With blocks. Closes #3343

### LINKED ISSUE #3343: Member not found inspection doesn't fire for With Block variables
The member not found inspection fires for early-bound variables, but not for With Block variables:

```vb
Sub test1()
  
  Dim coll As Collection
  Set coll = New Collection
  
  Debug.Print coll.FooBar 'Member not found
  
  With coll
    Debug.Print .FooBar 'Member not found
  End With

  With New Collection
    Debug.Print .FooBar 'No inspection
  End With

End Sub
```

comintern:


RubberduckTests/Inspections/MemberNotOnInterfaceInspectionTests.cs
```diff
@@ -255,6 +255,70 @@ public void MemberNotOnInterface_Ignored_DoesNotReturnResult()
             }
         }
 
+#region https://github.com/rubberduck-vba/Rubberduck/issues/3343
```
Vogel612: not sure I like the use of region for that...

comintern: I think you're right. It helped when I was debugging them, but I should have taken it out.

Issue #3210, opened 2017-08-02
Indenter indents line numbers/labels that include line continuations


ThunderFrame:
After indent:

```
Sub foo()
 _
    10
 _
    foo _
    : Beep
    
20 bar: Beep
End Sub
```

comintern:
Clarification for my own sanity. The compiler is treating 10 as a line number here, but the IDE is as confused as I am by this syntax - it doesn't follow its formatting rule of pinning the line number to the left margin. 

I can't even figure out how this is compiling by stepping through it with the debugger:
```
    foo _
    : Beep
```
By my read, `foo` should be a recursive call to `Sub foo()` (followed by a Beep after the line combining `:`) and this should never exit.

It seems to be interpreted like this:
```
Sub foo()
10
foo: Beep
20 bar: Beep
End Sub
```
Note that this is a parse error too.

ThunderFrame:
No, `foo` is a line label on the line after the line number, then an instruction separator, then the beep statement. i.e. statements with line-numbers 10 and 20 are identical, and should be read as:

```vb
Sub foo()
10 foo: Beep    
20 bar: Beep
End Sub
```

MDoerner:
I have no idea how the compiler interprets the code example. Per spec, line labels must be at the start of the logical line.

ThunderFrame:
IIUC, when determining the logical line, the VBE treats the physical line continuation as if it is:

`" *_ *\n *"`

and then treats line-continuations as if they don't exist, so, to the VBE, the line number _looks_ like it **is** at the start of the logical line.

comintern:
I'm still not exactly clear on why the `foo:` label is on line 10.

This outputs 10, but I'm not sure if that's because of the behavior of `Erl` (it always just returns the last number encountered) or the compile itself:

```
Sub foo()
On Error GoTo bar
 _
    10
 _
    foo _
    : Debug.Print 1 / 0

20 bar: Debug.Print Erl
End Sub
```
This also outputs 10:
```
Sub test()
On Error GoTo label
10
    Debug.Print 1 / 0
label:
    Debug.Print Erl
End Sub
```
Do we think that all whitespace (_including_ newlines) is ignored after a line number?

bclothier:
I'm not so sure.

There is a missing line continuation for the line where `10` is.

This works:

```
Sub foo()
 _
    10 _
 _
    foo _
    : Beep
    
20 bar: Beep
End Sub
```

[...]

PR #4208, opened 2018-07-18
Fix minor grammar issues

### LINKED ISSUE #4196: Parse Error : Comment Line with Underscore
I have a VBA project and trying to take advantage of rubberduck features. However, it seems that rubberduck doesn't handle one of the end-ifs in the following function. 

Error Context reads as : mismatched input 'End If' expecting PABS, ANY, ARRAY, CBOOL, CBYTE, CCUR, ...}

I have also isolated only the function below into a blank/new workbook and same error happens with rubberduck. The source of issue seems to be a commented header line for if condition over line break "_" character ! I tried to remove the comment header with "_" in it and no more error was generated.

P.S. Option Explicit is included in the module in case that might be a question.

image

> Version 2.2.0.3505
> OS: Microsoft Windows NT 10.0.16299.0, x64
> Host Product: Microsoft Office 2016 x86
> Host Version: 16.0.4705.1000
> Host Executable: EXCEL.EXE

```
Option Explicit

'Check for invalid characters in name (return message if any found blank otherwise)
Public Function CheckNameInvalidCharacters(ByVal name As String) As String
    Dim foundInvalidChar As Boolean
    Dim returnMessage As String
    
    returnMessage = "Invalid characters found: "
    foundInvalidChar = False
    'Single quote can't be first character
    If Left(name, 1) = "'" Then
        returnMessage = returnMessage & "' "
        foundInvalidChar = True
    End If
    '=
    If InStr(name, "=") > 0 Then
        returnMessage = returnMessage & "= "
        foundInvalidChar = True
    End If
    'Double quotes "
    If InStr(name, Chr(34)) > 0 Then
        returnMessage = returnMessage & Chr(34) & " "
        foundInvalidChar = True
    End If
    '_                                '<======================= Actual Cause of Error
    '_  Underscore            '<================= No More Error if I add Text after Underscore
    If InStr(name, "_") > 0 Then
        returnMessage = returnMessage & "_ "
        foundInvalidChar = True
    End If                         '<======================= RubberDuck Error Line
    '.
    If InStr(name, ".") > 0 Then
        returnMessage = returnMessage & ". "
        foundInvalidChar = True
    End If
    
    If foundInvalidChar Then
        CheckNameInvalidCharacters = returnMessage
    Else
        CheckNameInvalidCharacters = ""
    End If
End Function
```

MDoerner:
This PR closes #4196 
Moreover, it fixes the parser part of #3210 
(Unfortunately, I messed up when amending commits so that the commit with the corresponding tests now has the wrong commit message.) 

The fix for the first issue is simply to make the leading whitespace in a `LINE_CONTINUATION` mandatory.
The fix for the other issue is kind of a hack and I would welcome better solutions. I allowed line numbers and labels right after line continuations as well. This is less strict than the VBE, but it should not cause any problems. Then the problem is that there usually is whitespace after the `LINE_CONTINUATION`, which the VBE seems to count as part of the line continuation. (This is all part of the preceding `endOfStatement`.) So I added the trailing whitespace to the `LINE_CONTINUATION` unless it is followed by something looking like a line continuation, since consuming all whitespace in front of a line continuation breaks it.

In addition, this PR makes all but the test projects dependent on the code analysis project to avoid failing builds due to a wrong build order.

### DIFF
--- a/RubberduckTests/Grammar/VBAParserTests.cs
+++ b/RubberduckTests/Grammar/VBAParserTests.cs
@@ -550,7 +550,7 @@ Attribute view.VB_VarHelpID = -1
         public void TestEmptyModule()
         {
             string code = @"
-_
+ _
 
    _
 
@@ -1748,7 +1748,7 @@ Sub Test()
                     a = m + n + c + d
                         For k = 0 To 100
                             t = a + k
-    Next k, d, m,_
+    Next k, d, m, _
             c, _
             n
 End Sub";
@@ -2912,7 +2912,7 @@ Sub Test()
 Dim dict As Scripting.Dictionary
 
 Debug.Print dict _
-._
+. _
 Item(""a"")
 End Sub
 ";
@@ -2928,7 +2928,7 @@ Sub Test()
 Dim dict As Scripting.Dictionary
 
 Debug.Print dict _
-    ._
+    . _
 Item(""a"")
 End Sub
 ";
@@ -3012,7 +3012,6 @@ End Sub
 ";
             var parseResult = Parse(code);
         }
-
 
         [Category("Parser")]
         [Test]
--- a/RubberduckTests/Grammar/VBAParserTests.cs
+++ b/RubberduckTests/Grammar/VBAParserTests.cs
@@ -2815,6 +2815,207 @@ End Sub
 
         [Category("Parser")]
         [Test]
+        public void ParserDoesNotFailOnUnderscoreComment()
+        {
+            const string code = @"
+Sub Test()   
+    '_
+    If True Then
+    End If
+End Sub
+";
+            var parseResult = Parse(code);
+            AssertTree(parseResult.Item1, parseResult.Item2, "//ifStmt", matches => matches.Count == 1);
+        }
+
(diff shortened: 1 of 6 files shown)


RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -2813,6 +2813,290 @@ End Sub
...
+        [Test]
+        public void ParserDoesNotFailOnLineNumberAndLineContinuedLabelNotOnStartOfLineAfterALineContinuation()
+        {
+            const string code = @"
+Sub foo()
+ _
+    10
+ _
+    foo _
+    : Beeb
```
Vogel612: inconsequential typo

RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -2813,6 +2813,290 @@ End Sub
...
+Sub foo()
+ _
+ _
+ _
+    10
+ _
+ _
+ _
+    foo _
+    : Beeb
```
Vogel612: same inconsequential typo

Issue #2775, opened 2017-03-02
COM reflection should resolve AsTypeNames for function return types as the alias identifier.


comintern:
Just noticed this.  Given this code:
```
Private Sub Foo()
    Debug.Print StrPtr("foo")
End Sub
```
The `Declaration` for `StrPtr` has an `AsTypeName` of "LONG_PTR", which is defined as an alias:

![LONG_PTR in command bar](https://cloud.githubusercontent.com/assets/8944005/23521513/f5b99994-ff44-11e6-97f7-fcf62ae2c3d4.png)

This is probably just a UI issue more than anything, given that you can `Dim bar As LONG_PTR` and VBA can't tell the difference, but the `AsTypeName` should probably match the type displayed in the IntelliSense:

![LongPtr in IntelliSense](https://cloud.githubusercontent.com/assets/8944005/23521487/d925c24e-ff44-11e6-8561-ab98d03a6f7a.png)

PR #4211, opened 2018-07-19
Clean up COM Collector

comintern:
Getting ready to address some more type naming issues - see #2775 and #1328. Also did some small (but surprisingly effective) performance tweaks.

### DIFF
--- a/Rubberduck.Parsing/ComReflection/ComField.cs
+++ b/Rubberduck.Parsing/ComReflection/ComField.cs
@@ -1,127 +1,23 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Runtime.InteropServices;
+﻿using System.Diagnostics;
 using System.Runtime.InteropServices.ComTypes;
 using Rubberduck.Parsing.Symbols;
-using Rubberduck.VBEditor.Utility;
-using TYPEATTR = System.Runtime.InteropServices.ComTypes.TYPEATTR;
-using TYPEDESC = System.Runtime.InteropServices.ComTypes.TYPEDESC;
-using TYPEKIND = System.Runtime.InteropServices.ComTypes.TYPEKIND;
 using VARDESC = System.Runtime.InteropServices.ComTypes.VARDESC;
 using VARFLAGS = System.Runtime.InteropServices.ComTypes.VARFLAGS;
 
 namespace Rubberduck.Parsing.ComReflection
 {
-    [DebuggerDisplay("{Name}")]
-    public class ComField
+    [DebuggerDisplay("{" + nameof(Name) + "}")]
+    public class ComField : ComTypedElement
     {
-        public string Name { get; }
-        public int Index { get; }
-        public DeclarationType Type { get; }
-        public object DefaultValue { get; }
-        public bool IsReferenceType { get; private set; }
-
-        private string _valueType = "Object";
-        public string ValueType => IsArray ? $"{_valueType}()" : _valueType;
-
-        private Guid _enumGuid = Guid.Empty;
-        public bool IsEnumMember => !_enumGuid.Equals(Guid.Empty);
-
-        public bool IsArray { get; private set; }
+        public override DeclarationType Type { get; }
         public VARFLAGS Flags { get; }
 
-        public ComField(ITypeInfo info, string name, VARDESC varDesc, int index, DeclarationType type)
+        public ComField(ITypeInfo info, VARDESC varDesc, string name, int index, DeclarationType type) : base(info, varDesc, name, index, type == DeclarationType.Constant)
         {
-            Name = name;
-            Index = index;
             Type = type;
-
             Flags = (VARFLAGS)varDesc.wVarFlags;
-
-            if (Type == DeclarationType.Constant)
-            {
-                var value = new ComVariant(varDesc.desc.lpvarValue);
-                DefaultValue = value.Value;
-
-                if (ComVariant.TypeNames.TryGetValue(value.VariantType, out string typeName))
-                {
-                    _valueType = typeName;
-                }
-
-                if (value.VariantType.HasFlag(VarEnum.VT_ARRAY))
-                {
-                    IsArray = true;
-                }
-            }
-            else
-            {
-                GetFieldType(varDesc.elemdescVar.tdesc, info);
-                if (!IsEnumMember || !ComProject.KnownEnumerations.TryGetValue(_enumGuid, out ComEnumeration enumType))
-                {
-                    return;
-                }
-                var member = enumType.Members.FirstOrDefault(m => m.Value == (int)DefaultValue);
-                _valueType = member != null ? member.Name : _valueType;
-            }
         }
 
-        private static readonly HashSet<TYPEKIND> ReferenceTypeKinds = new HashSet<TYPEKIND>
-        {
-            TYPEKIND.TKIND_DISPATCH,
-            TYPEKIND.TKIND_COCLASS,
-            TYPEKIND.TKIND_INTERFACE
-        };
-
-        private void GetFieldType(TYPEDESC desc, ITypeInfo info)
-        {
-            var vt = (VarEnum)desc.vt;
-            TYPEDESC tdesc;
-
-            switch (vt)
-            {
-                case VarEnum.VT_PTR:
-                    tdesc = (TYPEDESC)Marshal.PtrToStructure(desc.lpValue, typeof(TYPEDESC));
-                    GetFieldType(tdesc, info);
-                    break;
-                case VarEnum.VT_USERDEFINED:
-                    int href;
-                    unchecked
-                    {
-                        href = (int)(desc.lpValue.ToInt64() & 0xFFFFFFFF);
-                    }
-                    try
-                    {
-                        info.GetRefTypeInfo(href, out ITypeInfo refTypeInfo);
-                        refTypeInfo.GetTypeAttr(out IntPtr attribPtr);
-                        using (DisposalActionContainer.Create(attribPtr, refTypeInfo.ReleaseTypeAttr))
-                        {
-                            var attribs = Marshal.PtrToStructure<TYPEATTR>(attribPtr);
-                            if (attribs.typekind == TYPEKIND.TKIND_ENUM)
-                            {
-                                _enumGuid = attribs.guid;
-                            }
-                            IsReferenceType = ReferenceTypeKinds.Contains(attribs.typekind);
-                            _valueType = new ComDocumentation(refTypeInfo, -1).Name;
-                        }
-                    }
-                    catch (COMException) { }
-                    break;
-                case VarEnum.VT_SAFEARRAY:
-                case VarEnum.VT_CARRAY:
-                case VarEnum.VT_ARRAY:
-                    tdesc = Marshal.PtrToStructure<TYPEDESC>(desc.lpValue);
-                    GetFieldType(tdesc, info);
-                    IsArray = true;
-                    break;
-                default:
-                    if (ComVariant.TypeNames.TryGetValue(vt, out string result))
-                    {
-                        _valueType = result;
-                    }
-                    break;
-            }
-        }
+        public ComField(ITypeInfo info, TYPEDESC desc) : base(info, desc) { }
     }
 }
--- a/Rubberduck.Parsing/ComReflection/ComModule.cs
+++ b/Rubberduck.Parsing/ComReflection/ComModule.cs
@@ -12,6 +12,7 @@ using CALLCONV = System.Runtime.InteropServices.ComTypes.CALLCONV;
 
 namespace Rubberduck.Parsing.ComReflection
 {
+    [DebuggerDisplay("{" + nameof(Name) + "}")]
     public class ComModule : ComType, IComTypeWithMembers, IComTypeWithFields
     {
         private readonly List<ComMember> _members = new List<ComMember>();
@@ -52,7 +53,7 @@ namespace Rubberduck.Parsing.ComReflection
                 info.GetNames(desc.memid, names, names.Length, out int length);
                 Debug.Assert(length == 1);
 
-                _fields.Add(new ComField(info, names[0], desc, index, DeclarationType.Constant));
+                _fields.Add(new ComField(info, desc, names[0], index, DeclarationType.Constant));
                 info.ReleaseVarDesc(varPtr);
             }
         }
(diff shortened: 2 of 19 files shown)


Rubberduck.Parsing/ComReflection/ComField.cs
```diff
@@ -1,127 +1,23 @@
...
         public VARFLAGS Flags { get; }
 
-        public ComField(ITypeInfo info, string name, VARDESC varDesc, int index, DeclarationType type)
+        public ComField(ITypeInfo info, VARDESC varDesc, string name, int index, DeclarationType type) : base(info, varDesc, name, index, type == DeclarationType.Constant)
         {
-            Name = name;
-            Index = index;
             Type = type;
-
             Flags = (VARFLAGS)varDesc.wVarFlags;
```
bclothier: This seems to be the only use of the `varDesc`. Why not just pass in the `wVarFlags` instead of `VarDesc`? I think I'd prefer to avoid passing type structure across methods when it's not absolutely essential.

Rubberduck.Parsing/ComReflection/ComModule.cs
```diff
@@ -52,7 +53,7 @@ private void GetComFields(ITypeInfo info, TYPEATTR attrib)
                 info.GetNames(desc.memid, names, names.Length, out int length);
                 Debug.Assert(length == 1);
 
-                _fields.Add(new ComField(info, names[0], desc, index, DeclarationType.Constant));
+                _fields.Add(new ComField(info, desc, names[0], index, DeclarationType.Constant));
                 info.ReleaseVarDesc(varPtr);
```
bclothier: Potential to leak pointer on an exception. Should use dispoal action container.

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -1,135 +1,53 @@
...
+        public ComParameter(ITypeInfo info, ELEMDESC elemDesc, string name, int index) : base(info, elemDesc, name, index)
         {
-            Debug.Assert(name != null, "Parameter name is null");
-
-            Name = name;
             var paramDesc = elemDesc.desc.paramdesc;
-            GetParameterType(elemDesc.tdesc, info);
             IsOptional = paramDesc.wParamFlags.HasFlag(PARAMFLAG.PARAMFLAG_FOPT);
             IsReturnValue = paramDesc.wParamFlags.HasFlag(PARAMFLAG.PARAMFLAG_FRETVAL);
+            IsByRef = (VarEnum) elemDesc.tdesc.vt == VarEnum.VT_PTR;
```
bclothier: Isn't `VarEnum` used like a bitmask and thus needs to be `And'`d here?

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
```diff
@@ -0,0 +1,165 @@
...
+            _type = value.TypeName;
+            IsArray = value.VariantType.HasFlag(VarEnum.VT_ARRAY);
+        }
+
+        private void GetFieldType(TYPEDESC desc, ITypeInfo info)
+        {
+            // ReSharper disable once SwitchStatementMissingSomeCases  <-- Don't care, the processed cases are the exceptions, default is the rule.
+            switch ((VarEnum)desc.vt)
+            {
+                case VarEnum.VT_PTR:
```
bclothier: Just checking - according to [this doc](https://docs.microsoft.com/en-us/windows/desktop/api/wtypes/ne-wtypes-varenum), `TYPEDESC` doesn't use `VT_PTR`. Does this case get hit? (I'd think it's a documentation error but want to be absolutely sure). I also have same question regarding whether it'll be used as a bitmask.

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -39,7 +42,6 @@ public class ComVariant : IEquatable<ComVariant>
             {VarEnum.VT_R8, Tokens.Double}
         };
 
-
         [StructLayout(LayoutKind.Sequential)]
         private struct Variant
```
bclothier: FYI, we seem to have a similar definition for `Variant` here: 
https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeLibsSupport.cs#L69

Probably want to unify the COM structures.

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -53,11 +55,12 @@ private struct Variant
 
         public VarEnum VariantType { get; }
         public object Value { get; }
+        public string TypeName => TypeNames.TryGetValue(VariantType, out var typeName) ? typeName : "Object";
 
         public ComVariant(IntPtr variant)
         {
             Value = Marshal.GetObjectForNativeVariant(variant);
-            var members = (Variant)Marshal.PtrToStructure(variant, typeof(Variant));
+            var members = Marshal.PtrToStructure<Variant>(variant);
```
bclothier: Leaking a pointer here?

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
comintern: I thought it does get hit, but I'd need to verify - that's shared with the PARAMDESC type resolution, so I'm not 100% sure that there would be another way to signify a parameter that took a pointer.

Rubberduck.Parsing/ComReflection/ComVariant.cs
comintern: I was looking at that last night thinking the same same thing. IIR, my rationale for doing that was that the Variant structures in the TypeInfo aren't really in working memory and they'll always be either intrinsic types or contain null data pointers. But... the only reason that's in there is to get default values for things like constants and parameters, there probably isn't a reason to hang onto the object.

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Yep. That needs to be fixed.

Rubberduck.Parsing/ComReflection/ComModule.cs
comintern: Damnit! I thought I found all of those. Good spot.

comintern:
Note that e7e2593 only implements variant types that we should be seeing via its current use in COM collection. It is not intended to be a comprehensive COM -> VBA marshaller at this time.

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
```diff
@@ -116,49 +116,49 @@ private void GetValue(IntPtr lpvarValue)
...
-                    GetFieldType(Marshal.PtrToStructure<TYPEDESC>(desc.lpValue), info);
-                    IsArray = true;
-                    break;
-                default:
-                    _type = ComVariant.TypeNames.TryGetValue((VarEnum)desc.vt, out var typeName) ? typeName : "Object";
-                    break;
+                }
+                catch (COMException) { }
+            }
+            else if (Convert.ToBoolean(desc.vt & (int)VarEnum.VT_SAFEARRAY) ||
```
WaynePhillipsEA: and here, but note the handling for VT_CARRAY and VT_ARRAY on the next line is correct, as they are proper flags.

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -37,7 +38,7 @@ public ComParameter(ITypeInfo info, ELEMDESC elemDesc, string name, int index) :
             var paramDesc = elemDesc.desc.paramdesc;
             IsOptional = paramDesc.wParamFlags.HasFlag(PARAMFLAG.PARAMFLAG_FOPT);
             IsReturnValue = paramDesc.wParamFlags.HasFlag(PARAMFLAG.PARAMFLAG_FRETVAL);
-            IsByRef = (VarEnum) elemDesc.tdesc.vt == VarEnum.VT_PTR;
+            IsByRef = Convert.ToBoolean(elemDesc.tdesc.vt & (int)VarEnum.VT_PTR);
```
WaynePhillipsEA: I believe this test should be more like:   vt & VT_TYPEMASK == VT_PTR

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
```diff
@@ -116,49 +116,49 @@ private void GetValue(IntPtr lpvarValue)
...
-                    int href;
-                    unchecked
-                    {
-                        href = (int)(desc.lpValue.ToInt64() & 0xFFFFFFFF);
-                    }
-                    try
+                // ReSharper disable once TailRecursiveCall  <-- A stack allocation is *not* the heavy part of this call, the marshal is. Micro optimization at best.
+                GetFieldType(Marshal.PtrToStructure<TYPEDESC>(desc.lpValue), info);
+            }
+            else if (Convert.ToBoolean(desc.vt & (int)VarEnum.VT_USERDEFINED))
```
WaynePhillipsEA: and here

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
```diff
@@ -116,49 +116,49 @@ private void GetValue(IntPtr lpvarValue)
 
         private void GetFieldType(TYPEDESC desc, ITypeInfo info)
         {
-            // ReSharper disable once SwitchStatementMissingSomeCases  <-- Don't care, the processed cases are the exceptions, default is the rule.
-            switch ((VarEnum)desc.vt)
+            if (Convert.ToBoolean(desc.vt & (int)VarEnum.VT_PTR))
```
WaynePhillipsEA: same here

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -59,12 +88,79 @@ private struct Variant
...
+
+            // Note that these are technically flags, but it they are VT_BYREF or VT_ARRAY, the data area will always be a pointer.
+            // ReSharper disable once SwitchStatementMissingSomeCases
+            switch ((VarEnum)members.vt)
             {
-                Value = string.Empty;
+                case VarEnum.VT_VOID:
+                case VarEnum.VT_EMPTY:
+                case VarEnum.VT_NULL:
+                case VarEnum.VT_ERROR:
```
bclothier: Just checking - are we losing out any interesting information for the COM collector (e.g. default value for optional parameters?) here?

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -59,12 +88,79 @@ private struct Variant
...
+                    Value = members.iVal;
+                    break;
+                case VarEnum.VT_R4:
+                    Value = members.fltVal;
+                    break;
+                case VarEnum.VT_R8:
+                    Value = members.dblVal;
+                    break;
+                case VarEnum.VT_CY:
+                    Value = members.dblVal * 10000;
```
bclothier: This feels iffy to me. 1, it's not clear what the resulting type will be -- if it get converted to a `double` then that would be bad because it's supposed to be precise. 2, I would rather use a proper conversion function -- perhaps `Var***FromCy` in the oleaut32 function. IDK if `VarDecFromCy` would be easily converted to the `decimal` data type but that would be the most faithful conversion, I think.

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -59,12 +88,79 @@ private struct Variant
...
+                    Value = members.uiVal;
+                    break;
+                case VarEnum.VT_I8:
+                    Value = members.lVal;
+                    break;
+                case VarEnum.VT_UI8:
+                    Value = members.ulVal;
+                    break;
+                case VarEnum.VT_INT:
+                    Value = IntPtr.Size == 4 ? members.iVal : members.lVal;
```
bclothier: Shouldn't you be using `IsProcess32Bit` here? Keep it DRY.

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -59,12 +88,79 @@ private struct Variant
...
+                    break;
+                case VarEnum.VT_UINT:
+                    Value = IntPtr.Size == 4 ? members.uiVal : members.ulVal;
```
bclothier: Ditto RE: DRY here

Rubberduck.Parsing/ComReflection/ComVariant.cs
```diff
@@ -42,15 +42,44 @@ public class ComVariant : IEquatable<ComVariant>
...
+            public readonly ulong ulVal;
+            [FieldOffset(8)]
+            public readonly float fltVal;
+            [FieldOffset(8)]
+            public readonly double dblVal;
+            [FieldOffset(8)]
+            public readonly bool boolVal;
+            //end union
+
+            public IntPtr ToPointer() => IntPtr.Size == 4 ? new IntPtr(iVal) : new IntPtr(lVal);
```
bclothier: Ditto RE: DRY

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
comintern: That's basically a shortcut on my part - the reason I wasn't filtering for the basic types was because anything that is a pointer should be recursing until it is a basic type.

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Fixed.

Rubberduck.Parsing/ComReflection/ComTypedElement.cs
comintern: Fixed.

Rubberduck.Parsing/ComReflection/ComVariant.cs
comintern: I'm not exactly sure what you would pass in VBA as an equivalent if you were explicitly passing in the default other than `Null`.

Rubberduck.Parsing/ComReflection/ComVariant.cs
comintern: IIR there's an Interop type for `CoCurrency` or something like that, but I can't imagine what it's doing differently. My understanding is that VT_CY is simply a VT_R8 scaled by 10000. I *do* know that it shouldn't be `decimal`.

Rubberduck.Parsing/ComReflection/ComVariant.cs
comintern: Yes. I completely forgot that I'd defined that last night.

Rubberduck.Parsing/ComReflection/ComVariant.cs
comintern: Oh, crap - I completely misremembered that. I'll see if I can dig up a managed equivalent and if not import `VarDecFromCy`.

comintern:
Closing to flatten history - came at this from a different direction.

PR #4228, opened 2018-07-24
Initial work towards converting output to an actual tree structure, r…

### LINKED ISSUE #2775: COM reflection should resolve AsTypeNames for function return types as the alias identifier.

comintern:
…esolve enums and aliases lazily.

Note: Partial implementation - there are still some corner cases to cover and `ComAlias` needs to have a (potential) recursive `TypeName` member. That's @future though, WIP pending anyone's reviews and (more importantly), reserialization of the test libs.

Closes #2775


Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -87,48 +93,60 @@ private void GetParameterType(TYPEDESC desc, ITypeInfo info)
...
+                tdesc = Marshal.PtrToStructure<TYPEDESC>(desc.lpValue);
+                GetParameterType(tdesc, info);
+                IsByRef = true;
+            }
+            else if (vt == VarEnum.VT_USERDEFINED)
+            {
+                int href;
+                unchecked
+                {
+                    href = (int)(desc.lpValue.ToInt64() & 0xFFFFFFFF);
```
bclothier: I think a comment is needed here to explain what you are doing here.

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -87,48 +93,60 @@ private void GetParameterType(TYPEDESC desc, ITypeInfo info)
...
+                {
+                    info.GetRefTypeInfo(href, out ITypeInfo refTypeInfo);
+                    refTypeInfo.GetTypeAttr(out IntPtr attribPtr);
+                    using (DisposalActionContainer.Create(attribPtr, refTypeInfo.ReleaseTypeAttr))
                     {
-                        info.GetRefTypeInfo(href, out ITypeInfo refTypeInfo);
-                        refTypeInfo.GetTypeAttr(out IntPtr attribPtr);
-                        using (DisposalActionContainer.Create(attribPtr, refTypeInfo.ReleaseTypeAttr))
+                        var attribs = Marshal.PtrToStructure<TYPEATTR>(attribPtr);
+                        var type = new ComDocumentation(refTypeInfo, -1).Name;
```
bclothier: Thinking we should avoid magical number and use a constant `MEMBER_NIL` (or whatever is it that `-1` should be).

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -87,48 +93,60 @@ private void GetParameterType(TYPEDESC desc, ITypeInfo info)
...
-                default:
-                    if (ComVariant.TypeNames.TryGetValue(vt, out string result))
-                    {
-                        _type = result;
-                    }
-                    break;
+                }
+                catch (COMException)
+                {
+                    _typeName = new ComTypeName(Project, "Object");
```
bclothier: `Tokens.Object`?

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -87,48 +93,60 @@ private void GetParameterType(TYPEDESC desc, ITypeInfo info)
...
+            }
+            else if (vt == VarEnum.VT_SAFEARRAY || vt == VarEnum.VT_CARRAY || vt.HasFlag(VarEnum.VT_ARRAY))
+            {
+                tdesc = Marshal.PtrToStructure<TYPEDESC>(desc.lpValue);
+                GetParameterType(tdesc, info);
+                IsArray = true;
+            }
+            else if (vt == VarEnum.VT_HRESULT)
+            {
+                _typeName = new ComTypeName(Project, "Long");
```
bclothier: `Tokens.Long`? Also, I think a comment is needed here as to why we are making it a long (I'm stumbling primarily because I don't expect VB to ever show a `HRESULT`)

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -87,48 +93,60 @@ private void GetParameterType(TYPEDESC desc, ITypeInfo info)
...
+            }
+            else
+            {
+                _typeName = new ComTypeName(Project, (ComVariant.TypeNames.TryGetValue(vt, out string result)) ? result : "Object");
```
bclothier: `Tokens.Object`?

Rubberduck.Parsing/ComReflection/ComProject.cs
```diff
@@ -44,21 +45,15 @@ public class ComProject : ComBase
...
-                //instance where the TypeLib defines a module named "Enums" or "Types".
-                return _modules.Cast<IComType>()
-                    .Union(_interfaces)
-                    .Union(_classes)
-                    .Union(_enumerations)
-                    .Union(_structs);
-            }
-        } 
+        //Note - Enums and Types should enumerate *last*. That will prevent a duplicate module in the unlikely(?)
+        //instance where the TypeLib defines a module named "Enums" or "Types".
```
bclothier: Isn't that an indication of a smelly design? If it falls over flat just because we encountered a module named `Enum`, then something's wrong, no? We should be able to tell an `enum` apart from a module named `enum` simply by looking at the type, no?

Rubberduck.Parsing/ComReflection/ComBase.cs
```diff
@@ -13,6 +13,8 @@ public interface IComBase
         ComDocumentation Documentation { get; }
         string Name { get; }
         DeclarationType Type { get; }
+        IComBase  Parent { get; }
+        ComProject Project { get; }
```
bclothier: I wonder if we'd benefit having a `IEnumerable<IComBase> Children` to support two-way traversal in a general manner?

Rubberduck.Parsing/ComReflection/ComInterface.cs
```diff
@@ -9,11 +9,12 @@
 using FUNCDESC = System.Runtime.InteropServices.ComTypes.FUNCDESC;
 using CALLCONV = System.Runtime.InteropServices.ComTypes.CALLCONV;
 using TYPEFLAGS = System.Runtime.InteropServices.ComTypes.TYPEFLAGS;
+using TYPELIBATTR = System.Runtime.InteropServices.ComTypes.TYPELIBATTR;
 using VARDESC = System.Runtime.InteropServices.ComTypes.VARDESC;
 
 namespace Rubberduck.Parsing.ComReflection
 {
-    [DebuggerDisplay("{Name}")]
+    [DebuggerDisplay("{" + nameof(Name) + "}")]
```
bclothier: Why is this not wrapped in `#If DEBUG` while the other is?

Rubberduck.Parsing/ComReflection/ComField.cs
```diff
@@ -79,48 +83,47 @@ private void GetFieldType(TYPEDESC desc, ITypeInfo info)
...
-                    IsArray = true;
-                    break;
-                default:
-                    if (ComVariant.TypeNames.TryGetValue(vt, out string result))
-                    {
-                        _valueType = result;
-                    }
-                    break;
+                }
+                catch (COMException) { }
```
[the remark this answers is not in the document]
comintern: The generic catch there is precisely to swallow all `COMException`s. I'm fine with trucking along after an `E_FAIL`, but at this point, any HRESULT that isn't wrapped into a managed exception by the RCW is something that I would consider as "recoverable", in that the user type would just resolve to the default "Object" or be missing a default value. I was working on the assumption that it's much more likely that the tlb is corrupt or malformed than anything "unrecoverable".

Rubberduck.Parsing/ComReflection/ComInterface.cs
```diff
@@ -29,14 +30,32 @@ public class ComInterface : ComType, IComTypeWithMembers
...
+                typeLib.GetLibAttr(out IntPtr attribPtr);
+                using (DisposalActionContainer.Create(attribPtr, typeLib.ReleaseTLibAttr))
+                {
+                    var typeAttr = Marshal.PtrToStructure<TYPELIBATTR>(attribPtr);
+                    Parent = typeAttr.guid.Equals(parent?.Project.Guid) ? parent?.Project : null;
+                }
+            }
+            catch
+            {
+                Parent = null;
```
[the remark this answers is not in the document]
comintern: Same thing here. I really don't care if that block throws _anything_ - it just means the Parent doesn't get set (and I have to account for that possibility anyway).

Rubberduck.Parsing/ComReflection/ComParameter.cs
```diff
@@ -12,71 +12,77 @@
...
+
+        public string DefaultAsEnum
+        {
+            get
+            {
+                if (!_typeName.IsEnumMember || !ComProject.KnownEnumerations.TryGetValue(_typeName.EnumGuid, out ComEnumeration enumType))
+                {
+                    return string.Empty;
+                }
+                var member = enumType.Members.FirstOrDefault(m => m.Value == (int)DefaultValue);
```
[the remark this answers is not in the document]
comintern: Yes. In a case where there are members with identical values, there isn't a way to discriminate between them. For example, `vbIMEModeOff` and `vbIMEOff` both have a value of 2. Makes no real difference functionally which one gets picked, because they are equivalent.

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Done, but might not be able to sync it with this PR without pulling in a load of other crap.

`//The href is a long, but the size of lpValue depends on the platform, so truncate it after the lword.`

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Good idea. It doesn't really have a name that I could find in the documentation, so I just called it `ComDocumentation.LibraryIndex`. Save caveat about the sync of unrelated crap.

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Yep, and a couple other places too.

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Above re the token (fixed). The MSXML6 library has a hidden interface that takes an `HRESULT` as a freaking _parameter_, so it has to be handled. Who knew?

Rubberduck.Parsing/ComReflection/ComParameter.cs
comintern: Yep.

Rubberduck.Parsing/ComReflection/ComProject.cs
comintern: Probably lazy design more than anything - building that assumption into the interface makes [this section of code](https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.Parsing/ComReflection/ReferencedDeclarationsCollector.cs#L144) much, much more efficient and terse. The _root_ of that problem (IIR) was that there isn't an `EnumerationDeclaration` - it just uses the generic one because the only thing that differs is the `DeclarationType` .  The assumption has pretty much always been there (even prior to the refactor that split them out I think).  #LetSleepingBearsLie.

Rubberduck.Parsing/ComReflection/ComBase.cs
comintern: That might not be a bad idea @future, I didn't give it much thought because the current use case is always down the tree.

Rubberduck.Parsing/ComReflection/ComInterface.cs
comintern: The other one is calling `nameof` on a function that is also wrapped in `#if DEBUG`. R# suggested the `nameof` change, and I generally think it's a good idea, but it compiles (and ignores?) the annotation in a release build. It's a little silly in how it's designed, because well, debugger?

PR #4233, opened 2018-07-26
Pimp the toolbar (for some values of "pimp")

### LINKED ISSUE #1328: COM reflection should pick up constant values

comintern:
Add default values to enumeration members, consts, and parameters, closes #1328


comintern:
WIP pending unit tests (which VS managed to wipe out...).

Rubberduck.Parsing/Symbols/ParameterDeclaration.cs
```diff
@@ -39,43 +41,51 @@ public class ParameterDeclaration : Declaration
...
             IsByRef = isByRef;
-            IsImplicitByRef = isByRef && ((VBAParser.ArgContext) context)?.BYREF() == null;
+            IsImplicitByRef = isByRef && argContext?.BYREF() == null;
             IsParamArray = isParamArray;
+
+            if (!isOptional || argContext?.argDefaultValue() == null)
+            {
+                return;
+            }
+            DefaultValue = argContext.GetDescendent<VBAParser.LiteralExprContext>()?.GetText() ?? string.Empty;
```
MDoerner: This does not look appropriate. Default members are not restricted to literals, just to compile time constants.

The following is completely legal:
```vba
Private Const bar As Long = 2

Private Enum Foo
   FooBar = bar + 15 - Module1.OtherModuleConst
End Enum


Private Sub DoSomething(Optional baz As Long = bar * 15 + Foo.FooBar + Module1.OtherModuleConst * FooBar)
   Debug.Print baz
End Sub
```
Here, `OtherModuleConst` is a public constant declared in `Module1`.

Until we can evaluate the expression context, I would propose to use `argContext.argDefaultValue().Expression.GetText()`.

comintern: Not sure why I was thinking it had to be a literal. This is fixed.

Rubberduck.Parsing/Symbols/DeclarationSymbolsListener.cs
```diff
@@ -289,6 +289,22 @@ private void DeclareControlsAsMembers(IVBComponent form)
...
+                        new QualifiedMemberName(_qualifiedModuleName, identifierName),
+                        _parentDeclaration, 
+                        _currentScope, 
+                        asTypeName, 
+                        asTypeContext, 
+                        typeHint, 
+                        annotations,
+                        accessibility, 
+                        declarationType,
+                        context.GetDescendent<VBAParser.ExpressionContext>()?.GetText() ?? string.Empty,
```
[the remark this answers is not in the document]
comintern: Good to know about the performance implications of `GetDescendent`. This is also fixed.

PR #4254, opened 2018-08-07
COM Collector parameter fix, add type names for property [ls]ets

comintern:
No issues opened, [see chat](https://chat.stackexchange.com/transcript/message/46051314#46051314).


Rubberduck.Parsing/ComReflection/ComMember.cs
```diff
@@ -36,10 +36,13 @@ public class ComMember : ComBase
...
-        public ComParameter ReturnType { get; private set; } = ComParameter.Void;
+        public ComParameter AsTypeName { get; private set; } = ComParameter.Void;
 
-        private readonly List<ComParameter> _parameters = new List<ComParameter>();
-        public IEnumerable<ComParameter> Parameters => _parameters;
+        private List<ComParameter> _parameters = new List<ComParameter>();
+
+        //See https://docs.microsoft.com/en-us/windows/desktop/midl/retval
+        //"Parameters with the [retval] attribute are not displayed in user-oriented browsers."
+        public IEnumerable<ComParameter> Parameters => _parameters.Where(param => !param.IsReturnValue);
```
bclothier: Do we have a use cases where we need to know about the parameters from the type lib's POV, rather than from user's POV?

Rubberduck.Parsing/ComReflection/ComMember.cs
```diff
@@ -100,13 +103,24 @@ private void LoadParameters(FUNCDESC funcDesc, ITypeInfo info)
...
                 _parameters.Add(param);
             }
+
+            // See https://docs.microsoft.com/en-us/windows/desktop/midl/propput
+            // "A function that has the [propput] attribute must also have, as its last parameter, a parameter that has the [in] attribute."
+            if (funcDesc.invkind.HasFlag(INVOKEKIND.INVOKE_PROPERTYPUTREF) ||
+                funcDesc.invkind.HasFlag(INVOKEKIND.INVOKE_PROPERTYPUT))
+            {
+                AsTypeName = _parameters.Last();
+                _parameters = _parameters.Take(funcDesc.cParams - 1).ToList();
```
bclothier: I wonder if we really want to use LINQ. As a rule we avoid using LINQ with COM objects, and type lib stuff are COM objects. Are we inadvertently creating leaks this way?

Rubberduck.Parsing/ComReflection/ComMember.cs
comintern: None currently that I can think of off the top of my head, although it would make generating MIDL easier if we decide to do that in the future.

Rubberduck.Parsing/ComReflection/ComMember.cs
comintern: This should be perfectly safe in this context. `ComParameter` is a completely managed type.

PR #4257, opened 2018-08-08
Parser NRE hotfix.

comintern:


comintern:
@MDoerner - You're probably right, although `VBACodeStringParser` is called directly in the unit tests. I just added the minimal coalescing required to successfully parse an empty workbook and pass the unit tests. This PR wasn't intended to be the definitive solution.

PR #4263, opened 2018-08-09
More safety around handling the temp files in the parser

MDoerner:
This PR adds some more safety in the `ISourceFileHandler` (now `ITempSourceFileHandler`) around reading the files. I chose to rename the interface and the implementations to convey that they are not to be used to handle permanent exports and imports. 

I also ensured that the `ISourceCodeHandler` implementations never retun `null` and made the attributes passs use the file exported by the VBE or documant modules in order to allow getting the attributes.

### DIFF
--- a/Rubberduck.API/VBA/Parser.cs
+++ b/Rubberduck.API/VBA/Parser.cs
@@ -97,7 +97,7 @@ namespace Rubberduck.API.VBA
             _state = new RubberduckParserState(_vbe, projectRepository, declarationFinderFactory, _vbeEvents);
             _state.StateChanged += _state_StateChanged;
 
-            var sourceFileHandler = _vbe.SourceFileHandler;
+            var sourceFileHandler = _vbe.TempSourceFileHandler;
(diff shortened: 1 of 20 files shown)


MDoerner:
The problem here is that we need separate interfaces for handling temporary source files for VBA and handling permanent source files for VBA because the handling follows different contracts.

MDoerner:
An alternative to the rename could be a change of the interface and the implementations pushing the responsibility for cleanup to the caller. That would require to add a `Delete` method to the interface, which has to be a no-go in the VB6 implementation, and removing all automatic cleanup from the VBA implementation. Moreover, we would require a new parameter on the `Export` method to indicate whether document modules should receive special treatment.

bclothier:
I want to float a suggestion. What if the export and import specifications were implemented per module type? If I'm not mistaken, in VB6, we will have forms that come in 2 files, `.frm` and `.frx`? Wouldn't that require a different way to handle such thing? By allowing each module type to choose a particular implementation of import/export. We would use a provider to retrieve the correct implementation for a particular module type. Would that work?

PR #4264, opened 2018-08-09
Make parser aware of CDecl, add inspection for its use

### LINKED ISSUE #4167: Parse Error when CDecl specified in Declare statement
It seems the VBA team were planning on adding support for `__cdecl` calling conventions, and allow for `CDecl` to be specified as part of a `Declare` statement, even though it seems that `__cdecl` support never quite got delivered. This statement compiles, but causes an RD parse error:

`Private Declare Sub Beep CDecl Lib "kernel32" (dwFreq As Any, dwDuration As Any)`

The parser needs to allow for these statements, but should also perhaps have an inspection for removing the `CDecl` as it's not implemented by VBA.

comintern:
Closes #4167

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/ObsoleteCallingConventionInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/ObsoleteCallingConventionInspection.cs
@@ -1,11 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Antlr4.Runtime;
 using Rubberduck.Inspections.Abstract;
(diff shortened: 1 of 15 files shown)


rkapka:
Implementing a new inspection requires adding some inspection settings to `app.config` and `Settings.settings`. Maybe the inspection will work without it using default values of settings, but they should be added nonetheless.

comintern:
I'd just assumed that using the implicit defaults was sufficient. Added.

I am curious about the design decision to put these in a settings file that lives outside of the assembly that that the classes they configure reside in. The reason for moving `Rubberduck.SettingsProvider` to its own assembly was to allow individual assemblies to be responsible for their own configurations. `Rubberduck.Core` should not have more information about how to instantiate objects from `Rubberduck.CodeAnalysis` than `Rubberduck.CodeAnalysis` does.

Did this get overlooked when inspections were moved to their own assembly?

PR #4268, opened 2018-08-11
Fix the leaky (event) sink!

### LINKED ISSUE #3949: Wrap VB6 CommandBarEvents
See https://github.com/rubberduck-vba/Rubberduck/pull/3935#discussion_r182843138

Currently, the Office8 CommandBarButton wrapper (as used by VB6) has a chained COM reference on its base constructor call. This will be leaking RCWs. We need to wrap the events interface and add it to IVBE so that the event source can be obtained without chaining.

mansellan:
Closes #3949 

OK, so technically it's an event *source* that was (perhaps) leaking... but I just couldn't resist that PR title.

This PR addresses an item of technical debt spotted during PR #3935. For expediency at the time, VB6 events for CommandBarButtons were hooked up using chained access to raw RCW objects. This was entirely due to the unpleasant way that older Office libraries (such as the v8 that VB6 uses) modelled button click events.

To remediate, I've created some new SafeComWrappers for CommandBarEvents. These are then used to safely pass the correct RCW into a refactored SafeRedirectedEventComWrapper base class. ~~Note that as we have no need for any of this on the VBA side, the relevant property just throws NotSupportedException. I think this is valid, as its something that could potentially be added later, just that there's no current need.~~ Edit: per chat, VBE.Events must not be used in 64-bit VBA for Office 2000 and up, as it will crash the host. 

One oddity was the need to have one RCW wrapped at two levels (CommandBarEvents and CommandBarButtonEvents) - this is due to the COM OM using indexed properties here - this is the first time we've had to wrap them AFAICT.

### DIFF
--- a/Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
@@ -35,15 +35,18 @@ namespace Rubberduck.VBEditor.SafeComWrappers
             if (eventSource == null || eventSink == null)
             {
                 return;
-            }
-
-            // Only one event source and one event sink supported at any given time
-            DetachEvents();
+            }            
 
             lock (_lock)
             {
                 if (IsWrappingNullReference)
                 {
+                    return;
+                }
+                
+                // Check that events not already attached
+                if (_eventSource != null || _eventSink != null)
+                {                   
                     return;
                 }
 
--- a/Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+﻿using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
 using Rubberduck.VBEditor.SafeComWrappers.Abstract;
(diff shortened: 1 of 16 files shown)


Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
```diff
@@ -45,24 +42,31 @@ public void AttachEvents()
...
-
-                var sink = this as TEventInterface;
-
-                if (sink == null)
-                {
-                    throw new InvalidOperationException($"The class {this.GetType()} does not implement the required event interface {typeof(TEventInterface)}");
-                }
-                
-                _icp.Advise(sink, out _cookie);
+                _icp.Advise(_eventSink, out _cookie);
```
mansellan: Cast no longer needed - strongly typed param passed to method.

bclothier:
I think the VBA implementation should be decorated with [this overloaded `Obsolete` attribute](https://msdn.microsoft.com/en-us/library/961hff5d(v=vs.110).aspx) so it becomes a compile time error rather runtime.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/CommandBarButtonEvents.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Rubberduck.VBEditor.SafeComWrappers.VB6
+{
+    public class CommandBarButtonEvents : SafeComWrapper<VB.CommandBarEvents>, ICommandBarButtonEvents, IEventSource<VB.CommandBarEvents>
+    {
+        public CommandBarButtonEvents(VB.CommandBarEvents target, bool rewrapping = false)
+            : base(target, rewrapping)
+        {            
+        }
+
+        public VB.CommandBarEvents EventSource => Target;
```
mansellan: This would be better as an explicit interface implementation to make it less visible.

Rubberduck.VBEEditor/SafeComWrappers/SafeRedirectedEventedComWrapper.cs
```diff
@@ -13,73 +13,80 @@ public abstract class SafeRedirectedEventedComWrapper<TSource, TEventSource, TEv
...
         {
+            Debug.Assert(eventSource != null, "Unable to attach events - eventSource is null");
+            Debug.Assert(eventSink != null, "Unable to attach events - eventSink is null");
+            if (eventSource == null || eventSink == null)
+            {
+                return;
+            }
+
+            // Only one event source and one event sink supported at any given time
+            DetachEvents();
```
bclothier: After thinking about it, I think I don't want that particular change here and on the original line 43-45 below. As discussed, the "last one wins" model would likely do more damage than the current "first one wins" model. The latecomer should be handed a no-op instead so that the first can be assured that it can clean itself up properly.

That said, the requirement to have only one source is separate issue, and we probably could stick in a `Debug.Assert` to warn of accidental re-attaching by a different source, which is most likely a bug that needs to be fixed because we want to minimize the contact between the .NET event consumers and the COM event sink.

PR #4274, opened 2018-08-14
Misc bug fixes

comintern:


Rubberduck.Core/UI/Command/AddTestModuleCommand.cs
```diff
@@ -247,9 +247,14 @@ protected override void OnExecute(object parameter)
...
-            var index = names.Count(n => n.StartsWith(TestModuleBaseName)) + 1;
 
-            return string.Concat(TestModuleBaseName, index);
+            var index = 1;
+            while (names.Contains($"{TestModuleBaseName}{index}"))
+            {
+                index++;
+            }
+
+            return $"{TestModuleBaseName}{index}";
```
Vogel612: this change may need to be duplicated over to `AddTestModuleExpectedErrorCommand.cs`

bclothier: I also think need tests for the cases of `TestModule2`, `TestModule3`, and another for where there's a `TestModule11`. Both should return `TestModule1`, if I understood the changes correctly.

comintern: @Vogel612 I'm assuming you mean the add method commands, and yes, they have the same numbering issue (I didn't think to check those). I'll fix those too.

PR #4276, opened 2018-08-14
InnoSetup Czech Translation

SonGokussj4:
Made a quick translation for InnoSetup. Quick != bad :-)


Vogel612:
this is the same PR as #4275

PR #4277, opened 2018-08-15
Test Explorer fixes

### LINKED ISSUE #3590: Test Explorer shows duration of 0ms for all tests if run by Rubberduck>UnitTests>Run All Tests
Going through `Rubberduck>UnitTests>Run All Tests` produces the image below. Note the 0ms timer for each test.
untitled

Using `TestExplorer>Run>All Tests` produces the times show below.
untitled2

### LINKED ISSUE #4122: Test Explorer>Group by Location duration never updates
Duration always shows `0ms` even when group by outcome shows a non-zero time. Originally a comment in #3590 I "rediscovered" this because I never reported this individually.

rkapka:
Closes  #3590, closes #4122.


Rubberduck.Core/UI/UnitTesting/TestExplorerControl.xaml.cs
```diff
@@ -37,9 +35,17 @@ private void OnTestCompleted(object sender, EventArgs eventArgs)
...
-                    resource.View.Refresh();
+                    outcome.View.Refresh();
+                }
+                if (FindResource("ResultsByModule") is CollectionViewSource module)
+                {
+                    module.View.Refresh();
+                }
+                if (FindResource("ResultsByCategory") is CollectionViewSource category)
+                {
+                    category.View.Refresh();
```
Vogel612: Since these are all collectionViews we should be updating the backing collections to fire a PropertyChanged event that will automatically update the UI... I'll try to make sure that happens in my changes to UnitTesting

Rubberduck.Core/UI/UnitTesting/TestExplorerViewModel.cs
```diff
@@ -262,11 +263,6 @@ private void ExecuteRefreshCommand(object parameter)
             SelectedTest = null;
         }
 
-        private bool CanExecuteRefreshCommand(object parameter)
-        {
-            return !Model.IsBusy && _state.IsDirty();
-        }
```
Vogel612: note to self: carry over this change in #4146

Rubberduck.Main/Root/RubberduckIoCInstaller.cs
```diff
@@ -149,7 +148,10 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
 
             RegisterWindowsHooks(container);
 
-            RegisterHotkeyFactory(container);
+            container.Register(Component.For<HotkeyFactory>()
+                .LifestyleSingleton());
+            container.Register(Component.For<ITestEngine>().ImplementedBy<TestEngine>()
+                .LifestyleSingleton());
```
Vogel612: this registration should already be covered by the interface convention ...

Rubberduck.Core/UI/UnitTesting/TestExplorerViewModel.cs
```diff
@@ -34,7 +34,8 @@ public class TestExplorerViewModel : ViewModelBase, INavigateSelection, IDisposa
              IClipboardWriter clipboard,
              IGeneralConfigService configService,
              ISettingsFormFactory settingsFormFactory,
-             IMessageBox messageBox)
+             IMessageBox messageBox,
+             ReparseCommand reparseCommand)
```
Vogel612: this will break #4146 mergeability. I'll deal with it.

Rubberduck.Main/Root/RubberduckIoCInstaller.cs
rkapka: I believe the default lifestyle is `Transient`, is it not?

MDoerner: It is

rkapka:
This PR fixes #3590, because, as a result of the IoC registration change, the same `TestEngine` instance is shared between `TestExplorerViewModel` and `RunAllTestsCommand`. The former registers a `_testEngine.TestCompleted` handler which updates the UI. That handler will be invoked from the command, so UI gets updated.

Rubberduck.Main/Root/RubberduckIoCInstaller.cs
rkapka: Without this change `ITestEngine` was registered as `Transient`, which caused the #3590 bug.

Vogel612:
:bulb: Thanks, that makes sense, now I understand

Issue #4283, opened 2018-08-17
Problem with AutoCAD VBA 


RICVBA:
I installed RD v 2.2.0.3675 and then tried opening AutoCAD  2014 product version I.18.0.0 VBA IDE with following issues:

- RD splash screen pops out with "Initializing..."

- then it pops out a "Rubberduck Add-In could not be loaded" dialog doc saying: _"Rubberduck's startup sequence threw an unexpected exception. Please check the Rubberduck logs for more information and report an issue if necessary"_ and giving the only "OK" button to click

- upon clicking "OK" button there it opens the VBA IDE showing an "AutoCAD" dialog box saying: _"Unhandled exception has occurred in a component in your application. If you click Continue, the application will ignore this error and attempt to continue.

Value cannot be null."_

and whose "Details" are:

```
System.ArgumentNullException: Value cannot be null.
   at System.Threading.Monitor.Enter(Object obj)
   at Rubberduck.VBEditor.SafeComWrappers.SafeComWrapper`1.Dispose(Boolean disposing) in C:\projects\rubberduck\Rubberduck.VBEEditor\SafeComWrappers\SafeComWrapper.cs:line 144
   at Rubberduck.VBEditor.SafeComWrappers.Abstract.HostApplicationBase`1.Finalize() in C:\projects\rubberduck\Rubberduck.VBEEditor\SafeComWrappers\Abstract\HostApplicationBase.cs:line 135
```

- clicking Continue an "AutoCAD Error Aborting" dialog box pops out saying _"FATAL ERROR: Unhandled e0434352h at 54cb9d98h"_ with only "OK" button to exit

- upon clicking "OK" button everything (VBA IDE, AutoCAD application) shuts down

RICVBA:
and here's the "RubberduckLog.txt]"

[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2298733/RubberduckLog.txt)

MDoerner:
I have my doubts that this really is a regression. As mentioned in chat, the error occured when loading RD in a newer AutoCAD version that does no longer support VBA out of the box via a VBA enabler. I bet the VBA enabler does not register with the same GUID the old original AutoCAD host did. In that case, the original GUID is probably not registerd at all and the cast is failing because of this.

RICVBA:
Rubberduck v2.2.0.3684-pre solved it!
Thanks you all!

PR #4285, opened 2018-08-18
Dirty hack to not blow up for hosts with unexpected COM setups

MDoerner:
This PR tries te address #4283.

Tha hacky solution is to castch the `InvalidCastException`, log it and use `null` as target. Currently, we only ever use the application name anyway, which is saved independently of the underlying COM object.

### DIFF
--- a/Rubberduck.VBEEditor/SafeComWrappers/Abstract/HostApplicationBase.cs
+++ b/Rubberduck.VBEEditor/SafeComWrappers/Abstract/HostApplicationBase.cs
@@ -26,14 +26,21 @@ namespace Rubberduck.VBEditor.SafeComWrappers.Abstract
             {
                 application = (TApplication) Marshal.GetActiveObject($"{applicationName}.Application");
             }
-            catch (COMException)
+            catch (COMException exception)
             {
-                application = null; // unit tests don't need it anyway.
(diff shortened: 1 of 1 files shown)


Rubberduck.VBEEditor/SafeComWrappers/Abstract/HostApplicationBase.cs
```diff
@@ -23,12 +24,18 @@ private static TApplication ApplicationFromComReflection(string applicationName)
...
             try
             {
-                application = (TApplication)Marshal.GetActiveObject($"{applicationName}.Application");
+                application = (TApplication) Marshal.GetActiveObject($"{applicationName}.Application");
             }
             catch (COMException)
             {
                 application = null; // unit tests don't need it anyway.
             }
+            catch (InvalidCastException exception)
```
mansellan: Why not just catch Exception to cover all eventualities?

Rubberduck.VBEEditor/SafeComWrappers/Abstract/HostApplicationBase.cs
```diff
@@ -52,6 +59,11 @@ private static TApplication ApplicationFromVbe(IVBE vbe, string applicationName)
...
```
mansellan: ditto

PR #4287, opened 2018-08-19
Contexts allowing Attributes

MDoerner:
This PR introduces a new field `AttributesPassContext` on the `Declaration` holding the context from the attributes pass for all declarations allowing attributes. I would like to have added this only to a more specific type. However, that would have required some rework on the declaration type hierarchy since we have no `ModuleVariableDeclaration`. (That is something for another PR at some time in the furture.)

The `membersAllowingAttributes` are acquired along with the attributes and stored in an analogous way.

In addition, this PR fixed the key assigned to attributes for module variables and for module attributes inside procedures, functions and properties. 

This PR should provide the necessary foundation for an `AddAttributeCommand`.


Rubberduck.Parsing/VBA/Parsing/AttributeListener.cs
```diff
@@ -52,13 +56,23 @@ public override void ExitSubStmt(VBAParser.SubStmtContext context)
                 _attributes.Add(_currentScope, _currentScopeAttributes);
                 context.AddAttributes(_currentScopeAttributes);
             }
+
+            ResetScope();
+        }
+
+        private void ResetScope()
```
bclothier: I'm not clear why we need to reset the scope? Or is it more like update the scope?

MDoerner: This ensures that we do not pick up procedure attributes outside of their contexts be resetting the current scope and current scope attributes to their inital values.

bclothier: I wonder if it'll help with understanding the intention if we used `PushScope` and PopScope` instead of an assignment statement in the entering rule and a call to `ResetScope` in the exiting rule?

RubberduckTests/Grammar/AttributesTests.cs
```diff
@@ -0,0 +1,1692 @@
...
+                var value = attribute.Values.Single();
+                Assert.AreEqual(expectedAttributeName, attributeName);
+                Assert.AreEqual(expectedAttributeValue, value);
+            }
+        }
+
+        [Category("Grammar")]
+        [Category("Attributes")]
+        [Test]
+        public void VariableAttributeBetweenVariableDeclarationsAndModuleBaodyGetsRecognized()
```
bclothier: Typo -- `ModuleBody` rather than `ModuleBaody`

RubberduckTests/Grammar/AttributesTests.cs
```diff
@@ -0,0 +1,1692 @@
...
+using System.Threading;
+using NUnit.Framework;
+using Rubberduck.Inspections.Concrete;
+using Rubberduck.Parsing.Symbols;
+using RubberduckTests.Mocks;
+
+namespace RubberduckTests.Grammar
+{
+    [TestFixture]
+    public class AttributesTests
```
bclothier: It seems to me that a number of tests, especially those dealing with different placement of the attributes for the function/sub/property{let/get/set} could have been collapsed into a single data-driven method since the `inputCode` is the only thing that's different.

Rubberduck.Parsing/VBA/Parsing/AttributeListener.cs
```diff
@@ -44,98 +44,88 @@ public override void EnterModuleVariableStmt(VBAParser.ModuleVariableStmtContext
...
             {
                 _attributes.Add(_currentScope, _currentScopeAttributes);
                 context.AddAttributes(_currentScopeAttributes);
             }
-
-            ResetScope();
         }
 
-        private void ResetScope()
+        private void PopScope()
```
bclothier: Much more readable! A minor nitpick; don't you think the `PushScope` and `PopScope` should be placed together, rather being few methods apart?

RubberduckTests/Grammar/AttributesTests.cs
MDoerner: I did not use data-driven tests because the input code would be a giant parameter. It is more than half the size of the test methods. Putting that large parameters into the square brackets above the test would be unwieldy and putting them into a separate test source would move them too far away from the actual test.

Moreover, with data-driven tests, I do not get different names for the tests.

PR #4300, opened 2018-08-24
Massive rework of interfaces and members

comintern:
**Highlights:**

- Introduces a couple new base classes for `Declaration`s
- `ClassModuleDeclaration`s now track their members
- Interface members now resolve their own interface relationships
- Moved extension methods for `Declaration`s dealing with interfaces to the `DeclarationFinder`
- Introduces `VariableDeclaration` to support public fields on interfaces
- Tracked down rogue instances of inspections and refactorings doing custom interface resolution (might have missed some)
- Probably some other stuff I'm forgetting

WIP'ing this one pending rewriting of the tests from my first refactoring go.


Rubberduck.Parsing/Symbols/ClassModuleDeclaration.cs
```diff
@@ -210,6 +210,22 @@ private bool HasPredeclaredIdToCache()
...
+
+        public bool IsUserInterface => Subtypes.Any(s => s.IsUserDefined);
+
+        public IEnumerable<ClassModuleDeclaration> ImplementedInterfaces =>
+            _supertypes.Cast<ClassModuleDeclaration>().Where(type => type.Subtypes.Contains(this));
+
+
+        private readonly List<Declaration> _members = new List<Declaration>();
+        public IReadOnlyList<Declaration> Members => _members;
+
```
MDoerner: I think this would also make sense on a `ProceduralModuleDeclaration`. Maybe we should extract an abstract `ModuleDeclaration` all module declarations derive from.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -77,64 +82,61 @@ protected virtual void ExecuteCollectionConstructionActions(List<Action> collect
...
+                        .Where(declaration => declaration.IsUserDefined)
+                        .GroupBy(declaration => declaration.DeclarationType)
+                        .ToDictionary(),
+                () =>
+                    _referencesByModule = declarations
+                        .SelectMany(declaration => declaration.References)
+                        .GroupBy(reference =>
+                            Declaration.GetModuleParent(reference.ParentScoping).QualifiedName.QualifiedModuleName)
+                        .ToDictionary(),
+                () =>
```
MDoerner: I think the original code looked a bit more structured, but this is OK with me as well.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -158,15 +160,19 @@ private void InitializeLazyCollections()
...
+
+            _implementingMembers = new Lazy<IDictionary<Tuple<VBAParser.ImplementsStmtContext, Declaration>, List<ModuleBodyElementDeclaration>>>(FindAllImplementingMembers, true);
+            _interfaceMembers = new Lazy<IDictionary<ClassModuleDeclaration, List<Declaration>>>(FindAllIinterfaceMembersByModule, true);
+            _membersByImplementsContext = new Lazy<IDictionary<VBAParser.ImplementsStmtContext, List<ModuleBodyElementDeclaration>>>(FindAllImplementingMembersByImplementsContext, true);
+            _interfaceImplementions = new Lazy<IDictionary<ClassModuleDeclaration, List<ClassModuleDeclaration>>>(FindAllImplementionsByInterface, true);
+            _implementationsByMember = new Lazy<IDictionary<ICanBeInterfaceMember, List<ModuleBodyElementDeclaration>>>(FindAllImplementingMembersByMember, true);
         }
 
-        private IDictionary<VBAParser.ImplementsStmtContext, List<Declaration>> FindAllImplementingMembersByImplementsContext()
+        private IDictionary<Tuple<VBAParser.ImplementsStmtContext, Declaration>, List<ModuleBodyElementDeclaration>> FindAllImplementingMembers()
```
MDoerner: Value tuple?

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -176,62 +182,59 @@ private void InitializeLazyCollections()
...
-                    item.Context,
-                    Members = Members(item.IdentifierReference.QualifiedModuleName)
-                        .Where(implementingTypeMember => implementingNames.Contains(implementingTypeMember.IdentifierName))
-                        .ToList()
-                });
+            var output = new Dictionary<Tuple<VBAParser.ImplementsStmtContext, Declaration>, List<ModuleBodyElementDeclaration>>();
+            foreach (var impl in implementsInstructions)
+            {
+                output.Add(
+                    new Tuple<VBAParser.ImplementsStmtContext, Declaration>(impl.Context,
```
MDoerner: That would look a bit tidier using a value tuple.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -176,62 +182,59 @@ private void InitializeLazyCollections()
...
-                        {
-                            IdentifierReference = reference,
-                            Context = reference.Context is VBAParser.ImplementsStmtContext ?
-                                (VBAParser.ImplementsStmtContext)reference.Context
-                                : reference.Context.GetAncestor<VBAParser.ImplementsStmtContext>()
-                        }
-                    )
-                );
+            return UserDeclarations(DeclarationType.ClassModule)
+                .Cast<ClassModuleDeclaration>()
```
MDoerner: Do we want to use `OfType<T>` instead to avoid throwing if something has the declaration type but does not derive from `ClassModuleDeclaration` or do we throw intentionally in that case?

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -176,62 +182,59 @@ private void InitializeLazyCollections()
...
-            var interfaceMembers = interfaceModules.Select(item =>
-                new
-                {
-                    InterfaceModule = item,
-                    InterfaceMembers = Members(item.QualifiedName.QualifiedModuleName)
-                        .Where(member => member.DeclarationType.HasFlag(DeclarationType.Member))
-                });
+        private IDictionary<VBAParser.ImplementsStmtContext, List<ModuleBodyElementDeclaration>> FindAllImplementingMembersByImplementsContext()
+        {
+            return _implementingMembers.Value.ToDictionary(pair => pair.Key.Item1, pair => pair.Value);
```
MDoerner: That would be more descriptive with a value tuple.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -176,62 +182,59 @@ private void InitializeLazyCollections()
...
+        }
 
-            return interfaceMembers.ToDictionary(
-                        item => item.InterfaceModule,
-                        item => item.InterfaceMembers.ToList()
-                     );
+        private IDictionary<ClassModuleDeclaration, List<Declaration>> FindAllIinterfaceMembersByModule()
+        {
+            return UserDeclarations(DeclarationType.ClassModule)
+                .Cast<ClassModuleDeclaration>()
```
MDoerner: See above

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -321,8 +324,7 @@ public IEnumerable<Declaration> Members(Declaration module)
 
         public IEnumerable<Declaration> Members(QualifiedModuleName module)
```
MDoerner: This is not coming from this PR, but this is a somewhat confusing misnomer. I think it should better bu called something like `DeclarationInModule`, which is also not completely accurate since the module declarations itself is part of the returned collection.

Rubberduck.Parsing/Symbols/ClassModuleDeclaration.cs
comintern: I think that's a good idea and it would be a really easy change. It would likely provide opportunities for some other code simplifications similar to this PR in the `DeclarationFinder` (and elsewhere).  

I'll implement that when I fix the tests - might as well do it with the hairy merge...

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
comintern: That was an R# oops - I hit "fix in file" instead of "fix" somewhere along the line.  When I noticed it, I didn't have any strong feelings one way or the other so I just left it.  I can revert changes to that function if you'd prefer.

MDoerner: I guessed it was that. As already stated, I am OK with the new version.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
MDoerner: As long as it is intentional, I am OK with a hard cast.

Rubberduck.Parsing/Symbols/FunctionDeclaration.cs
```diff
@@ -67,23 +61,15 @@ public sealed class FunctionDeclaration : Declaration, IParameterizedDeclaration
...
-        public IEnumerable<ParameterDeclaration> Parameters => _parameters.ToList();
-
-        public void AddParameter(ParameterDeclaration parameter)
+        protected override bool Implements(ICanBeInterfaceMember interfaceMember)
         {
-            _parameters.Add(parameter);
+            return interfaceMember.DeclarationType == DeclarationType.Function
+                && interfaceMember.IsInterfaceMember
+                && IsInterfaceImplementation
+                && IdentifierName.Equals($"{interfaceMember.InterfaceDeclaration.IdentifierName}_{interfaceMember.IdentifierName}");
```
MDoerner: How about extracting the string into a function `InterfaceMemberImplementationName` or something like that to make it a bit more self-documenting?

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                classModule.AddMember(this);
+            }
+        }
+   
+        public bool IsInterfaceImplementation => (_implementsResolved ? _implements : InterfaceImplemented) != null;
+
+        private bool _implementsResolved;
+        private ClassModuleDeclaration _implements;
+        public ClassModuleDeclaration InterfaceImplemented
```
MDoerner: I hope this does not get called concurrently because it is not thread-safe. Maybe use a `Lazy<T>` instead of the flag `_implementsResolved`.

Rubberduck.Parsing/Symbols/FunctionDeclaration.cs
comintern: Works for me. Inbound.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: Didn't even occur to me that this would need to be thread safe. `Lazy<T>` would be a much cleaner implementation anyway. Fix coming.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -45,62 +46,17 @@ public abstract class ModuleBodyElementDeclaration : Declaration, IParameterized
                 annotations,
                 attributes)
         {
-            if (parent is ClassModuleDeclaration classModule)
+            if (!(parent is ModuleDeclaration module))
             {
-                classModule.AddMember(this);
+                return;
```
Vogel612: feels weird to see an early return in a ctor, but it's rather appropriate here

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -117,10 +73,78 @@ protected void AddParameters(IEnumerable<ParameterDeclaration> parameters)
...
+
+        /// <inheritdoc/>
+        public string ImplementingIdentifierName => this.ImplementingIdentifierName();
+
+        /// <summary>
+        /// Returns true if this Declaration is a concrete implementation of the passed member.
+        /// </summary>
+        /// <param name="interfaceMember">The member to test for implementation of.</param>
+        /// <returns>True if this Declaration is a concrete implementation of interfaceMember.</returns>
+        protected abstract bool Implements(ICanBeInterfaceMember interfaceMember);
```
Vogel612: The XML-Doc here makes me wonder whether `ICanBeInterfaceMember` is a misnomer.

Just because an InterfaceMember has not been implemented that doesn't mean it's not an interface member anymore... This distinction only makes sense in the context of the implementing class module.

Could you clarify which context `ICanBeInterfaceMember` should be used in?

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -117,10 +73,78 @@ protected void AddParameters(IEnumerable<ParameterDeclaration> parameters)
...
+            if (!(ParentDeclaration is ClassModuleDeclaration))
+            {
+                return;
+            }
+
+            _interface = new Lazy<(bool IsInterfaceImplementation, ClassModuleDeclaration InterfaceImplemented)>(() => ResolveInterface(this), true);
+            _implemented = new Lazy<Declaration>(() => MemberImplemented(this), true);
+        }
+
+        private static (bool IsInterfaceImplementation, ClassModuleDeclaration InterfaceImplemented) ResolveInterface(ModuleBodyElementDeclaration element)
```
Vogel612: this might make more sense as a Try pattern. `InterfaceImplemented` will always be `null` if the boolean is `false`:

```
private static bool TryResolveInterface(ModuleBodyElementDeclaration member, out ClassModuleDeclaration interface)
```

Makes it a bit clearer that the boolean determines what's in the second tuple part.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -117,10 +73,78 @@ protected void AddParameters(IEnumerable<ParameterDeclaration> parameters)
...
+            }
+
+            var identifiers = element.IdentifierName.Split('_');
+
+            if (identifiers.Length == 1)
+            {
+                return (false, null);
+            }
+
+            var supertype = string.Join("_", identifiers.Take(identifiers.Length - 1));
```
Vogel612: Is this assumption generally true? As in: Is it guaranteed that only the last part of the identifier refers to the member name? If yes, that would definitely warrant a comment for the scrubs like me that don't know their VBA :wink:

Rubberduck.CodeAnalysis/Inspections/Concrete/OptionExplicitInspection.cs
```diff
@@ -25,9 +25,11 @@ public OptionExplicitInspection(RubberduckParserState state)
 
         protected override IEnumerable<IInspectionResult> DoGetInspectionResults()
         {
-            return Listener.Contexts.Select(context => new QualifiedContextInspectionResult(this,
-                                                                      string.Format(InspectionResults.OptionExplicitInspection, context.ModuleName.ComponentName),
-                                                                      context));
+            return Listener.Contexts
+                .Where(context => !IsIgnoringInspectionResultFor(context.ModuleName, context.Context.Start.Line))
```
Vogel612: nice bugfix :+1:

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
```diff
@@ -378,28 +378,90 @@ public IEnumerable<Declaration> DeclarationsWithType(DeclarationType type)
...
 
+        /// <summary>
+        /// Finds the interface declaration for a QualifiedSelection contained by a statement similar to "Implements IClass1"
+        /// </summary>
+        /// <param name="selection">The QualifiedSelection to search.</param>
+        /// <returns>The selected interface if found, null if not found.</returns>
+        public ClassModuleDeclaration FindInterface(QualifiedSelection selection)
+        {
+            return FindAllUserInterfaces().FirstOrDefault(declaration => declaration.References
+                .Where(refrnce => refrnce.Context.GetAncestor<VBAParser.ImplementsStmtContext>() != null)
```
Vogel612: disemvoweled lambda parameter

Vogel612: Also AFAICT the `Where` and the `Any` can be merged into a single `Any`

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
```diff
@@ -0,0 +1,150 @@
...
+            _implemented = new Lazy<Declaration>(() => MemberImplemented(this), true);
+        }
+
+        /// <inheritdoc/>
+        public bool IsDefaultMember => this.IsDefaultMember();
+
+        private readonly List<ParameterDeclaration> _parameters = new List<ParameterDeclaration>();
+        public IEnumerable<ParameterDeclaration> Parameters => _parameters.ToList();
+
+        public void AddParameter(ParameterDeclaration parameter)
```
Vogel612: why is this public, shouldn't it be `internal`? Same justification as for `AddMembers`

Rubberduck.Parsing/Symbols/ModuleDeclaration.cs
```diff
@@ -0,0 +1,51 @@
...
+                null,
+                isUserDefined,
+                annotations,
+                attributes)
+        { }
+
+        private readonly List<Declaration> _members = new List<Declaration>();
+        public IReadOnlyList<Declaration> Members => _members;
+
+        public void AddMember(Declaration member)
```
Vogel612: shouldn't this be `internal`? We don't really want "external" consumers of Rubberduck.Parsing to mess with the state we parsed and resolved, except through rewriting and reparsing

Rubberduck.Refactorings/ImplementInterface/ImplementInterfaceRefactoring.cs
```diff
@@ -125,126 +129,68 @@ private void AddItems(List<Declaration> missingMembers, IModuleRewriter rewriter
...
-
+                    return string.Format(template, "Function", string.Empty, signature, asType);
                 case DeclarationType.PropertyGet:
-                    return PropertyGetStmt(member, interfaceName);
-
+                    return string.Format(template, "Property", " Get", signature, asType);
                 case DeclarationType.PropertyLet:
-                    return PropertyLetStmt(member, interfaceName);
-
+                    return string.Format(template, "Property", " Let", signature, string.Empty);
```
Vogel612: In this whole switch block, we should work with the relevant tokens from the Parser to avoid accidental typos.

Also the template can be changed a bit to have redundant space. double spaces can then be dropped with `string.ReplaceAll("  ", " ");`

Rubberduck.Refactorings/IntroduceParameter/IntroduceParameterRefactoring.cs
```diff
@@ -129,49 +134,37 @@ private bool PromptIfMethodImplementsInterface(Declaration targetVariable)
...
                 functionDeclaration.IdentifierName, interfaceImplementation.ComponentName,
                 interfaceImplementation.IdentifierName);
+
             return _messageBox.Question(message, RubberduckUI.IntroduceParameter_Caption);
         }
 
         private void UpdateSignature(Declaration targetVariable)
         {
-            var functionDeclaration = _declarations.FindTarget(targetVariable.QualifiedSelection, ValidDeclarationTypes);
+            var functionDeclaration = (ModuleBodyElementDeclaration)_declarations.FindTarget(targetVariable.QualifiedSelection, ValidDeclarationTypes);
```
Vogel612: is the hard-cast appropriate here? Maybe comment an explanation why this will always work

Rubberduck.Refactorings/ReorderParameters/ReorderParametersModel.cs
```diff
@@ -44,7 +43,7 @@ private void LoadParameters()
         {
             if (TargetDeclaration == null) { return; }
 
-            Parameters = ((IParameterizedDeclaration) TargetDeclaration).Parameters.Select((s, i) => new Parameter((ParameterDeclaration)s, i)).ToList();
+            Parameters = ((IParameterizedDeclaration) TargetDeclaration).Parameters.Select((param, idx) => new Parameter(param, idx)).ToList();
```
Vogel612: I wonder whether that can be simplified to something like Java's `Parameter::new`

RubberduckTests/Symbols/DeclarationExtensionTests.cs
```diff
@@ -1,145 +1,655 @@
...
+//            var declarations = parser.State.DeclarationFinder.AllDeclarations.ToList();
+//            var member = declarations.Single(decl => decl.IdentifierName.Equals("Foo") && decl.DeclarationType == DeclarationType.Variable);
+//            var setter = declarations.Single(decl => decl.IdentifierName.Equals("TestInterface_Foo") && decl.DeclarationType == DeclarationType.PropertySet);
+//            var letter = declarations.Single(decl => decl.IdentifierName.Equals("TestInterface_Foo") && decl.DeclarationType == DeclarationType.PropertyLet);
+
+//            Assert.IsTrue(setter.ImplementsInterfaceMember(member));
+//            Assert.IsTrue(letter.ImplementsInterfaceMember(member));
+//        }
+//    }
+//}
```
Vogel612: you are fixing those, right? :smile:

RubberduckTests/Symbols/ModuleBodyElementDeclarationTests.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace RubberduckTests.Symbols
+{
+    [TestFixture]
+    public class ModuleBodyElementDeclarationTests
+    {
+        [Test]
+        [Category("Resolver")]
+        public void FindFindInterfaceMemberMatchesDeclarationTypes()
+        {
+            var intrface =
```
Vogel612: trying to use keywords as variable names is really a bit annoying, innit?

how about `interfaceModule`?

Rubberduck.Parsing/Symbols/PropertySetDeclaration.cs
```diff
@@ -61,41 +55,38 @@ public sealed class PropertySetDeclaration : Declaration, IParameterizedDeclarat
...
-            false,
-            null,
-            attributes)
+        public PropertySetDeclaration(ComField field, Declaration parent, QualifiedModuleName module, Attributes attributes) 
+            : this(
+                module.QualifyMemberName(field.Name),
+                parent,
+                parent,
+                field.ValueType,
+                Accessibility.Global,
```
ThunderFrame: There is no `Accessibility.Global`, only `Accessibility.Public`?

Rubberduck.Parsing/Symbols/VariableDeclaration.cs
```diff
@@ -0,0 +1,63 @@
...
+                context,
+                attributesPassContext,
+                selection,
+                isArray,
+                asTypeContext,
+                true,
+                annotations,
+                attributes)
+        {
+            if ((accessibility == Accessibility.Public || accessibility == Accessibility.Implicit)
```
ThunderFrame: `Implcit` isn't an accessibility type - whether a declaration's accessibility is implicitly defined should be a property of the declaration, not forced as a psudo-type

Rubberduck.Refactorings/ImplementInterface/ImplementInterfaceRefactoring.cs
```diff
@@ -125,126 +129,68 @@ private void AddItems(List<Declaration> missingMembers, IModuleRewriter rewriter
...
-                                .OrderBy(d => d.Selection.StartLine)
-                                .ThenBy(d => d.Selection.StartColumn);
-        }
+            if (member.DeclarationType == DeclarationType.Variable)
+            {
+                return new List<Parameter>
+                {
+                    new Parameter
+                    {
+                        Accessibility = Tokens.ByVal,
```
ThunderFrame: `Accessibility` here, is conflating a declaration's accessibility with whether the parameter is passed as a reference or as a copy. Accessibility can only ever be `Public`, `Private` or `Friend`. It can't be `ByVal` or `ByRef`, and it can't be `Implicit` or `Global` either

Rubberduck.Parsing/Symbols/ProceduralModuleDeclaration.cs
```diff
@@ -18,21 +18,12 @@ public sealed class ProceduralModuleDeclaration : Declaration
...
-                  Selection.Home,
-                  false,
-                  null,
                   isUserDefined,
                   annotations,
-                  attributes) { }
+                  attributes)
+        { }
 
         public ProceduralModuleDeclaration(ComModule statics, Declaration parent, QualifiedModuleName module,
```
ThunderFrame: Is `statics` the best name for this parameter, and why does it differ so much from the constructors for `ClassModuleDeclaration`?

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: I borrowed the naming convention for `ICanBeInterfaceMember` from `ICanBeDefaultMember`, but I can't claim to be thrilled with either of them.  It's basically anything that can be exposed on a VB interface. The `ICanBeInterfaceMember` exists because public variables in interface classes are members, so `VariableDeclaration` and `ModuleBodyElementDeclaration` need the functionality that it provided.  I'll put XML-Doc on the interface itself, but any naming suggestions would be welcome.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: That came up in chat the other day. I had it wrong on *my* first go because I originally assumed that VBA restricted you to a single underscore.  I'll see if I can track down a reference in the language spec to comment it with.

Rubberduck.Parsing/Symbols/DeclarationFinder.cs
comintern: Fixed.  My indentation sucked there too.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: Found it that way.  It's also member of `IParameterizedDeclaration`, so it currently has to be public because that interface is used outside of `Rubberduck.Parsing`.

Rubberduck.Parsing/Symbols/ModuleDeclaration.cs
comintern: That one's on me.  Fixed.

Rubberduck.Refactorings/ImplementInterface/ImplementInterfaceRefactoring.cs
comintern: Agreed and fixed.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
Vogel612: how about `IInterfaceExposable`?

Rubberduck.Refactorings/IntroduceParameter/IntroduceParameterRefactoring.cs
comintern: All the members of `ValidDeclarationTypes` inherit from `ModuleBodyElementDeclaration`, so the cast will always succeed.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: I like that. Done.

Rubberduck.Parsing/Symbols/ModuleBodyElementDeclaration.cs
comintern: Not sure I can do that with a `Lazy<T>` initializer without putting a wrapper function around it, which would probably make it less clear what's going on.

RubberduckTests/Symbols/DeclarationExtensionTests.cs
comintern: Yep - I'm refactoring them to use the `DeclarationFinder` instead. That's coming later today.

RubberduckTests/Symbols/ModuleBodyElementDeclarationTests.cs
comintern: There might be some more of those. I'll try to change them as I find them.

Rubberduck.Parsing/Symbols/PropertySetDeclaration.cs
comintern: This came up several weeks earlier when we were discussing whether the scoping assigned to the built in declarations was appropriate. AFAIK these have always been hard coded to global accessibility since we started doing COM reflection, so I just preserved the original behavior when I did the refactor.  Tracking down the assumptions that are being made around this in the resolver and elsewhere is probably a major undertaking in and of itself - my fear is that changing this might cause as many regressions as it fixes things. This may need to be opened as a new issue.

Rubberduck.Parsing/Symbols/VariableDeclaration.cs
comintern: More tech debt. I believe that this is being driven by the resolver, but I agree that at some point we need to remove `Accessibility.Implicit` entirely.

Rubberduck.Refactorings/ImplementInterface/ImplementInterfaceRefactoring.cs
comintern: See above.

Rubberduck.Parsing/Symbols/ProceduralModuleDeclaration.cs
comintern: IIR I picked the name `statics` because a COM module that does not define an interface or a class (`TKIND_MODULE`) is analogous to a `static class` in .NET, and `module` gets confusing due to how it is (over) used in VB parlance.

The ctor is a stripped down version of `ClassModuleDeclaration` because it doesn't carry some of the flags that are only relevant to coclasses and interfaces (e.g. a standard module can't have a default member, can't be a control, can't predeclared, etc.).

comintern:
Tests are back in, fixed an issue with circular evaluation of Lazy<T>s. No longer WIP (assuming AV passes it).

RubberduckTests/Symbols/DeclarationFinderTests.cs
```diff
@@ -1303,6 +1304,667 @@ End Sub
...
+            {
+                var declarations = state.DeclarationFinder.AllDeclarations.ToList();
+                var declaration = declarations.Single(decl => decl.DeclarationType == DeclarationType.PropertyGet && decl.IdentifierName.Equals("Foo"));
+
+                var expected = declarations.Single(decl => decl.DeclarationType == DeclarationType.PropertyGet && decl.IdentifierName.Equals("TestInterface_Foo"));
+                var actual = state.DeclarationFinder.FindInterfaceImplementationMembers(declaration).ToList();
+                var results = actual.Count;
+
+                Assert.AreEqual(1, results, "Expected {0} Declarations, received {1}", expected, results);
+                Assert.AreEqual(expected, actual.First(), "Expected {0}, resolved to {1}", expected, actual);
```
bclothier: Shouldn't we also assert the `Let`, too?

RubberduckTests/Symbols/DeclarationFinderTests.cs
```diff
@@ -1303,6 +1304,667 @@ End Sub
...
+                .AddComponent("TestImplementation", ComponentType.ClassModule, implementation, new Selection(1, 1))
+                .AddProjectToVbeBuilder()
+                .Build();
+
+            using (var state = MockParser.CreateAndParse(vbe.Object))
+            {
+                var actual = state.DeclarationFinder.FindAllInterfaceMembers().Count();
+                const int expected = 3;
+
+                Assert.AreEqual(expected, actual, "Expected {0} Declarations, received {1}", expected, actual);
```
bclothier: The test says matches public variable but we're just counting 3 declarations without verifying they're actually the same data type? Shouldn't we assert that using `Assert.That  .... Is.Equivalent`?

RubberduckTests/Symbols/DeclarationFinderTests.cs
```diff
@@ -1303,6 +1304,667 @@ End Sub
...
+                var letter = declarations.Single(decl => decl.IdentifierName.Equals("TestInterface_Foo") && decl.DeclarationType == DeclarationType.PropertyLet);
+
+                var actualSetter = (setter as ModuleBodyElementDeclaration)?.InterfaceMemberImplemented;
+                var actualLetter = (letter as ModuleBodyElementDeclaration)?.InterfaceMemberImplemented;
+
+                Assert.AreEqual(expected, actualSetter, "Expected {0}, resolved to {1}", expected, actualSetter);
+                Assert.AreEqual(expected, actualLetter, "Expected {0}, resolved to {1}", expected, actualLetter);
+            }
+        }
+
```
bclothier: Don't you think we need an unit test to verify that we can still match the property even if the name is different? All unit tests use `rhs` or `NewValue`; will it still find if the definition uses `NewValue` while the implementation uses `rhs` instead?

RubberduckTests/Symbols/DeclarationFinderTests.cs
comintern: That's implicit in the first `Assert`, isn't it?

RubberduckTests/Symbols/DeclarationFinderTests.cs
comintern: I suppose I *could*, but the count tests that all of the declared members are interface members. Maybe it just needs a better name?

RubberduckTests/Symbols/DeclarationFinderTests.cs
comintern: Yep, I didn't think to add that because it doesn't even look at the parameters (if there's a type or signature mismatch that's a compile error IIR).  I can add it a test though to make sure that the naming doesn't matter.

RubberduckTests/Symbols/DeclarationFinderTests.cs
bclothier: Unless I misunderstand the test filters out the `Let` and only looks at the `Get` even though the input code has both `Get` and `Let`.

comintern: Right, but if there is only 1 result it has to be either a `Get` ***or*** a `Let`, and `expected` is a `PropertyGet`.

PR #4303, opened 2018-08-24
Grammar tweak for indexed properties in form header

### LINKED ISSUE #4302: Indexed properties in form headers cause parse error
Properties in form headers can be indexed:

      BeginProperty ColumnHeader(1) {0713E8C7-850A-101B-AFC0-4210102A8DA7} 
         Key             =   ""
         Object.Tag             =   ""
         Text            =   "Source"
         Object.Width           =   2540
      EndProperty

causes `MainParseSyntaxErrorException: mismatched input '('`

mansellan:
Fixes #4302

Also added missing using to fix `next`

### DIFF
--- a/Rubberduck.Parsing/Grammar/VBAParser.g4
+++ b/Rubberduck.Parsing/Grammar/VBAParser.g4
@@ -56,7 +56,7 @@ moduleConfig :
 ;
 
 moduleConfigProperty :
-    BEGINPROPERTY whiteSpace unrestrictedIdentifier (whiteSpace GUIDLITERAL)? endOfStatement
+    BEGINPROPERTY whiteSpace unrestrictedIdentifier (LPAREN numberLiteral RPAREN)? (whiteSpace GUIDLITERAL)? endOfStatement
         (moduleConfigProperty | moduleConfigElement)+
     ENDPROPERTY endOfStatement
 ;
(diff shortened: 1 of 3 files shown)


Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -56,7 +56,7 @@ moduleConfig :
 ;
 
 moduleConfigProperty :
-    BEGINPROPERTY whiteSpace unrestrictedIdentifier (whiteSpace GUIDLITERAL)? endOfStatement
+    BEGINPROPERTY whiteSpace unrestrictedIdentifier (LPAREN numberLiteral RPAREN)? (whiteSpace GUIDLITERAL)? endOfStatement
```
bclothier: Just making sure --- is it legal to have whitespace between the identifier and the `(`? It might get pretty-printed out but the grammar tries to work even with non-pretty-printed version.

mansellan: Whitespace causes VB6 to warn on open and emit to a log:

    Line 43: The CLSID (1) for ColumnHeader is invalid.
    Line 44: Syntax error: property {0713E8C7-850A-101B-AFC0-4210102A8DA7} in ColumnHeader was missing an '='.

It repairs it on save. Should I allow in the grammar?

bclothier: I wouldn't. From the warning message, the grammar has changed, so it's fine as it is.

PR #4304, opened 2018-08-24
Duplicated Annotation Inspection

rkapka:
This PR introduces an inspection that finds duplicated annotation declarations and proposes a quick fix. The fix is not very smart, since it can sometimes leave empty single quotes (see tests).

While working on this, I have made some not-so-small changes to a couple of things and I think I found one parser issue.

### Changes

- I added a `Context` property to `IAnnotation` in order to apply the quick fix. I could have designed the inspection as a parse tree inspection, gather all the annotation contexts and pass them along with the declaration in the dynamic property bag, but the context is available in the annotation factory so I used it.
- I removed the check for duplicate annotations from `IllegalAnnotationInspection` because it is handled in the new inspection. What's more, in my opinion the other inspection did not work correctly because it disallowed any multiple module annotations, whereas `IgnoreModuleAnnotation` can be applied multiple times.

### Parser issue

Some quick fix tests are ignored because I encountered an unexpected parse error. I was trying to put multiple annotations on the same line which should be possible according to the following part of the grammar:

```
// Annotations must come before comments because of precedence. ANTLR4 matches as much as possible then chooses the one that comes first.
commentOrAnnotation :
    (annotationList 
    | remComment
    | comment) 
    // all comments must end with a logical line. See VBA Language Spec 3.3.1
    (NEWLINE | EOF)
;
(...)
annotationList : SINGLEQUOTE (AT annotation whiteSpace?)+ (COLON commentBody)?;
```

For some reason `'@Obsolete @NoIndent` fails to parse. Writing such code in the VBE results in a resolver error.


Vogel612:
Could you try `'@Obsolete() @NoIndent()` for me, please?

MDoerner:
The problem with multiple annotations in one line is that whoever adapted the grammar to allow this did not adapt `DeclarationSymbolsListener.FindAnnotations(int line)`. The fix would be to remove the `SingleOrDefault`, make the condition an `Any` check and then use `AddRange` instead of `Add`.

rkapka:
@Vogel612 It didn't help.

I don't feel good about the test fix commit. Originally I assumed that annotations are added to a declaration in the order in which they appear in VBA code. This assumption seemed to be correct since all tests were passing. After merging with the `next` branch, tests started failing and assuming a reverse (!) order made them pass again. I am confused as to what is happening here...

rkapka:
All right, now I kind of see what's going on. `DeclarationSymbolListener.FindAnnotations(int line)` searches for annotations in reverse order of lines. No one changed this lately, so I can't see why the first version of the quick fix passed all tests.

RubberduckTests/Inspections/DuplicatedAnnotationInspectionTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                var inspection = new DuplicatedAnnotationInspection(state);
+                var inspectionResults = inspection.GetInspectionResults(CancellationToken.None);
+
+                Assert.AreEqual(0, inspectionResults.Count());
+            }
+        }
+
+        [Test]
+        [Category("Inspections")]
+        public void MemberAnnotationDuplicatedForModuleDeclaration_DoesNotReturnResult()
```
Vogel612: I don't think we should check that. This intermingles the responsibility of determining legality (member annotations on members, module annotations on modules) with the responsibility of determining duplication.

These are separate things. I'd expect the code tested here to spawn the following inspection results:

 - IllegalAnnotation: member Annotation "TestInitialize" used at module level (2)
 - DuplicateAnnotation: Annotation "TestInitialize" is not intended to be specified mutliple times

That'd simplify the condition mentioned above

RubberduckTests/Inspections/DuplicatedAnnotationInspectionTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                var inspection = new DuplicatedAnnotationInspection(state);
+                var inspectionResults = inspection.GetInspectionResults(CancellationToken.None);
+
+                Assert.AreEqual(0, inspectionResults.Count());
+            }
+        }
+
+        [Test]
+        [Category("Inspections")]
+        public void ModuleAnnotationDuplicatedForMemberDeclaration_DoesNotReturnResult()
```
Vogel612: same thing as above, only in "reverse" :)

RubberduckTests/Inspections/IllegalAnnotationsInspectionTests.cs
```diff
@@ -164,51 +164,6 @@ End Sub
...
-            using (var state = MockParser.CreateAndParse(vbe.Object))
-            {
-
-                var inspection = new IllegalAnnotationInspection(state);
-                var inspector = InspectionsHelper.GetInspector(inspection);
-                var inspectionResults = inspector.FindIssuesAsync(state, CancellationToken.None).Result;
-
-                Assert.IsFalse(inspectionResults.Any());
-            }
-        }
```
Vogel612: not sure why this test was deleted

RubberduckTests/QuickFixes/RemoveDuplicatedAnnotationQuickFixTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        public void RemoveDuplicatedAnnotation_QuickFixWorks_RemoveDuplicate()
+        {
+            const string inputCode = @"
+'@Obsolete
+'@Obsolete
+Public Sub Foo
+End Sub";
+
+            const string expectedCode = @"
+'
```
Vogel612: it'd be really cool if the quickfix could remove the leading `'` if all annotations in a line are removed

RubberduckTests/QuickFixes/RemoveDuplicatedAnnotationQuickFixTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        public void RemoveDuplicatedAnnotation_QuickFixWorks_RemoveDuplicateFromOtherAnnotationList()
+        {
+            const string inputCode = @"
+'@Obsolete @NoIndent
+'@Obsolete
+Public Sub Foo
+End Sub";
+
+            const string expectedCode = @"
+' @NoIndent
```
Vogel612: hmm .. I'd want to remove this whitespace. Not sure `'@NoIndent` is recognized like that, considering that the grammar rule starts `annotationList : SINGLEQUOTE (AT annotation whiteSpace?)+[..]`. Note the lack of permitted whitespace after the `SINGLEQUOTE`

Vogel612: I think this can be fixed by adding `rewriter.Remove(annotationList.whiteSpace(index));`, but I'm not definitively sure

RubberduckTests/Inspections/IllegalAnnotationsInspectionTests.cs
rkapka: This test was checking whether an annotation is not duplicated, which is now the responsibility of the new inspection.

Rubberduck.CodeAnalysis/Inspections/Concrete/DuplicatedAnnotationInspection.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            foreach (var declaration in State.AllUserDeclarations)
+            {
+                var duplicateAnnotations = declaration.Annotations
+                    .GroupBy(annotation => annotation.AnnotationType)
+                    .Where(group => !group.First().AllowMultiple && group.Count() > 1 &&
+                                    (group.Key.HasFlag(AnnotationType.ModuleAnnotation) &&
+                                     declaration.DeclarationType.HasFlag(DeclarationType.Module) ||
+                                     group.Key.HasFlag(AnnotationType.MemberAnnotation) &&
+                                     declaration.DeclarationType.HasFlag(DeclarationType.Member)));
```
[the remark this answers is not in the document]
rkapka: The reason I did this was because a duplicate annotation result would show up for illegal annotations. My reasoning was that if an annotation is illegal, then no inspection results should be shown for such annotations, other than it being illegal.
I understand that this will cause duplicate checks in different inspections, like this one. I was thinking that it would be nice to have a `IsLegal` property on the annotation class and display annotation-specific inspection results only for legal annotations. I'm not sure if that's possible, though.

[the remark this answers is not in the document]
rkapka: Anyway, if it's OK to show inspection results for illegal annotations, then I'm fine with removing this check and associated tests.

RubberduckTests/QuickFixes/RemoveDuplicatedAnnotationQuickFixTests.cs
rkapka: I will try to correct this.

RubberduckTests/QuickFixes/RemoveDuplicatedAnnotationQuickFixTests.cs
rkapka: I will try to correct this.

rkapka:
I removed the checks from inspection code as suggested by @Vogel612, some tests started failing and this reminded me that there was a second reason for this check, other than the one in my review comment. Whenever a duplicate annotation is placed on the first member of a module, two inspection results will trigger - one for the member and one for the module (is this a bug?). I think `IllegalAnnotationInspection` deals with it in some way using the `_isFirstMemberProcessed` field, but I decided these additional check would suffice for my case (by the way, I just noticed that another check is required for the case when an annotation does not have `MemberAnnotation` nor `ModuleAnnotation` flag; currently no inspection results are shown; note again that fixing this will result in two inspections when a module's first member is annotated).

rkapka:
Wait a second... I don't understand anything any more...

Removing the mentioned checks makes this test fail, because there are 2 results:
```
public void AnnotationDuplicated_ReturnsResult()
{
    const string inputCode = @"
'@Obsolete
'@Obsolete
Public Sub Foo()
End Sub";

    var vbe = MockVbeBuilder.BuildFromSingleStandardModule(inputCode, out _);
    {
        var inspection = new DuplicatedAnnotationInspection(state);
        var inspectionResults = inspection.GetInspectionResults(CancellationToken.None);

        Assert.AreEqual(1, inspectionResults.Count());
    }
}
```

What's surprising, the Code Inspections window shows the correct result:
[
2018-09-15 16_13_43-window
]()

MDoerner:
I think the problem with the tests might be that the mock implementation for the member of `ICodeModule` finding the end of the module declarations section does not do the same as the one in the VBE API.

We really should replace that VBE API call in the declaration resolver by something based on the parse tree. (That might not be in scope for this PR, though.) 

Could you add another procedure above the annotated one in the test and see whether that fixes the issue? 

If that works, the test recognizes the annotation both as a module and member annotation, which are both duplicated, the module one being a bug in the setup of the mocks.

rkapka:
@MDoerner That's what I did initially and it worked, but I thought at first that the same member/module result issue will occur in VBE so I added the check instead. Thank you for clarifying.

rkapka:
Apart from review comments, the last commit has a couple of fixes/enhancements. There is one quick fix scenario that will break existing annotations - when all that's left from an annotation list is a comment and another list follows the comment.

```
'@Obsolete
'@Obsolete : Comment
'@TestMethod
```
will result in
```
'@Obsolete
': Comment
'@TestMethod
```

Vogel612:
Ideally the comments on multiple annotations of the same type would be merged. We want to avoid removing too much when we apply that quick fix.

As such the current result is exactly what we want (at least for now)

PR #4324, opened 2018-08-30
Fix massive parser slowdown 

### LINKED ISSUE #4307: Massive parser slowdown
On the `next` from last night, I'm observing that the parser seems to be much much slower to parse VBA projects of moderate complexity in both Access and Excel. Depending on which I open, it can take from 20 seconds to > 2 minutes to parse projects that used to parse in 2-5 seconds. Is anyone else observing the same?

MDoerner:
Closes #4307 

This PR corrects my mistake in the parser refactoring PR to implement the parser fallback mode in the wrong order making it always use LL mode. 

Note that this does not undo the whole slow down: a doubling of the parsing time is legitimate. This is caused by a fix of a previous registration bug that basically disabled the attributes parse pass. 

In addition to the fix, this PR introduces some debug level logging of the time the main parsing steps take. These do not add up to the whole parse time since that also includes the startup work, cache invalidation, generation of the 'DeclarationFinder` and running the status changed event handlers.


bclothier:
LGTM; the only thing to consider is the noise of the log which may make finding those times a bit harder. If they all had the same distinct prefix (e.g. `ParserTimer: ...`) that would aid in locating all the times even in a verbose log.

MDoerner:
@bclothier It should be possible now to search the log for `ParsingStageTimer` in order to find the timer log entries.

PR #4343, opened 2018-08-31
Missing commit from #4314 

comintern:
Ignore implicit sets in For Each statements. Addresses #4294 Ref #4318

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableRequiresSetAssignmentEvaluator.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableRequiresSetAssignmentEvaluator.cs
@@ -42,7 +42,8 @@ namespace Rubberduck.Inspections
             }
 
             // For Each iterators are implicitly set.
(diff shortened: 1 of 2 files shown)


Rubberduck.CodeAnalysis/Inspections/Concrete/VariableRequiresSetAssignmentEvaluator.cs
```diff
@@ -41,6 +41,12 @@ public static bool RequiresSetAssignment(IdentifierReference reference, Rubberdu
                 return false;
             }
 
+            // For Each iterators are implicitly set.
+            if (reference.Context.GetAncestor<VBAParser.ForEachStmtContext>() != null)
```
Vogel612: This will ignore all assignments within the block of a foreach Statement... I'm pretty sure that's not what was intended...

comintern: You're correct - that was my bias in assuming that nobody would assign to the iterator inside the loop.

Vogel612:
Could We also get a unit-test to prevent a repeated regression?

PR #4354, opened 2018-09-06
Chained COM Wrapper Analyzer

retailcoder:
Commits from #4238


retailcoder:
ugh didn't checkout off the right branch....

PR #4403, opened 2018-10-03
Fix default members and parameters when unpacking. Closes #4390

### LINKED ISSUE #4390: Serialized declarations seem to be missing default member attributes.
Noticed this when I was writing the (missing) tests for #4327. Given a test setup that uses the Excel reference, this code...

```
            var interestingDeclarations =
                State.AllDeclarations.Where(item =>
                    item.AsTypeDeclaration != null
                    && ClassModuleDeclaration.HasDefaultMember(item.AsTypeDeclaration));
```
...returns no results.  The declarations collected from the typelib are fine.

comintern:


comintern:
WIPing this. There are other deserialization issues to address.

PR #4409, opened 2018-10-04
Fix unit tests grouping by outcome

### LINKED ISSUE #4386: Unit tests grouping by outcome doesn't group by outcome
As title indicates.
untitled

The ability to navigate to tests by double clicking on the result has also stopped.

IvenBach:
Close #4386. Adds implicit categorization for tests.

### DIFF
--- a/Rubberduck.Core/UI/UnitTesting/TestResultToOutcomeTextConverter.cs
+++ b/Rubberduck.Core/UI/UnitTesting/TestResultToOutcomeTextConverter.cs
@@ -10,9 +10,9 @@ namespace Rubberduck.UI.UnitTesting
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if (value is TestResult result)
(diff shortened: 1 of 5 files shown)


rkapka:
I am accepting this PR, but there is one thing that I would like to check. Before your changes sorting tests by category put the no-category group as last. I believe your change breaks this sorting - `Uncategorized` will be treated as any user-created category name.

IvenBach:
@rkapka Thanks for bringing the sorting order to my attention. I just tested the sorting and the uncategorized tests remained at the top with one exception. Adding an explicit Uncategorized category would group together both the implicit and explicit Uncategorized methods together.

In testing I discovered sorting is case-sensitive. Should we worry about this?

rkapka:
Whatever name we choose for uncategorised tests there is always a change someone will create such a category. I don't think that's an issue though - I believe the user will figure out what's going on and change their naming.

As for case-sensitivity, it would be better to have insensitive ordering.

PR #4416, opened 2018-10-08
Add Flags annotation

### LINKED ISSUE #3375: Bitwise annotation for Enums, and Flags property for declarations
RD could expose several inspections for bitwise enums, if it only knew that the enums should be regarded as being bitwise.

We *could* make a best-efforts guess about whether a user-defined enum is bitwise, but that could yield false-positives with an enum like this:

```vb
Enum Foo
  Am
  I
  Bitwise
End Enum
```

If we could add an annotation (and I suggest we use the .NET-inspired `@Flags` annotation, then RD would know the intent of the Enum:

```vb
'@Flags
Enum Foo
  Must
  Be
  Bitwise
End Enum
```

Then RD could inspect for enums that implicitly/explicitly assign values that aren't bitwise, although there might need to be some consideration for enum members that are assigned using bitwise expressions (for example, where a user declares a bitwise combination member, for convenience):

```vb
Enum UserResponse
  'Bitwise members
  Cancel = 1
  Ok = 2
  Abort = 4
  Retry = 8
  'Convenience members
  OkOrCancel = OK Or Cancel
  AbortOrRetry = Abort Or Retry
End Enum
```
When collecting COM enums defined in type-libraries, we could manually define the bitwise nature of known bitwise enums:

Eg: **VbFileAttribute** is a pure bitwise enum
```vb
Enum VbFileAttribute
  vbNormal = 0
  vbReadOnly = 1
  vbHidden = 2
  vbSystem = 4
  vbVolume = 8
  vbDirectory =16
  vbAlias = 64
  vbArchive =32
End Enum
```

Whereas **VbStrConv** seems like it might be bitwise (VB.NET documentation marks it with a Flags attribute), but with an explicitly non-bitwise member:

```vb
'declared in alphabetical order - the same as the VB.Net declaration order...
Enum VbStrConv
  Hiragana = 32;
  Katakana = 16;
  LinguisticCasing = 1024;
  Lowercase = 2;
  Narrow = 8;
  None = 0;
  ProperCase = 3; 'Explicitly non-bitwise, but happens to be an invalid combination of Uppercase Or Lowercase
  SimplifiedChinese = 256;
  TraditionalChinese = 512;
  Uppercase = 1;
  Wide = 4;
End Enum
```

Hosch250:
Reference #3375

### DIFF
--- a/Rubberduck.Parsing/Annotations/AnnotationType.cs
+++ b/Rubberduck.Parsing/Annotations/AnnotationType.cs
@@ -33,6 +33,8 @@ namespace Rubberduck.Parsing.Annotations
         Folder = 1 << 17 | ModuleAnnotation,
         NoIndent = 1 << 18 | ModuleAnnotation,
         Interface = 1 << 19 | ModuleAnnotation,
+        Flags = 1 << 20 | MemberAnnotation,
+
         [AttributeAnnotation("VB_Description")]
         Description = 1 << 13 | Attribute | MemberAnnotation,
         [AttributeAnnotation("VB_UserMemId", "0")]
--- a/Rubberduck.Parsing/Annotations/FlagsAnnotation.cs
+++ b/Rubberduck.Parsing/Annotations/FlagsAnnotation.cs
@@ -1,0 +1,18 @@
+﻿using Rubberduck.VBEditor;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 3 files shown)


Vogel612:
I think it's premature to close the linked issue with this pull-request. We're not using the flags annotation for much of what it could achieve yet.

PR #4426, opened 2018-10-11
Introduce Serialize method for COM Safe

bclothier:
This PR is primarily for developers' convenience and was made necessary because there can be a large number of objects being stored in the COM safe, which can make it difficult to track object's lifetime using only Watch or Immediate Window.

This provides a debug-only serialize which generates a tab-delimited file that can be then opened in Excel:

[!image](https://user-images.githubusercontent.com/2367644/46806239-b8f32e00-cd2c-11e8-98cf-8633be85065e.png)

At this point, there is no UI for this functionality. I've been considering overloading the existing `Serialize` button for this. Any ideas welcome.

For now, I wanted to put this out for reviews in case others might have some useful suggestions or something else.

Todo:

1) Work out the UI, if any at al
2) Implement for the `StrongComSafe`

### DIFF
--- a/Rubberduck.VBEEditor/ComManagement/WeakComSafe.cs
+++ b/Rubberduck.VBEEditor/ComManagement/WeakComSafe.cs
@@ -3,13 +3,17 @@ using System.Collections.Concurrent;
 using System.Runtime.CompilerServices;
 using Rubberduck.VBEditor.SafeComWrappers.Abstract;
 
+#if DEBUG
+using System.Linq;
+using System.Runtime.InteropServices;
+#endif
+
 namespace Rubberduck.VBEditor.ComManagement
 {
     public class WeakComSafe : IComSafe
     {
         //We use weak references to allow the GC to reclaim RCWs earlier if possible.
-        private readonly ConcurrentDictionary<int, WeakReference<ISafeComWrapper>> _comWrapperCache = new ConcurrentDictionary<int, WeakReference<ISafeComWrapper>>();
-
+        private readonly ConcurrentDictionary<int, (DateTime insertTime, WeakReference<ISafeComWrapper> weakRef)> _comWrapperCache = new ConcurrentDictionary<int, (DateTime, WeakReference<ISafeComWrapper>)>();
 
         public void Add(ISafeComWrapper comWrapper)
         {
@@ -17,8 +21,8 @@ namespace Rubberduck.VBEditor.ComManagement
             {
                 _comWrapperCache.AddOrUpdate(
                     GetComWrapperObjectHashCode(comWrapper), 
-                    key => new WeakReference<ISafeComWrapper>(comWrapper), 
-                    (key, value) => new WeakReference<ISafeComWrapper>(comWrapper));
+                    key => (DateTime.UtcNow, new WeakReference<ISafeComWrapper>(comWrapper)), 
+                    (key, value) => (DateTime.UtcNow, new WeakReference<ISafeComWrapper>(comWrapper)));
             }
 
         }
@@ -47,7 +51,7 @@ namespace Rubberduck.VBEditor.ComManagement
 
             foreach (var weakReference in _comWrapperCache.Values)
             {
-                if(weakReference.TryGetTarget(out var comWrapper))
+                if(weakReference.weakRef.TryGetTarget(out var comWrapper))
                 {
                     comWrapper.Dispose();
                 }
@@ -55,5 +59,57 @@ namespace Rubberduck.VBEditor.ComManagement
 
             _comWrapperCache.Clear();
         }
+
+#if DEBUG
+        /// <summary>
+        /// Provide a serialized list of the COM Safe
+        /// to make it easy to analyze what is inside
+        /// the COM Safe at the different points of
+        /// the session's lifetime.
+        /// </summary>
+        public void Serialize()
+        {
+            using (var stream = System.IO.File.AppendText($"comSafeOutput {DateTime.UtcNow:yyyyMMddhhmmss}.csv"))
+            {
+                stream.WriteLine("Ordinal\tKey\tCOM Wrapper Type\tWrapping Null?\tIUnknown Pointer Address");
+                var i = 0;
+                foreach (var kvp in _comWrapperCache.OrderBy(kvp => kvp.Value.insertTime))
+                {
+                    var line = kvp.Value.weakRef.TryGetTarget(out var target) 
+                        ? $"{i++}\t{kvp.Key}\t\"{target.GetType().FullName}\"\t\"{target.IsWrappingNullReference}\"\t\"{(target.IsWrappingNullReference ? "null" : GetPtrAddress(target.Target))}\"" 
+                        : $"{i++}\t{kvp.Key}\t\"null\"\t\"null\"\t\"null\"";
+                    stream.WriteLine(line);
+                }
+            }
+        }
+
+        private static string GetPtrAddress(object target)
+        {
+            if (target == null)
+            {
+                return IntPtr.Zero.ToString();
+            }
+
+            if (!Marshal.IsComObject(target))
+            {
+                return "Not a COM object";
+            }
+
+            var pointer = IntPtr.Zero;
+            try
+            {
+                pointer = Marshal.GetIUnknownForObject(target);
+            }
+            finally
+            {
+                if (pointer != IntPtr.Zero)
+                {
+                    Marshal.Release(pointer);
+                }
+            }
+
+            return pointer.ToString();
+        }
+#endif
     }
 }


Rubberduck.VBEEditor/ComManagement/WeakComSafe.cs
```diff
@@ -3,22 +3,26 @@
...
         public void Add(ISafeComWrapper comWrapper)
         {
             if (comWrapper != null)
             {
                 _comWrapperCache.AddOrUpdate(
                     GetComWrapperObjectHashCode(comWrapper), 
-                    key => new WeakReference<ISafeComWrapper>(comWrapper), 
-                    (key, value) => new WeakReference<ISafeComWrapper>(comWrapper));
+                    key => (DateTime.UtcNow, new WeakReference<ISafeComWrapper>(comWrapper)), 
+                    (key, value) => (DateTime.UtcNow, new WeakReference<ISafeComWrapper>(comWrapper)));
```
bclothier: I was a bit too hasty and didn't read properly - the updateFactory parameter should be doing `value.insertTime`, not `DateTime.UtcNow` in order to avoid changing the insertion time. But I am wondering why is the `WeakReference` new'd up here without any checks to the previous reference. Should we be disposing the old `WeakReference` or something?

MDoerner: You really should not dispose the COM wrapper to which the weak reference points. With all likelihood it is the wrapper passed to you. (They have the same object hash.) In the very unlikely case that there you got a new wrapper with exactly the same hash as one allready in the dictionary, I prefer to safe the new one.

bclothier: That makes sense but what about the case where the original `WeakReference` is now wrapping a null reference? My worry is that because it points to a SCW and if the SCW in turn wraps a null reference, the `WeakReference` would still think it's alive even though it's just an empty shell at this point. In that case, we need to dispose... right?

bclothier: Just to summarize the discussion - `Debug.Assert` will be added to the `updateValueFactory` primarily to verify whether this get called as under normal circumstances, all operations should be additions or deletions, not updates. There might be rare update due to hash collisions. We will investigate if this is a contributing factor or not first before any further changes beside correcting the `DateTime.UtcNow` => `value.insertTime` is needed.

bclothier:
The last commit adds stack some limited stack tracing which I think helps enormously in understanding from where they are made. A large majority of calls has came from the references so that may be a place to do some investigating for any leaks.

In case anyone is interested, here's an excel spreadsheet with the latest serialized dumps, both pre and post parse.

[comSafeOutput Before Parse.xlsx](https://github.com/rubberduck-vba/Rubberduck/files/2471681/comSafeOutput.Before.Parse.xlsx)
[comSafeOutput After Parse.xlsx](https://github.com/rubberduck-vba/Rubberduck/files/2471680/comSafeOutput.After.Parse.xlsx)


I've come to the conclusion that because I'm sprinkling way too much `DEBUG` in too many places, I will be creating a `ComSafeBase` abstract class so that the implementations can just call the base method and thus keep the debugging stuff tucked away and not clutter the implementation details. I will work on that next.

Rubberduck.VBEEditor/ComManagement/StrongComSafe.cs
```diff
@@ -1,32 +1,43 @@
...
 
 namespace Rubberduck.VBEditor.ComManagement
 {
-    public class StrongComSafe: IComSafe
+    public class StrongComSafe: ComSafeBase
     {
         //We override the equality comparison and hash code because subclasses of SafeComWrapper<T> override the corresponding methods.
         //We need to distinguish between the individual wrapper instances no matter whether they are semantically equal.
-        private readonly ConcurrentDictionary<ISafeComWrapper, byte> _comWrapperCache = new ConcurrentDictionary<ISafeComWrapper, byte>(new ReferenceEqualityComparer());
+        private readonly ConcurrentDictionary<int, ISafeComWrapper> _comWrapperCache = new ConcurrentDictionary<int, ISafeComWrapper>();
```
bclothier: The change was made primarily to simplify the implementation of `GetWrappers` and encourage consistency in how we hash between the 2 versions of safe. If this is unwanted, please let me know.

bclothier:
The changes were refactored into an abstract class which helps somehow with minimizing the duplication of the debug-only serialization implementation. Unfortunately, not all debug-only operations can be easily encapsulated in the abstract so there are 2 places where both strong & weak implementations has to have debug section. While an improvement over the mess I made previous, I'm hoping someone might suggest even better improvement to keep things clean.

Rubberduck.VBEEditor/ComManagement/StrongComSafe.cs
MDoerner: I do not like this change because it removes the safety net for hash collisions. (When encountering the same hash, an equality check gets performed.) Having the wrapper as a key is the clean way to simulate a `ConcurrentHashSet`. Unfortunately, this is not possible in the weak COM safe because it requires a hard reference to the wrapper.

bclothier: Ok, will revert that and modify the `GetWrappers` method to provide the hashcode. Does that imply that the weak COM safe lacks that safety net?

MDoerner: Yes, the weak COM safe does not have the safety net because it is impossible to have it without holding a strong reference to the SCW.

bclothier:
I had realized my original trace implementation was totally boneheaded. As a result, I made the decision to split the logs into 2 files; one for reporting the contents at the instant within the COM Safe and other for tracing the activity within the COM safe.

They can be correlated using Excel, with column `Key`, which is derived from the hash codes, which will enable us to see which GC'd SCWs came from and how long they were in the safe.

I also have overloaded the `Serialize` button to allow easy serialization of the COM safe's contents and the trace log. Note that due to size of trace log, I cannot hold it in-memory, so I flush the content to the file as soon as I get any activity in the COM Safe. However, if nobody clicks the `Serialize` button, the file gets deleted. If it has been ever serialized at once; an additional file will be saved to same directory (currently `%APPDATA%\Rubberduck\COM Trace\`) with ` final` in the name to indicate the full trace log up to the shutdown of the COM Safe. 

I had ran into some failing tests due to parallel runs of tests trying to use same file. This is a test-only problem since there would be only one instance of COM safe. Nonetheless, I elected to add `GetHashCode` to the log file name to ensure that it won't error. There are also synchronization around the file writes for obvious reasons. This implies that when running with `DEBUG` flag, COM Safe may perform a bit more slowly. 

Just for others' references, I'm attaching the formatted files from last run:
[COM Safe Stack Trace 20181014030024.xlsx](https://github.com/rubberduck-vba/Rubberduck/files/2476565/COM.Safe.Stack.Trace.20181014030024.xlsx)
[COM Safe Content Snapshot 20181014030024.xlsx](https://github.com/rubberduck-vba/Rubberduck/files/2476566/COM.Safe.Content.Snapshot.20181014030024.xlsx)

Rubberduck.Core/UI/Command/MenuItems/CommandBars/SerializeProjectsCommandMenuItem.cs
```diff
@@ -76,6 +76,16 @@ protected override void OnExecute(object parameter)
                 Logger.Warn($"Serializing {library.Path}.");
                 _serializationProvider.SerializeProject(library);
             }
+
+#if DEBUG
```
comintern: Isn't the precompiler directive redundant here?

Rubberduck.VBEEditor/ComManagement/IComSafe.cs
```diff
@@ -7,5 +7,8 @@ public interface IComSafe: IDisposable
     {
         void Add(ISafeComWrapper comWrapper);
         bool TryRemove(ISafeComWrapper comWrapper);
+#if DEBUG
+        void Serialize(string targetDirectory);
```
comintern: I'm not entirely thrilled with the idea of a conditionally compiled interface, especially using DEBUG (I generally always have that flag set unless I'm running tests right before a PR, and I sometimes forget to do that).  

If we're using the DEBUG flag here, it needs XML doc warning not to use it outside of a DEBUG directive, and concrete implementations should inherit that doc. That should probably be a general rule for conditionally compiled interfaces.

Rubberduck.Core/UI/Command/MenuItems/CommandBars/SerializeProjectsCommandMenuItem.cs
```diff
@@ -76,6 +76,16 @@ protected override void OnExecute(object parameter)
...
+            //Dumb hack cos I'm too lazy to wire up another button...
+            var traceDirectory = Path.Combine(Path.GetDirectoryName(_serializationProvider.Target), "COM Trace");
```
comintern: Nitpick. Can this go in its own folder? I don't want to forget about it if we eventually decide to perform some build actions around the serialized `_serializationProvider.Target`.

bclothier: It is already going into its own folder called `COM Trace`? Besides, even if we went with build action, we would not need to do this via the `CommandMenuItem` but rather the `Serialize***` method for the respective classes, no?

Rubberduck.Core/UI/Command/MenuItems/CommandBars/SerializeProjectsCommandMenuItem.cs
bclothier: AFAICT, this file doesn't have any other `#DEBUG` anywhere else. Besides, given that the `Serialize` method can only exist in a `DEBUG` build, it makes sense to wrap the invocation inside a `DEBUG` block to ensure it doesn't end up calling a non-existent method.

comintern: My thought when I suggested that was that the command only gets injected in DEBUG. I missed the fact that `Serialize` is also conditionally on the interface.

Rubberduck.Core/UI/Command/MenuItems/CommandBars/SerializeProjectsCommandMenuItem.cs
comintern: Own folder as in not a sub-folder of the serialized projects path (which should probably be named something other than Declarations now, but that's on me...).  Like I said, nitpick.

PR #4428, opened 2018-10-12
Revert use of 'default' language feature to lower c# version

IvenBach:
Use of `default` was keeping me from building.


Vogel612:
FWIW we should be consistently on language version 7.2 as of now. With the csproj update that will be true for every project in the solution. 

`default` will be available in all projects then.

IIRC newer language versions require a newer .NET framework. We're currently supporting back to .NET 4.6, which notably includes Vista (and subsequent versions)

PR #4438, opened 2018-10-15
Misc bug fixes

comintern:


comintern:
Failing tests only fail in Release, and only fail if they are not run individually. Is the test setup borked again?

PR #4441, opened 2018-10-15
Cache deserialized ComProjects in the tests instead of declarations

MDoerner:
Previously, we statucally cached serialized declarations we loadedfrom file in tests to avoid loading several MB of text files multiple times in different tests. This is no longer sensible because we derserialize ComProjects now and generate the declarations based on them.


MDoerner:
I just realized that this change has the added benefit that we no longer cache objects we change while parsing. More precisely, previously we had to worry about the references we save on the declarations. These got removed on the each load, whoch made it inherently unsafe to run tests in parallel. Now, this is a non-issue because we create the declarations each time anew from the cached command projects.

PR #4447, opened 2018-10-18
Trace logging by default; change to None after successful run

### LINKED ISSUE #3861: The log should default to `Trace` for first startup after install
It would be much easier to troubleshoot the cases where Rubberduck crashes after having just installed. In this situation, the users have no way to get to the Settings to ask for trace and log the crash because they kind of have to *start* it. 

It is suggested that a Nlog rule be added so if it is the first time Rubberduck's launching on a system, (or maybe always for all startup) that log run at `Trace` level, thus ensuring that even if a crash happens the users has logs to share with us so we can make it *not* crash.

bclothier:
Closes #3861 

Introduces new setting that isn't UI-visible, `UserEditedLogLevel` that allows us to ship Rubberduck with `Trace` as the default logging level. If the user do not edit it, at the shutdown of the Rubberduck, the log level will be changed to `None`.

### DIFF
--- a/Rubberduck.Core/Rubberduck.Core.csproj
+++ b/Rubberduck.Core/Rubberduck.Core.csproj
@@ -123,9 +123,10 @@
       <AutoGen>True</AutoGen>
       <DesignTimeSharedInput>True</DesignTimeSharedInput>
       <DependentUpon>Settings.settings</DependentUpon>
+      <DesignTime>True</DesignTime>
     </Compile>
     <None Include="Properties\Settings.settings">
-      <Generator>PublicSettingsSingleFileGenerator</Generator>
+      <Generator>SettingsSingleFileGenerator</Generator>
       <LastGenOutput>Settings.Designer.cs</LastGenOutput>
     </None>
   </ItemGroup>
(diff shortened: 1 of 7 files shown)


Rubberduck.Core/Rubberduck.Core.csproj
```diff
@@ -123,9 +123,10 @@
       <AutoGen>True</AutoGen>
       <DesignTimeSharedInput>True</DesignTimeSharedInput>
       <DependentUpon>Settings.settings</DependentUpon>
+      <DesignTime>True</DesignTime>
     </Compile>
     <None Include="Properties\Settings.settings">
-      <Generator>PublicSettingsSingleFileGenerator</Generator>
+      <Generator>SettingsSingleFileGenerator</Generator>
```
bclothier: This change wasn't made by me directly. I am not sure if that's a good thing or not.

Vogel612: might be an issue. The SettingsSingleFileGenerator just generates internal-visible code. Especially for Rubberduck.Resources that change would break compilation. If it doesn't break compilation, that should be fine

bclothier: I see. I wonder how that got changed as I don't remember fiddling with that. It did build and run tests locally, however. I'll need to actually F5 test it tonight, though.

Vogel612: If it builds, then it's not an issue to have that change. I assume that it correctly defaults to least visibility modifying the resource itself. Since you changed the file that's involved here, VS automatically pushed it into the lower visibility on build

PR #4456, opened 2018-10-24
Clean up COM leaks in Rubberduck

bclothier:
Using the serialize method introduced in PR #4426, I was able to locate several locations in code where we failed to properly dispose of COM wrappers and remedied them. 

This may not be exhaustive as the serialize technique is only good when the code has ran. This covers basic startup/doing parsing/shutdown. Other commands may be still leaking but it's a start.

Of particular interest is the changes to the event args to avoid marshaling COM wrappers across event boundaries.


Rubberduck.Core/Common/RubberduckHooks.cs
```diff
@@ -18,8 +18,16 @@ public class RubberduckHooks : SubclassingWindow, IRubberduckHooks
...
+        {
+            using (var window = vbe.MainWindow)
+            {
+                return (IntPtr)window.HWnd;
+            }
+        }
+
         public RubberduckHooks(IVBE vbe, IGeneralConfigService config, HotkeyFactory hotkeyFactory, AutoCompleteService autoComplete)
-            : base((IntPtr)vbe.MainWindow.HWnd, (IntPtr)vbe.MainWindow.HWnd)
+            : base(GetVbeMainWindowPtr(vbe), GetVbeMainWindowPtr(vbe))
```
comintern: I'm not sure I like the double function call here (probably my fault originally) - this would be better with a single argument overload for the ctor that sets both the `hWnd` and the `subclassId` to the passed value.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs
```diff
@@ -68,8 +68,9 @@ public CodeExplorerComponentViewModel(CodeExplorerItemViewModel parent, Declarat
                         var component = _projectsProvider.Component(qualifiedModuleName);
                         string parenthesizedName;
                         using (var properties = component.Properties)
+                        using (var nameProperty = properties["Name"])
                         {
-                            parenthesizedName = properties["Name"].Value.ToString() ?? string.Empty;
+                            parenthesizedName = nameProperty.Value.ToString() ?? string.Empty;
```
comintern: Missing in the original code too, but this could use a null check on `nameProperty`.

Rubberduck.Core/UI/Command/ExportAllCommand.cs
```diff
@@ -32,17 +31,32 @@ protected override bool EvaluateCanExecute(object parameter)
...
+            using (var activeProject = _vbe.ActiveVBProject)
+            {
+                return Evaluate(activeProject);
+            }
         }
 
         private bool Evaluate(IVBProject project)
         {
-            return project != null && !project.IsWrappingNullReference && project.VBComponents.Count > 0;
+            if (project == null || project.IsWrappingNullReference)
```
comintern: Not really in scope for the PR (commenting for my own reference), but this wants to be a generic extension someplace due to how many places we make this check. I'm thinking something along the lines of:

```
        public static bool IsLiveWrapper<T>(this SafeComWrapper<T> wrapper) where T : class
        {
            return wrapper != null && !wrapper.IsWrappingNullReference;
        }
```

MDoerner: Doesn't the `SafeComWrapper` override `==` anyway to mean this?

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs
bclothier: I'm not sure I follow this. Wouldn't it have had thrown on the `var nameProperty = properties["Name"]` if it didn't exist? I don't think it's possible for a `IProperties` to contain a null property that's named? 

In fact, R# flags the `?? string.Empty` as redundant since `ToString()` can never be `null` apparently. I changed it to `nameProperty.Value?.ToString() ?? string.Empty;` because `Value` is an `object` and could be `null`. That made R# happy. Is what you meant?

Rubberduck.Core/UI/Command/ExportAllCommand.cs
comintern: The `==` override doesn't check whether it's wrapping a null reference.

MDoerner: Actually, it does but in an indirect way that makes the `==` asymmetrical. When `a` in `a == b` is not `null'', then the override returns `a.Equals(b)`. The implementation in `VBProject` uses `IsEqualIfNull` inherited from `SafeComWrapper`. This in turn returns true if `a` is wrapping a null reference.

I have to say that I really do not like that the operator is asymmetric in this way and that the derived classes can choose to honor this intended behaviour or not. (If it was not intended, why would there be the protected method `IsEqualIfNull`?)

.gitignore
```diff
@@ -181,3 +181,5 @@ Installers/
 
 CodeGraphData/
 /Rubberduck.Deployment/Properties/launchSettings.json
+/Rubberduck.Deployment/Rubberduck.API.idl
+/Rubberduck.Deployment/Rubberduck.idl
```
Vogel612: why not just `**/*.idl`?

Vogel612: worst case, those can be whitelisted... on the other hand it doesn't make all that much of a difference :)

comintern: We'd need a static .idl if we ever decided to put the COM collector under test, no?

bclothier: Indeed though they'd be in the `RubberduckTests` project, not here. But I think that's good enough reason to not wildcard them.

Rubberduck.Core/UI/UnitTesting/Commands/AddTestModuleCommand.cs
```diff
@@ -131,6 +131,8 @@ private IVBProject GetProject()
                 return activeProject;
             }
 
+            activeProject.Dispose();
```
MDoerner: Why do we try to dispose a project here that is wrapping a null reference?

Rubberduck.Core/UI/UnitTesting/Commands/AddTestModuleCommand.cs
```diff
@@ -154,98 +158,100 @@ protected override void OnExecute(object parameter)
         {
             var parameterIsModuleDeclaration = parameter is ProceduralModuleDeclaration || parameter is ClassModuleDeclaration;
 
-            var project = parameter as IVBProject ??
-                          (parameterIsModuleDeclaration ? ((Declaration) parameter).Project : GetProject());
-
-            if (project == null || project.IsWrappingNullReference)
+            using (var project = parameter as IVBProject ??
+                                 (parameterIsModuleDeclaration ? ((Declaration) parameter).Project : GetProject()))
```
MDoerner: This using statement does not seem to be safe: if the `parameter` is an `IVBProject`, the lifetime has to be managed by the caller; if it is a module declaration, the project is coming from the declarations, which ultimately is managed by the projects repository. Only the very last option should be wrapped in a using statement.

Rubberduck.VBEEditor/QualifiedModuleName.cs
```diff
@@ -149,4 +149,44 @@ public override bool Equals(object obj)
             return !a.Equals(b);
         }
     }
+
+    public static class QualifiedModuleNameExtensions
```
MDoerner: Why are the extensions in the same file as the struct itself? Either these should be part of the struct, which I actually would not prefer, or it should be in their own file in the `Extensions` namespace.

MDoerner:
Sorry for reviewing this only now, but I did not get around to it earlier (before it got merged) and have been pointed here from the issue mentioning the RaceOnRCWCleanup.

Rubberduck.Core/UI/UnitTesting/Commands/AddTestModuleCommand.cs
bclothier: This is primarily to remove it from the ComSafe and thus prevent it be GC'd. A `Dispose` on a null-wrapping reference will not actually `Release` the wrapped COM object so it's safe to do.

That said, I really should have fixed the method to use `using` to be more idiomatic.

Rubberduck.Core/UI/UnitTesting/Commands/AddTestModuleCommand.cs
bclothier: Good catch; fixed in my round 2 branch.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProjects.cs
```diff
@@ -119,26 +119,29 @@ void VB._dispVBProjectsEvents.ItemActivated([MarshalAs(UnmanagedType.Interface),
...
-            if (handler == null || !IsInDesignMode())
+            using (var project = new VBProject(vbProject))
             {
-                project.Dispose();
-                return;
-            }
+                var handler = dispatched;
+                if (handler == null || !IsInDesignMode())
+                {
+                    project.Dispose();
```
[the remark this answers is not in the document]
bclothier: Look like an oversight; I had removed that in the VBA's but not the VB6's implementation. Fixed in round 2.

Rubberduck.VBEEditor/QualifiedModuleName.cs
bclothier: I agree about the need for doc comments. I will add it. This is currently not in use. I wrote them up originally to help address the problems around the ownership when marshaling across the events. However, when I worked on the `EventArg`s classes, I found that we actually did not really need to marshal any SCW at all. I was able to achieve what I needed using the QMNs alone or in the case of the command bar, by copying some of its properties, since that was all what the callers needed. 

Should we run into a legitimate scenario where we need to marshal a SCW across the event, the intention is encourage the callers to take a dependency on the repository and take the SCW out there than directly via the `EventArg` implementations. 

The methods cannot be a part of the struct for reasons I can't remember. I left them in the same namespace primarily to avoid the need to add yet another namespace just to access the function and thus miss opportunities for where it may be used.

Rubberduck.Core/UI/UnitTesting/Commands/AddTestModuleCommand.cs
MDoerner: SCWs wrapping a null reference do not get added to the COM safe. So, the dispose here is actually a no-op.

bclothier: Hmm. You're right. Yet I was bought here because from the log, this got GC'd....

Issue #4273, opened 2018-08-14
COMException in FormDesignerRefactorRenameCommand


comintern:
Found this in the log file posted on #4202:

```
2018-07-17 16:24:49.7148;ERROR-2.2.0.3508;Rubberduck.UI.Command.Refactorings.FormDesignerRefactorRenameCommand;System.Runtime.InteropServices.COMException (0x800A09A3): Exception from HRESULT: 0x800A09A3
   at Microsoft.Vbe.Interop._VBComponent.get_Designer()
   at Rubberduck.VBEditor.SafeComWrappers.VBA.VBComponent.get_HasDesigner() in C:\projects\rubberduck\Rubberduck.VBEditor.VBA\SafeComWrappers\VB\VBComponent.cs:line 85
   at Rubberduck.UI.Command.Refactorings.FormDesignerRefactorRenameCommand.GetTarget(Nullable`1 qualifiedModuleName) in C:\projects\rubberduck\Rubberduck.Core\UI\Command\Refactorings\FormDesignerRefactorRenameCommand.cs:line 62
   at Rubberduck.UI.Command.Refactorings.FormDesignerRefactorRenameCommand.EvaluateCanExecute(Object parameter) in C:\projects\rubberduck\Rubberduck.Core\UI\Command\Refactorings\FormDesignerRefactorRenameCommand.cs:line 29
   at Rubberduck.UI.Command.CommandBase.CanExecute(Object parameter) in C:\projects\rubberduck\Rubberduck.Core\UI\Command\CommandBase.cs:line 28;System.Runtime.InteropServices.COMException (0x800A09A3): Exception from HRESULT: 0x800A09A3
   at Microsoft.Vbe.Interop._VBComponent.get_Designer()
   at Rubberduck.VBEditor.SafeComWrappers.VBA.VBComponent.get_HasDesigner() in C:\projects\rubberduck\Rubberduck.VBEditor.VBA\SafeComWrappers\VB\VBComponent.cs:line 85
   at Rubberduck.UI.Command.Refactorings.FormDesignerRefactorRenameCommand.GetTarget(Nullable`1 qualifiedModuleName) in C:\projects\rubberduck\Rubberduck.Core\UI\Command\Refactorings\FormDesignerRefactorRenameCommand.cs:line 62
[...]

PR #4458, opened 2018-10-25
Host application document

### LINKED ISSUE #3614: Exceptions on exiting
**NOTE**: This issue is meant to be an aggregation of other issues, to bring up to the date the current state of the persistent problem of getting a clean exit consistently everywhere. Apparently it is reported to occur on some machines, but not all.However, for those machines where it do happens, it seems to be fairly consistent. It is important to note that the issue may have had multiple causes and though we've made significant progress in reducing them. In fact, the current version of Rudderduck is much better behaved compared to previous versions where it was more apparent and crashed hard. Nonetheless we are still dealing with a silent zombie on the shutdown.

**Problem**

Whenever the following are true:

1) VBA host was opened
2) Rubberduck was loaded and has DockWindows attached. In testing, we used Code Explorer docked and visible.

Then whenever the user closes the application -- it does *not* matter how the user closes, whether the VBA window was closed previously, or not -- then application will zombify silently. There is no `There was a problem and Windows is trying to find a solution`. Rather the application is left hidden and must be then killed via the task manager.

Errors look similar to this:

`Exception thrown at 0x75CABB49 (ole32.dll) in MSACCESS.EXE: 0xC0000005: Access violation reading location 0x0ACCFC34. occurred`

image

or

*NOTE*: See next section for comments; this may be atypical.
*NOTE 2*: It's likely that `KernelBase.dll` is noise because this is exception handling. This should be ignored.

`Exception thrown at 0x74E0C54F (KernelBase.dll) in MSACCESS.EXE: 0xE0434352 (parameters: 0x80004001, 0x00000000, 0x00000000, 0x00000000, 0x720D0000). occurred`

image

**What has been already tried**

We already created a MCVE to demonstrate and while the numbers of errors aren't the same, they are similar enough to be possibly helpful. The slimmed version can be [found here](https://github.com/rubberduck-vba/SlimDucky).

This is *NOT* host-specific. Though the above examples were from Access, reports has been made for Excel and Word and likely other hosts and errors are originating from the DLLs that are loaded by those hosts.

At one point, we tried explicitly releasing the objects, as [documented here](https://rubberduckvba.wordpress.com/2016/10/10/2-0-stabilizing/). But in the end we commented out the code because it was actually doing more damage than help. Further research indicates that we should not wantonly call `Marshal.RelaseComObject`, especially for objects we don't own. 

However, @MDoerner  did investigate the possibility of adding `Marshal.ReleaseComObject` to the `CommandBarControls` which are in use on the toolbars and in the menus since we *do* create those. Those did not help eliminate the errors but did not increase it, either.

As [alluded already, in the earlier build](https://rubberduckvba.wordpress.com/2016/10/10/2-0-stabilizing/), we did a massive refactor where we wrap all objects in an abstraction, which was meant to implement the `Release`, but even without that, we at least ensure that all our access go through the abstraction so that we have a well-defined points where we instantiate and where we destroy the references to COM objects we do not own. 

We also tried to implement `IDisposable`, as documented [here](https://codereview.stackexchange.com/questions/142916/wrapping-com-objects-with-idisposable#) but this did not help and actually caused [problems](https://codereview.stackexchange.com/questions/143358/properly-destroying-the-vbe).

We did fix a very subtle bug with correctly unregistering the event sinks used with [this](https://github.com/rubberduck-vba/Rubberduck/pull/3277) and [that](https://github.com/rubberduck-vba/Rubberduck/pull/3416). 

In previous versions, whenever we had a debugger attached and `Native` mode is enabled, the program would have broke into the dissembly and waited on our response. However with the current version (2.1.1), the error does show up but only briefly; after few seconds, the debugger abruptly stops. We think it's because the .NET part has already shutdown, but we are getting access violation long after. If we change the debugger mode to `Native` exclusively, we are able to make the program break indefinitely, allowing us to inspect the stack trace. 

@bclothier also had installed [a hotfix](https://support.microsoft.com/en-us/help/2718841/-0xc0000005-error-message-and-windows-explorer-stops-responding-on-a-c) which was totally unrelated to the issue but did share same symptoms to see whether it would rule out the possibility of a defective Windows DLL outside the usercode. This did change the errors on the shutdown, causing it to error in `KernelBase.dll` rather than `ole.dll` but did not ultimately fix it. Also, this was limited to Windows 7 and the error has been reported on other operating systems.  That said, it was determined that the same error occurred even in Excel as well. 

**Additional details**

You can review the other issues in [this project](https://github.com/rubberduck-vba/Rubberduck/projects/14) to get more background or specific subsets that may or may not be relevant to the issue at hand.

**References**

Listed are links that we've already read in attempt to help us understand and implement any fixes.

https://msdn.microsoft.com/en-us/library/office/aa159887%28v=office.11%29.aspx?f=255&MSPPError=-2147217396

https://blogs.msdn.microsoft.com/visualstudio/2010/03/01/marshal-releasecomobject-considered-dangerous/

https://blogs.msdn.microsoft.com/cbrumme/2003/04/16/releasecomobject/

https://blogs.msdn.microsoft.com/oldnewthing/20100813-00/?p=13153

### LINKED ISSUE #3905: MS Access leaves a ghost process
Win10, Office 2016 64-bit.

Open Access, open VBE, parse project, close Access. MSACCESS.EXE remains in Task Manager.

[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/1889746/RubberduckLog.txt)

bclothier:
Closes #3905 
Closes #3614 

This introduces `IHostDocument` interface and `GetDocument`/`GetDocuments` to provide additional metadata about document modules from host application directly rather than via VBIDE API.

This resolves a problem with using the `VBComponent.Properties` which has been shown to be fragile when the document is not in design time. Accessing those cause runtime error and though there is a `try` and appropriate `using`s to swallow the exceptions and dispose of the wrappers object, it will cause a ghost process with Access and possibly other hosts where document modules may have a runtime and design-time state.

However, what I'm not happy with is the promiscuous COM slinging because there are no SCW wrappers defined for host-specific interfaces. Even if we did want to define one, it does not make much sense as that needs to be generic for all host documents as possible. Therefore, I welcome any ideas or feedback on making it less..... toxic.

Oh, BTW, unit tests are borked because of new parameter in the constructors. That will need to be fixed.


bclothier:
Reminder to myself to see if #4273 is in-scope for the PR as it likely needs the document state.

Rubberduck.VBEEditor/SafeComWrappers/VB/Abstract/IHostApplication.cs
```diff
@@ -1,36 +1,92 @@
...
+        /// The document is open in non-design mode. It may not be safe to parse the document in this state.
+        /// </summary>
+        ActiveView
+    }
+
+    public interface IHostDocument : IDisposable
+    {
+        string Name { get; }
+        string ClassName { get; }
+        WeakReference<object> Target { get; }
```
bclothier: Thinking about it - I think this is not ideal. This should be probably `bool TryGetTarget(out object)` instead. This way no references is held  within the implementation of `IHostDocument` and the consumer can dispose of the object. 

OTOH, YAGNI...

bclothier:
After thinking about it a bit - I decided I'd rather consistently wrap all COM objects in SCWs. Otherwise the debugging efforts will be exceedingly hard should the scope grow. 

We can leverage the fact that a `dynamic` object in .NET can be used exactly like a `IDispatch` and thus be able to late-bind to a COM object. That allows us to define SCWs without having to create a full blown SCW re-implementing the underlying COM interface. Since the work with the document interfaces are meant to be limited in scope and since it may come from a variety of host, that seems the best way to address the problem. Thus, we provide `SafeIDispatchWrapper` class for that reason. 

The `HostDocument` class now provide a function (which is still unused) to retrieve a non-generic `SafeIDispatchWrapper` class in case any other parts of RD need to access the host's OM without having to take on a dependency on the interop library. As a reminder, we want to keep a strict separation between the .NET and COM worlds with the `Rubberduck.VBEditor` being the seam; no other projects may directly reference interop libraries. The `SafeIDispatchWrapper` helps protect that separation.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerCustomFolderViewModel.cs
```diff
@@ -19,11 +20,14 @@ public class CodeExplorerCustomFolderViewModel : CodeExplorerItemViewModel
         };
 
         private readonly IProjectsProvider _projectsProvider;
+        private readonly IVBE _vbe;
```
retailcoder: Given `IVBE` being among the very last objects that get destroyed (if at all - or did that change?), does this field not make the viewmodel live longer than it needs to?

Rubberduck.VBEEditor/SafeComWrappers/SafeIDispatchWrapper.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            IDispatchPointer = GetPointer(target);
+        }
+
+        public IntPtr IDispatchPointer { get; }
+
+        public void Invoke(Action<dynamic> action)
+        {
+            action.Invoke(Target);
```
retailcoder: Are we shielded against a dynamic binding exception here?

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerCustomFolderViewModel.cs
bclothier: This is provided by CW, which is the same instance that we got from the `Extension.cs` class. A separate PR would need to review all the VBE usage and ensure that the classes are dead by time the .NET parts shut down.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs
```diff
@@ -113,7 +116,16 @@ private bool ContainsBuiltinDocumentPropertiesProperty()
             var component = _projectsProvider.Component(Declaration.QualifiedName.QualifiedModuleName);
             using (var properties = component.Properties)
             {
-                return properties.Any(item => item.Name == "BuiltinDocumentProperties");
+                foreach (var property in properties)
+                using(property)
```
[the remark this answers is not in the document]
bclothier: Confused. Thought we agreed to stacking. As some areas could require a number of `using`s, nesting each would make it quite... deep.

Rubberduck.VBEEditor/SafeComWrappers/SafeIDispatchWrapper.cs
bclothier: Good question! I added a test in live code temporarily:

```
//testing
accessObject.Invoke(d => _logger.Trace("Dynamically invoked method:" + d.Name));
```

and got the output in the log:
```
2018-10-27 00:10:13.9527;TRACE-2.2.6874.274;Rubberduck.VBEditor.SafeComWrappers.SafeComWrapper`1;Dynamically invoked method:Form1;
```

So, no problem there with dispatching via `dynamic` in a lambda. Note that we provide the `dynamic`, rather than the caller. 

That said, the whole lambda is now dynamic and therefore, it'll have to be the caller's responsibility that there's appropriate handling if things don't go as expected.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerComponentViewModel.cs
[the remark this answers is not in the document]
retailcoder: Stacked `using` is fine. Other blocks get their braces ;-)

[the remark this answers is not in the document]
retailcoder: Stacked `using` is fine. Other blocks get their braces ;-)

PR #4488, opened 2018-11-03
Add new inspection for Excel UDFs hidden by cells.

comintern:
After seeing yet *another* [question where this popped up](https://stackoverflow.com/questions/53134535/udf-works-in-vba-returns-ref-in-workbook) on SO, I decided to toss a quick inspection together for it.

The inspection will flag public functions in standard modules that have names that collide with cell identifiers.

> Functions that are visible to Excel as User-Defined Functions will return a '#REF' error when used on a Worksheet if they match the name of a valid cell reference. If the function is intended to be used as a UDF, it must be renamed. If the function is not intended to be used as a UDF, it should be scoped as 'Private' or moved out of a standard Module.


BZngr:
FWIW, if the inspection's evaluation logic can be packaged and made accessible to non-inspection classes as well, then it can be used to enhance `RenameRefactoring` (and maybe other refactorings) that proactively validate/check new procedure names entered by the user.

rkapka:
Your latest commit doesn't include any changes to `Settings.Designer.cs`.

PR #4490, opened 2018-11-04
Address potential race condition and potentially un-disposed objects …

comintern:
…on TryAdd failure.


bclothier:
Ref https://github.com/DotNetAnalyzers/IDisposableAnalyzers/issues/125#issuecomment-435656434

PR #4491, opened 2018-11-04
Clean COM wrappers leak: Round 2

bclothier:
1) This is more exhaustive investigation for any possible leaks. This was done by executing all actions in all toolwindows and tracking down any leaks.
2) Unloading & reloading was also tested for any leaks (Ref #3645)
3) Temporarily `IDisposableAnalyzers` which helped few legitimate misuse of `IDisposable`. However I ended up commenting it out as we have far too many false positive. I left it as comment so once a PR addressing the issue https://github.com/DotNetAnalyzers/IDisposableAnalyzers/issues/126 is merged, we can re-test using it.
4) Address some comments to the 1st round.

As noted in the comments for #3645, not all items in the COM safe get disposed due to `RaceOnRCWDisconnect` errors. This is attributed to the fact that there may be other threads running that accesses the SCWs during the shutdown. The PR does not address this as this should be its own PR. One possible fix for that is to implement the `VBEEvents.TerminateEvents` to put in the necessary kill switch for command buttons and any other objects that needs to query a SCW.


Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProject.cs
```diff
@@ -100,7 +100,17 @@ public override int GetHashCode()
 
         public IReadOnlyList<string> ComponentNames()
         {
-            return VBComponents.Select(component => component.Name).ToArray();
+            var names = new List<string>();
+            using (var components = VBComponents)
+            {
+                foreach(var component in components)
```
comintern: I generally prefer stacked `using`'s, but a stacked `foreach` is as dangerous as `if (foo) bar;`. I probably would have done something like:

`return VBComponents.Select(component => { var name = component.Name; component.Dispose(); return name; }).ToArray();`

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProjects.cs
```diff
@@ -96,23 +96,25 @@ void VB._dispVBProjectsEvents.ItemRemoved([MarshalAs(UnmanagedType.Interface), I
...
-            if (!IsInDesignMode() || VBProject.Protection == VB.vbext_ProjectProtection.vbext_pp_locked)
+            using (var project = new VBProject(VBProject))
             {
-                project.Dispose();
-                return;
-            }
+                if (!IsInDesignMode() || VBProject.Protection == VB.vbext_ProjectProtection.vbext_pp_locked)
+                {
+                    project.Dispose();
+                    return;
```
comintern: Doesn't this call `Dispose` twice?

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProjects.cs
```diff
@@ -96,23 +96,25 @@ void VB._dispVBProjectsEvents.ItemRemoved([MarshalAs(UnmanagedType.Interface), I
...
-            var handler = ProjectRenamed;
-            if (handler == null || projectId == null)
-            {
-                project.Dispose();
-                return;
+                var handler = ProjectRenamed;
+                if (handler == null || projectId == null)
+                {
+                    project.Dispose();
+                    return;
```
comintern: See above - the `return` should already be calling `Dispose` as the block exits.

Rubberduck.Core/Common/WinAPI/RegistryAccess.cs
```diff
@@ -18,9 +18,11 @@ internal static RegistryKey GetDeviceKey(string device)
 
         internal static string GetClassType(string classGuid)
         {
-            var classGuidKey = Registry.LocalMachine.OpenSubKey(@"SYSTEM\CurrentControlSet\Control\Class\" + classGuid);
-
-            return classGuidKey != null ? (string)classGuidKey.GetValue("Class") : string.Empty;
+            using (var classGuidKey =
+                Registry.LocalMachine.OpenSubKey(@"SYSTEM\CurrentControlSet\Control\Class\" + classGuid))
```
comintern: This isn't safe. `OpenSubKey` can return `null`.

Rubberduck.SmartIndenter/IndenterSettings.cs
```diff
@@ -134,11 +134,13 @@ public IndenterSettings(bool skipRegistry)
             {
                 try
                 {
-                    var reg = Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\VBA\6.0\Common", false) ??
-                              Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\VBA\7.0\Common", false);
-                    if (reg != null)
+                    using (var reg = Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\VBA\6.0\Common", false) ??
+                                     Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\VBA\7.0\Common", false))
```
comintern: `OpenSubKey` can return `null`. This would probably be better without a `using` block here, since there is already a `reg != null` test below where it can be explicitly `Dispose`'d

Rubberduck.SmartIndenter/IndenterSettings.cs
```diff
@@ -219,45 +221,52 @@ public void LoadLegacyFromRegistry()
...
-                ForceDebugAssertInColumn1 = ForceDebugStatementsInColumn1;
-                ForceStopInColumn1 = ForceDebugStatementsInColumn1;
-                ForceCompilerDirectivesInColumn1 = GetSmartIndenterBoolean(reg, "CompilerCol1", ForceCompilerDirectivesInColumn1);
-                IndentCompilerDirectives = GetSmartIndenterBoolean(reg, "IndentCompiler", IndentCompilerDirectives);
-                AlignDims = GetSmartIndenterBoolean(reg, "AlignDim", AlignDims);
-                AlignDimColumn = GetSmartIndenterNumeric(reg, "AlignDimCol", AlignDimColumn, MaximumAlignDimColumn);
-
-                var eolSytle = reg.GetValue("EOLComments") as string;
-                if (!string.IsNullOrEmpty(eolSytle))
+                using (var reg = Registry.CurrentUser.OpenSubKey(LegacySettingsSubKey, false))
```
comintern: `OpenSubKey` can return null - this would be better as an explicit call to `Dispose`.

Rubberduck.UnitTesting/UnitTesting/VBEInteraction.cs
```diff
@@ -67,36 +67,45 @@ public void EnsureProjectReferencesUnitTesting(IVBProject project)
             if (project == null || project.IsWrappingNullReference) { return; }
             var libFolder = IntPtr.Size == 8 ? "win64" : "win32";
             const string libGuid = RubberduckGuid.RubberduckTypeLibGuid;
-            var pathKey = Registry.ClassesRoot.OpenSubKey($@"TypeLib\{{{libGuid}}}\{_rubberduckVersion.Major}.{_rubberduckVersion.Minor}\0\{libFolder}");
-
-            var referencePath = pathKey?.GetValue(string.Empty, string.Empty) as string;
-            string name = null;
-
-            if (!string.IsNullOrWhiteSpace(referencePath))
+            using (var pathKey = Registry.ClassesRoot.OpenSubKey(
```
comintern: `OpenSubKey` can return null. This needs a null check after `pathKey` is set, and it should be explicitly `Dispose`'d.

Rubberduck.UnitTesting/UnitTesting/VBEInteraction.cs
```diff
@@ -67,36 +67,45 @@ public void EnsureProjectReferencesUnitTesting(IVBProject project)
...
-            }
-
-            if (string.IsNullOrWhiteSpace(referencePath) || string.IsNullOrWhiteSpace(name))
-            {
-                throw new InvalidOperationException("Cannot locate the tlb in the registry or the entry may be corrupted. Therefore early binding is not possible");
-            }
+                if (!string.IsNullOrWhiteSpace(referencePath))
+                {
+                    using (var tlbKey =
+                        Registry.ClassesRoot.OpenSubKey(
```
comintern: This throws if `tlbKey` is null when the `using` block exits.

Rubberduck.Core/Common/WinAPI/RegistryAccess.cs
bclothier: This is another case of IDisposable analyers at work. I was surprised to see those not being disposed -- in fact, I was surprised that `Registry` stuff were disposable. As such, I wasn't really thinking too much about the possibility of returning `null`. But that is obviously problematic. I will revise those.

bclothier: Ref https://github.com/DotNetAnalyzers/IDisposableAnalyzers/issues/128

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProject.cs
bclothier: FWIW, I'm told to avoid doing LINQ in VBEditor projects.

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProjects.cs
bclothier: Oversight; should been removed, yes.

Rubberduck.Core/UI/Command/MenuItems/CommandBars/AppCommandBarBase.cs
```diff
@@ -184,7 +202,6 @@ public void RemoveCommandBar()
                     Item.Delete();
                     Item.Dispose();
```
comintern: The dispose call is not necessary here any more - the assignment to null handles that now (as intended).

Rubberduck.Core/UI/Command/MenuItems/CommandBars/AppCommandBarBase.cs
```diff
@@ -181,9 +200,7 @@ public void RemoveCommandBar()
                     Logger.Debug("Removing commandbar.");
                     RemoveChildren();
                     Item.Delete();
-                    Item.Dispose();
                     Item = null;
-                    Parent?.Dispose();
                     Parent = null;
```
Vogel612: Could we get a comment here that nulling them out also disposes them? That helps me not remove these lines

Rubberduck.Core/UI/Command/MenuItems/ParentMenus/ParentMenuItemBase.cs
```diff
@@ -91,7 +110,6 @@ public void RemoveMenu()
             Logger.Debug($"Removing menu {_key}.");
             RemoveChildren();
             Item?.Delete();
-            Item?.Dispose();
             Item = null;
```
Vogel612: same as above re: the comment. Why is the Parent not disposed / nulled here?

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeInfos.cs
```diff
@@ -718,18 +718,19 @@ public TypeInfoWrapper GetControlType(string controlName)
...
         public TypeInfoWrapper GetSafeRefTypeInfo(int hRef)
         {
-            IntPtr typeInfoPtr = IntPtr.Zero;
             // we cast to our ITypeInfo_Ptrs interface in order to work with the raw IntPtr for aggregation
-            ((ITypeInfo_Ptrs)target_ITypeInfo).GetRefTypeInfo(hRef, out typeInfoPtr);
+            ((ITypeInfo_Ptrs)target_ITypeInfo).GetRefTypeInfo(hRef, out var typeInfoPtr);
             var outVal = new TypeInfoWrapper(typeInfoPtr, _isUserFormBaseClass ? (int?)hRef : null); // takes ownership of the COM reference
+
+            _typeInfosWrapped?.Dispose();
             _typeInfosWrapped = _typeInfosWrapped ?? new DisposableList<TypeInfoWrapper>();
```
Vogel612: this looks weird. The list is disposed and then reused? That makes no sense.

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeLibs.cs
```diff
@@ -218,10 +218,10 @@ public void Dispose()
 
         public TypeInfoWrapper GetSafeTypeInfoByIndex(int index)
         {
-            IntPtr typeInfoPtr = IntPtr.Zero;
             // We cast to our IVBETypeLib interface in order to work with the raw IntPtr for aggregation
-            ((ITypeLib_Ptrs)target_ITypeLib).GetTypeInfo(index, out typeInfoPtr);
+            ((ITypeLib_Ptrs)target_ITypeLib).GetTypeInfo(index, out var typeInfoPtr);
             var outVal = new TypeInfoWrapper(typeInfoPtr);
+            _typeInfosWrapped?.Dispose();
             _typeInfosWrapped = _typeInfosWrapped ?? new DisposableList<TypeInfoWrapper>();
```
Vogel612: same as above

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeLibs.cs
```diff
@@ -382,13 +382,15 @@ public void Document(StringLineBuilder output)
...
-            ppTInfo = outVal;
+            ((ITypeLib_Ptrs)target_ITypeLib).GetTypeInfoOfGuid(guid, out var typeInfoPtr);
+            using (var outVal = new TypeInfoWrapper(typeInfoPtr)) // takes ownership of the COM reference
+            {
+                ppTInfo = outVal;
 
-            _typeInfosWrapped = _typeInfosWrapped ?? new DisposableList<TypeInfoWrapper>();
-            _typeInfosWrapped.Add(outVal);
+                _typeInfosWrapped?.Dispose();
+                _typeInfosWrapped = _typeInfosWrapped ?? new DisposableList<TypeInfoWrapper>();
```
Vogel612: and again

Rubberduck.VBEEditor/Events/VbeEvents.cs
```diff
@@ -73,6 +73,7 @@ private void RegisterComponents(string projectId, string projectName)
                 if (item.ProjectId == projectId && item.Name == projectName)
                 {
                     project = item;
+                    item.Dispose();
```
Vogel612: this feels weird. Is that a value copy or a reference copy? Or rather: if we Dispose `item` is `project` also disposed?

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProjects.cs
```diff
@@ -92,23 +92,23 @@ void VB._dispVBProjectsEvents.ItemRemoved([MarshalAs(UnmanagedType.Interface), I
...
-            var handler = ProjectRenamed;
-            if (handler == null || projectId == null)
-            {
-                project.Dispose();
-                return;
+                var handler = ProjectRenamed;
+                if (handler == null || projectId == null)
+                {
+                    return;
+                }
```
Vogel612: this could be simplified to only checking the projectId and invoking the ProjectRenamed with an elvis-operator

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/CodePane.cs
```diff
@@ -102,17 +102,22 @@ private void ForceFocus()
...
+                var handle = mainWindow.Handle().FindChildWindow(window.Caption);
+
+                if (handle != IntPtr.Zero)
+                {
+                    NativeMethods.ActivateWindow(handle, mainWindowHandle);
+                }
+                else
+                {
+                    System.Diagnostics.Debug.WriteLine(
+                        "CodePane.ForceFocus() failed to get a handle on the MainWindow.");
```
Vogel612: we don't have a Logger available here?

RubberduckBaseProject.csproj
```diff
@@ -45,7 +45,15 @@
...
         <Analyzer Include="$(SolutionDir)\RubberduckCodeAnalysis\bin\RubberduckCodeAnalysis.dll" />
   </ItemGroup>
-
+  <!-- Commented out due to large number of false positives. Hopefully the package
+	   will get annotations, which would enable us to exclude those false positives.
+	   Ref: https://github.com/DotNetAnalyzers/IDisposableAnalyzers/issues/126
+  <ItemGroup>
+    <PackageReference Include="IDisposableAnalyzers">
+	  <Version>2.0.5</Version>
+	</PackageReference>
```
Vogel612: Can that be just an `Analyzer` reference or does the `PackageReference` automatically add the analyzer?

Rubberduck.Core/UI/CodeExplorer/Commands/PrintCommand.cs
```diff
@@ -87,19 +87,23 @@ protected override void OnExecute(object parameter)
                 {
                     while (index < text.Count)
                     {
-                        var font = new Font(new FontFamily("Consolas"), 10, FontStyle.Regular);
-                        printPageArgs.Graphics.DrawString(text[index++], font, Brushes.Black, 0, offsetY, new StringFormat());
+                        using (var font = new Font(new FontFamily("Consolas"), 10, FontStyle.Regular))
```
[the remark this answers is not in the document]
bclothier: TBH I never ever would have thought that fonts, registry and process were all disposable. We weren't disposing those. XML, we should have known better but goes to show how many types are really disposable.

Rubberduck.Core/UI/Command/ExportAllCommand.cs
```diff
@@ -61,12 +61,27 @@ private bool Evaluate(IVBProject project)
...
+                case IVBProject vbproject:
+                    Export(vbproject);
+                    break;
+                default:
+                {
+                    using (var project = _vbe.ActiveVBProject)
+                    {
+                        Export(project);
+                    }
+                    break;
```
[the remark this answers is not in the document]
bclothier: Credits goes to @comintern

RubberduckBaseProject.csproj
bclothier: Adding the `PackageReference` automatically adds its to the analyzer somehow. I would have liked to add to the `Analyzer` to be more explicit but not sure I can do that without making it go weird.

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/CodePane.cs
bclothier: Hmm. I missed that (wasn't really focusing on that part. Will check and make sure to use Nlog there.

Rubberduck.VBEEditor/Events/VbeEvents.cs
bclothier: I agree - I'll double check that.

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeInfos.cs
bclothier: I'm sure those three items got flagged because the analyzer thought those would be potentially reassigned. Reassignment without disposing the prior value would be problematic. cf. [IDISP003](https://github.com/DotNetAnalyzers/IDisposableAnalyzers/blob/master/documentation/IDISP003.md)

Rubberduck.API/VBA/Parser.cs
```diff
@@ -235,6 +233,9 @@ public void Dispose()
             }
             
             _disposed = true;
+            _parser?.Dispose();
+            _vbe?.Dispose();
+            _tokenSource.Dispose();
```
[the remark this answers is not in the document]
bclothier: I thought that was what we were supposed to -- if we throw during the dispose, the pooch is totally screwed anyway? At least subsequent calls would not throw again (but we're still screwed).

Rubberduck.Core/UI/Command/ExportAllCommand.cs
```diff
@@ -61,12 +61,27 @@ private bool Evaluate(IVBProject project)
...
+            {
+                case CodeExplorerProjectViewModel projectNode when projectNode.Declaration.Project != null:
+                    Export(projectNode.Declaration.Project);
+                    break;
+                case IVBProject vbproject:
+                    Export(vbproject);
+                    break;
+                default:
+                {
+                    using (var project = _vbe.ActiveVBProject)
```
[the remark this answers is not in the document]
bclothier: By convention, all properties of SCW that returns another SCW is always creating a new SCW, so it's ours to dispose. That's also why we need the annotation with the IDisposable analzyer - that gets flagged as a false positive, even if it was a method instead. Besides, we need it to be a property to remain compatible with the original COM interface.

Rubberduck.VBEEditor/Events/VbeEvents.cs
comintern: That's definitely a reference copy. AFAICT this will break component registration.

bclothier: I think that was a false positive and I did not check this carefully; I've removed the unneeded `Dispose`.

PR #4509, opened 2018-11-08
Fix Release Upload

Vogel612:
Upload Github artifacts by name instead of path


appveyor.yml
```diff
@@ -95,39 +94,45 @@ for:
   environment:
     # fall back to calling the installer after the branch we're on
     installer_name: Rubberduck.Setup.%APPVEYOR_BUILD_VERSION%-%APPVEYOR_REPO_BRANCH%.exe
```
Vogel612: this line results in a failure to rename the artifact later. let's not do that

PR #4520, opened 2018-11-14
Fixes 4349...take two

### LINKED ISSUE #4349: Rename clash check is a little too tight.

BZngr:
Resolves #4349.  Removes Enums/UDTs rename conflict detection false-positive.
Added a failing repro test for last reported false-positive and some tweaks to the VBA input code for prior #4349 tests.

### DIFF
--- a/Rubberduck.Parsing/VBA/DeclarationCaching/DeclarationFinder.cs
+++ b/Rubberduck.Parsing/VBA/DeclarationCaching/DeclarationFinder.cs
@@ -1034,12 +1034,8 @@ namespace Rubberduck.Parsing.VBA.DeclarationCaching
 
             if (IsEnumOrUDTMemberDeclaration(renameTarget)) 
             {
-                var memberMatches = identifierMatches.Where(idm =>
+                return identifierMatches.Where(idm =>
                     IsEnumOrUDTMemberDeclaration(idm) && idm.ParentDeclaration == renameTarget.ParentDeclaration);
-                if (memberMatches.Any())
-                {
-                    return memberMatches;
-                }
             }
 
             identifierMatches = identifierMatches.Where(nc => !IsEnumOrUDTMemberDeclaration(nc));
--- a/RubberduckTests/Refactoring/RenameTests.cs
+++ b/RubberduckTests/Refactoring/RenameTests.cs
@@ -2384,10 +2384,11 @@ End Sub
             tdo.MsgBox.Verify(m => m.NotifyWarn(It.IsAny<string>(), It.IsAny<string>()), Times.Never);
         }
 
+        //Issue: https://github.com/rubberduck-vba/Rubberduck/issues/4349
         [Test]
         [Category("Refactorings")]
         [Category("Rename")]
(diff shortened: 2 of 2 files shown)


Rubberduck.Parsing/VBA/DeclarationCaching/DeclarationFinder.cs
```diff
@@ -1034,12 +1034,8 @@ public IEnumerable<Declaration> FindNewDeclarationNameConflicts(string newName,
 
             if (IsEnumOrUDTMemberDeclaration(renameTarget)) 
             {
-                var memberMatches = identifierMatches.Where(idm =>
+                return identifierMatches.Where(idm =>
```
bclothier: I'm unclear how this fixes the issue 4349. This seems to be just refactoring the code without any changes in the code's behavior?

RubberduckTests/Refactoring/RenameTests.cs
```diff
@@ -2445,7 +2445,39 @@ End Enum
...
+            };
+
+            PerformExpectedVersusActualRenameTests(tdo, inputOutput);
+            tdo.MsgBox.Verify(m => m.ConfirmYesNo(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<bool>()), Times.Never);
+        }
+
+        [Test]
+        [Category("Refactorings")]
+        [Category("Rename")]
+        public void RenameRefactoring_DoesNotWarnForMember_Issue4349()
```
bclothier: This is a question to @rubberduck-vba/contributors  -- do we have any consensus on whether the issue # should be embedded in the test? I get the benefit but I would normally put them in a comment rather than in the name.

PeterMTaylor: I vote keeping issues and their URL links in comments as I believe improves/strengthens our Test driven culture for RD with better evidence tracking for regression anaylsis if need be.
Are there any limitation of how long the URL be kept in comments similar to 255 characters if so, any shortening URL thoughts?

Vogel612:
Poking something at Duga to verify she's back up. Please disregard

RubberduckTests/Refactoring/RenameTests.cs
BZngr: I'm in agreement with using a hyperlink to the issue in a comment...and have updated the tests accordingly.

PR #4521, opened 2018-11-15
Name AppVeyor Tags even more aggressively

Vogel612:


appveyor.yml
```diff
@@ -107,7 +107,7 @@ artifacts:
     name: Rubberduck
 deploy:
   - provider: GitHub
-    tag: Rubberduck Stable v$(appveyor_build_version)
+    tag: RETAGMEWITHAMESSAGE
```
bclothier: i feel that this needs a bit explanation - is this a special code to AV to retag it using the commit message? or is it just a comment to us meatbags to retag it?

appveyor.yml
```diff
@@ -118,11 +118,11 @@ deploy:
     on: 
       branch: master
   - provider: GitHub
-    tag: Rubberduck v$(appveyor_build_version)
+    tag: Prerelease-v$(appveyor_build_version)
     release: $(release_name)
     draft: false
     prerelease: true
-    description: "AppVeyor build on [$(appveyor_repo_branch)] - https://ci.appveyor.com/project/rubberduck-vba/rubberduck/build/$(appveyor_build_version)<br/><br/>Merges #$(appveyor_pull_request_number) - $(appveyor_pull_request_title)"
+    description: "AppVeyor build on [$(appveyor_repo_branch)] - https://ci.appveyor.com/project/rubberduck-vba/rubberduck/build/$(appveyor_build_version)"
```
bclothier: Bummer this isn't working. Would have been useful to see what's what just browsing the releases.

appveyor.yml
Vogel612: The latter

appveyor.yml
Vogel612: Ya.that was the idea

PR #4526, opened 2018-11-17
Cull build warnings

comintern:
Most of this is implementing the proscribed Dispose pattern, although there are a couple of legit warnings that were fixed and a couple non-legit ones that were suppressed.


Rubberduck.Core/UI/DockableToolwindowPresenter.cs
```diff
@@ -108,27 +108,34 @@ private void EnsureMinimumWindowSize(IWindow window)
...
-
+           
             _isDisposed = true;
         }
 
 #if DEBUG
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1063")] // only logging here.
         ~DockableToolwindowPresenter()
         {
             // destructor for tracking purposes only - do not suppress unless
```
bclothier: FWIW the comment seem to contradict the current behavior. We are now using the finalizer to ensure disposal in addition to logging (e.g. tracking)

Rubberduck.Core/UI/CustomComWrappers/IOleWrapper.cs
```diff
@@ -298,14 +329,18 @@ public Wrapper_IOleInPlaceFrame(IntPtr hostObjectPtr) : base(hostObjectPtr)
             }
         }
 
-        public override void Dispose()
+        protected override void Dispose(bool disposing)
```
bclothier: Ok, this was funky but also, thar be big scary dragons here. Please test the changes with a native debugger attached and ensure no exceptions get thrown on exit. This is a part of the code responsible for preventing exceptions / AVs that happens long after the AppDomain shutdown.

Rubberduck.Core/UI/IOpenFileDialog.cs
```diff
@@ -108,7 +108,20 @@ public bool DereferenceLinks
 
         public void Dispose()
         {
+            Dispose(true);
```
bclothier: Was confused to see an implementation of `Dispose` in a file named `IOpenFileDialog` - turns out that the file has both `IOpenFileDialog` and `OpenFileDialog`. If we're going to keep both interface and implementation in one file, I think it's more logical to call it after the class, not the interface.

Rubberduck.Core/UI/ISaveFileDialog.cs
```diff
@@ -109,7 +109,20 @@ public bool DereferenceLinks
 
         public void Dispose()
         {
+            Dispose(true);
```
bclothier: Same comment as IOpenFileDialog

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeLibs.cs
```diff
@@ -534,5 +534,7 @@ public TypeLibWrapper Get(string searchLibName)
             }
             return retVal;
         }
+
+        protected override void Dispose(bool disposing) => base.Dispose(disposing);
```
bclothier: Why expression body here but not in other places?

RubberduckTests/Mocks/Windows.cs
```diff
@@ -110,8 +106,8 @@ public IWindow CreateWindow(string name)
             return result.Object;
         }
 
-        public void ReleaseDockableHosts()
-        {
-        }
+        public void ReleaseDockableHosts() { }
```
bclothier: Should we clean this up? Seems silly to have a no-op method....

Rubberduck.Core/UI/DockableToolwindowPresenter.cs
comintern: Not sure what I was thinking there... fixed.

Rubberduck.Core/UI/CustomComWrappers/IOleWrapper.cs
comintern: I'm getting a clean exit here with Excel.

Rubberduck.Core/UI/IOpenFileDialog.cs
comintern: Agreed, wasn't even looking at the file names when I was burning through here.  Might as well rename these since I'm touching them.

Rubberduck.VBEEditor/ComManagement/TypeLibs/TypeLibs.cs
comintern: I was trying to be as consistent as I could - I only used the expression body for the overrides that did nothing but call `base.Dispose`.

RubberduckTests/Mocks/Windows.cs
comintern: That's on the `IWindows` interface - it's only a nop here because this is a test implementation.

PR #4613, opened 2018-12-05
Adjust column widths

### LINKED ISSUE #3831: Test Explorer Header widths not uniform
Header widths of Grouping by `Location` and `Outcome` don't use the same widths, they are independent.
Found while working on https://github.com/rubberduck-vba/Rubberduck/issues/3590#issuecomment-371970017

machthree:
I've added some code to the code behind - hopefully there isn't a simple way to link the column widths using XAML alone 

Closes #3831

### DIFF
--- a/Rubberduck.Core/UI/UnitTesting/TestExplorerControl.xaml.cs
+++ b/Rubberduck.Core/UI/UnitTesting/TestExplorerControl.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Windows.Data;
 using System.Windows.Threading;
+using Rubberduck.UI.Controls;
 
 namespace Rubberduck.UI.UnitTesting
 {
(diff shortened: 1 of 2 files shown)


Rubberduck.Core/UI/UnitTesting/TestExplorerControl.xaml.cs
```diff
@@ -68,5 +69,31 @@ protected virtual void Dispose(bool disposing)
...
+                    // and then copy column widths from the old now hidden element on the subsequent invocation
+                    if (groupingGrid.IsVisible)
+                        _newGroupingGrid = groupingGrid;
+                    else
+                    {
+                        // copy column widths from previously visible grouping grid to ensure widths are consistent
+                        for (int columnIndex = 0; columnIndex < groupingGrid.Columns.Count; columnIndex++)
+                        {
+                            // copy column widths from current now hidden columns to the new visible columns
+                            _newGroupingGrid.Columns[columnIndex].Width = groupingGrid.Columns[columnIndex].Width;
```
MDoerner: This line throws an NRE on shutdown.

machthree: I've not seen an examples of this causing problems on shutdown although admittedly I've only been testing using Excel as a host and my tests have been rather simplistic. Can you suggest how to reproduce?

bclothier: I think to see this, you need to attach a debugger and try to shutdown Excel. The exception is likely being caught and handled (but in suboptimal manner).

machthree: Ignore that - just found how to recreate.

Issue #4662, opened 2018-12-19
Add/Remove references dialog box isn't wide enough


daFreeMan:
As a follow on to #4647 (and after installing build .4340) the dialog box still isn't wide enough. The button visibility has been fixed, but that pushed other, arguably less important, information out of the box in its default size:
![2018-12-19 08_58_48-rubberduck - add_remove references](https://user-images.githubusercontent.com/11889733/50225738-55f6c680-036f-11e9-8347-b879f8a42815.png)

I'm running 1900x1200 resolution so there's _plenty_ of room to make that dialogue wider, though I realize that not everyone runs resolution that high (and some are higher).

Would it be possible for the two list boxen to dynamically size themselves to the necessary width, at least until the whole dialogue box reachs some sort of "reasonable" maximum size, say 90% of the VBE window's width? For those with the monitor size, there's no reason not to use it.

Yes, this info is available for each library by selecting it, but if we're going to the trouble of displaying it in the list box, may as well make it visible.

comintern:
Agreed that information shouldn't be hidden on the right hand side. It needs its scrollbars back. Dynamically resizing the form itself would be a little weird (it would also kick the scrollbar out from under the mouse). I think I'd find that behavior surprising as hell.

Definitely needs to persist the window size though.

PR #4673, opened 2018-12-26
Add HorizontalScrollBar Closes rubberduck-vba/Rubberduck#4662

ShimonLeb:
Closes rubberduck-vba/Rubberduck#4662  I changed AddRemoveReferencesWindow HorizontalScrollBarVisibility to Auto

### DIFF
--- a/Rubberduck.Core/UI/AddRemoveReferences/AddRemoveReferencesWindow.xaml
+++ b/Rubberduck.Core/UI/AddRemoveReferences/AddRemoveReferencesWindow.xaml
@@ -232,7 +232,7 @@
                                       SelectionMode="Single"
                                       ItemsSource="{Binding AvailableReferences}" 
                                       HorizontalContentAlignment="Stretch"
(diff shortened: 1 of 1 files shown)


Vogel612:
Correct me if I'm wrong: this patch does not affect the partial issue raised with persisting the dimensions of a manually resized dialog, correct? 

I'm just going to go on a limb and suggest that RD generally doesn't keep track of the dimensions of manually resized dialogs. (Not windows, but dialogs). If that's the case: LGTM. If not we need a follow-up issue to address that bug down the line.

ShimonLeb:
I checked the Rename Settings & Regex Assistant dialogs and its not persisting the dimensions

comintern:
@ShimonLeb Thanks for the PR - I let this one get buried on my todo list. The only RD windows that persist position or size are the ones that are wrapped in VBE tool windows, and that's only because the VBE stores their information in its registry key.

Issue #4533, opened 2018-11-22
File management in RD


bclothier:
There are few tech debts we have surrounding file management in RD.

1. We had copy'n'pasted method for building the root path to the `%APPDATA%\Rubberduck\*` folders for various things. With PR #4225 a `IPersistancePathProvider` was introduced to abstract the root path and thus make it easy for various subsystems to get their own folder for persisting their data.

2. We have TEMP_PATH constant scattered around and the files are not controlled in any manner. Ideally, there should be a method that allows a `using (var tempFile = MakeSomeTempFile(source))` block to explicitly scope the lifetime of a temporary file, while allowing the calling code to perform its work with the file.

3. We are not currently using `System.IO.Abstractions`; some subsystems has a direct dependency on the `System.IO`, hampering its testability, while other has their handrolled interfaces that wraps the IO operations. We need to make this consistent, as to keep `System.IO` out of those subsystems.

In the end, we need to ensure that we have an uniform system for accessing files (both permanent and temporary) that do not harm the testability and is predictable (e.g. doesn't write a file to a random location and then forget about it).

MDoerner:
Note that PR #4444 will introduce a file handling interface.

bclothier:
Ref #4794 - that more or less addresses the point 3.

PR #4699, opened 2019-01-10
Clean up the persistence services

bclothier:
This is a follow-on to the recently merged PR #4697, performing some additional cleanup in hopes of improving the readability of the classes. We also now use persistence path provider to abstract out the filesystem.

Some more unit tests is likely needed to complete the coverage on the settings which is inadequate. 

Ref #4533


Rubberduck.Main/Extension.cs
```diff
@@ -145,7 +145,7 @@ private void InitializeAddIn()
                     return;
                 }
 
-                var configLoader = new XmlPersistanceService<GeneralSettings>
+                var configLoader = new XmlPersistanceService<GeneralSettings>()
```
comintern: Redundant parens?

Rubberduck.SettingsProvider/XmlPersistanceServiceBase.cs
```diff
@@ -41,11 +47,16 @@ public virtual string FilePath
 
         protected static T FailedLoadReturnValue()
         {
-            return (T)Convert.ChangeType(null, typeof(T));
+            return new T();
```
comintern: I believe this has to return null to ensure that the defaults get loaded.

Rubberduck.Main/Extension.cs
bclothier: That was from my aborted attempt to ctor-inject which caused more problems than solutions. Will remove.

bclothier:
The PR now dries the settings view model some more, providing a factory to abstract out the creation of the viewmodel in a generic manner and also remove several redundant code in the setting view models. Furthermore, the persistence service now properly depend on the `IPersistancePathProvider`, abstracting out the filesystem some more.

Rubberduck.SettingsProvider/XmlPersistanceService.cs
```diff
@@ -7,29 +7,27 @@
...
-            var doc = GetConfigurationDoc(FilePath);
+        public override T Load(T toDeserialize, string nonDefaultFilePath = null)
+        {
+            var filePath = string.IsNullOrWhiteSpace(nonDefaultFilePath) ? FilePath : nonDefaultFilePath;
+            var doc = GetConfigurationDoc(filePath);
             var node = GetNodeByName(doc, typeof(T).Name);
             if (node == null)
             {
-                return FailedLoadReturnValue();
+                return Cached;
```
MDoerner: This does not look as if it is doing the same as before the refactoring. Previously, the cached value was used if it exists and no file read was performed in that case. Now, the cached value is used only if the load from file fails.

Moreover, using the cached value without testing that something got cached in the first place is counterintuitive.

comintern: The file read actually occurs in the `GetConfigurationDoc` call. This change does change the behavior, but it returns the cached settings object instead of deferring to the caller to obtain the defaults. The term "`Cached`" makes this confusing though, in that `Cached` will only be non-null if there had never been a successful read. If nothing has been cached, returning null is correct (and `Cached` would be null in that case). This covers the scenario where the settings file is deleted or broken at run-time, and changes the behavior to use the last successful load instead of the defaults in that case.

Rubberduck.Core/UI/Settings/GeneralSettingsViewModel.cs
```diff
@@ -20,36 +20,48 @@ public enum DelimiterOptions
...
-            });
-
-            LogLevels = new ObservableCollection<MinimumLogLevel>(LogLevelHelper.LogLevels.Select(l => new MinimumLogLevel(l.Ordinal, l.Name)));
+                new[]
+                {
+                    new DisplayLanguageSetting("en-US"),
+                    new DisplayLanguageSetting("fr-CA"),
+                    new DisplayLanguageSetting("de-DE"),
+                    new DisplayLanguageSetting("cs-CZ")
+                });
```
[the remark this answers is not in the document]
bclothier: Agreed - that should be a language service class or something. Which should be its own PR, IMO.

Rubberduck.Core/UI/Settings/SettingsForm.cs
```diff
@@ -45,42 +46,43 @@ public SettingsForm()
...
                      */
-                    Control = new GeneralSettings(new GeneralSettingsViewModel(config, operatingSystem, messageBox, vbeSettings, new List<Type>())),
+                    //Control = new GeneralSettings(new GeneralSettingsViewModel(config, operatingSystem, messageBox, vbeSettings, new List<Type>())),
+                    Control = new GeneralSettings(viewModelFactory.Create<Rubberduck.Settings.GeneralSettings>(config)),
                     View = SettingsViews.GeneralSettings
                 },
                 new SettingsView
                 {
-                    Control = new TodoSettings(new TodoSettingsViewModel(config)),
+                    Control = new TodoSettings(viewModelFactory.Create<ToDoListSettings>(config)),
```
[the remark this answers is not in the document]
bclothier: Yes it could be made into its own factory, similar to what I did with the viewmodels. However, that would broaden the scope of the PR. I want to do that in its own PR, too and preferably after #4072 is merged, since the changes should be more universal than just the refactoring dialogs (e.g. abstraction of the WinForm out of the picture, having a more uniform system of activating a dialog)

Rubberduck.Core/UI/Settings/GeneralSettingsViewModel.cs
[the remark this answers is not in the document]
Vogel612: yea, that one is definitely out of scope for the current PR

Rubberduck.SettingsProvider/XmlPersistanceService.cs
comintern: Max might have a different thought, but maybe a comment?  I know exactly what's going on because I initially wrote the code and recently discussed it on chat, but the future me only wrote the code...

Issue #4666, opened 2018-12-20
Data binding leak in the InspectionResultsControl


MDoerner:
In the grouping grid in the `InspectionResultsControl` we bind to the `Result` collection on the  view model, which is an `ObservableCollection<IInspectionResult>`, and base the grouping on the `Inspection` property of the items. 
Since `IInspectionResult` does not extend `INotifyPropertyChanged` this leaks all inspection results.

The cleanest way to fix this is probably to introduce an `InspectionResultViewModel` that implements `INotifyPropertyChanged` to wrap the results.

MDoerner:
As an alternative to the additional view model, we could change the binding mode to `OneTime`, which makes the UI use a snapshot and avoids registration to the value changed event.

comintern:
Of those options, it seems like the more reasonable one (at least in the short term) would be to switch to `OneTime` binding.  I'd assumed that was what we were doing but that's apparently not the case. I can't really think of a situation where an inspection result would "change" with our current inspection architecture unless the entire `ObservableCollection` was being replaced.

If we ever do change the overall architecture to allow more dynamic inspection, that would be the time to add an `InspectionResultViewModel`. I think this would just warrant a comment in the xaml as to why the binding is `OneTime`.

comintern:
This should be closed with #4717

PR #4661, opened 2018-12-19
Code Explorer Extreme Makeover (for some values of extreme)

comintern:
This PR addresses a bunch of current CE bugs and usability issues, leveraging the new references work.

Highlights:
- Sort now filters correctly (and uses the same visual style for the search bar as the references dialog. Closes #3707 and #4478
- Sort is now case insensitive. Closes #3860 
- Cleaned up the toolbar. Closes #4638 
- Updates expanded and selected state correctly on reparse.
- Refactor commands so that EvaluateCanExecute works right. Closes #4654 
- Remove command now prompts appropriately. Closes #3561 
- Added sync button for code pane selection. Will close #2344 when no longer WIP
- General cleanup and tech debt payments.

**Note** 
I intend to reorganize the `CodeExplorerViewModel` to make it easier to navigate (it was kind of a hot mess), but I'm waiting until reviews are done to make it easier to see changes.

PR #4713, opened 2019-01-16
Performance overhaul of Inspection Results window

### LINKED ISSUE #4598: Code Inspector Filter on refresh
If the code inspector window is filtered and you refresh, the filter will still be active but ignored.

You need to double click the filter to make it active again.

comintern:
This is primarily aimed at getting rid of the unnecessary churn in the `ObservableCollection`(s), and to improve filtering and grouping responsiveness in the results. Note that this is only tagged WIP because it is based from #4661 - that PR should be merged first when all the reviews are completed and addressed to avoid pulling in those commits. I also rolled in some UI tweaks while I was at it. The new toolbar looks like this:

screenshot from 2019-01-15 22-55-33

**Highlights:**
 - Now binding an `ICollectionView` directly instead of implicitly creating the view. I suspect this was leaking memory (and it simplifies ...well... everything).
 - The default grouping was not reachable from the UI after it was changed. This is fixed.
 - Adds 2 new groupings for inspection name and severity.
 - Now using `ToggleButton`s instead of drop-down menus to improve visual feedback.
 - Removed a couple backgrounds to fix rendering glitches.

BTW, this also closes #4598, and addresses #4666 (love that number given the issue...). It possibly closes some other open issues, but I haven't finished a comprehensive search yet.

**Note:** There's still some clean-up to be done here, and I'm also still investigating some issues surrounding the responsiveness of the `DataGrid` control, but if those don't sync in by the time this is merged, they're on the agenda.


rkapka:
Now this will be a nitpick of all nitpicks, but I suggest making the toolbar group names uniform. Either `Group`+`Filter` or `Grouping`+`Filtering`.

comintern:
Closing due to rebase.

PR #4717, opened 2019-01-17
Inspection Result window performance overhaul (Round 1)

### LINKED ISSUE #4598: Code Inspector Filter on refresh

comintern:
Re-opening of #4713

This is primarily aimed at getting rid of the unnecessary churn in the `ObservableCollection`(s), and to improve filtering and grouping responsiveness in the results. Note that this is only tagged WIP because it is based from #4661 - that PR should be merged first when all the reviews are completed and addressed to avoid pulling in those commits. I also rolled in some UI tweaks while I was at it. The new toolbar looks like this:

51227204-e09ff680-1918-11e9-8cc4-bb3b6588b437

**Highlights:**

- Now binding an `ICollectionView` directly instead of implicitly creating the view. I suspect this was leaking memory (and it simplifies ...well... everything).
- The default grouping was not reachable from the UI after it was changed. This is fixed.
- Adds 2 new groupings for inspection name and severity.
- Now using `ToggleButton`s instead of drop-down menus to improve visual feedback.
- Removed a couple backgrounds to fix rendering glitches.

BTW, this also closes #4598, and addresses #4666 (love that number given the issue...). It possibly closes some other open issues, but I haven't finished a comprehensive search yet.

**Note:** There's still some clean-up to be done here, and I'm also still investigating some issues surrounding the responsiveness of the `DataGrid` control, but if those don't sync in by the time this is merged, they're on the agenda.


Rubberduck.Core/UI/Inspections/InspectionResultsControl.xaml
```diff
@@ -27,23 +27,16 @@
...
             <BitmapImage x:Key="SettingsImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Fugue/gear.png" />
             <BitmapImage x:Key="RefreshImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Fugue/arrow-circle-double.png" />
             <BitmapImage x:Key="FixImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Fugue/tick.png" />
-            <BitmapImage x:Key="GroupByImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Custom/PNG/GroupBy.png" />
+
+            <BitmapImage x:Key="GroupByInspectionTypeImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Fugue/block.png" />
+            <BitmapImage x:Key="GroupByInspectionImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Fugue/light-bulb-code.png" />
+            <BitmapImage x:Key="GroupByLocationImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Custom/PNG/ObjectClass.png" />
+            <BitmapImage x:Key="GroupBySeverityImage" UriSource="pack://application:,,,/Rubberduck.Resources;component/Icons/Custom/PNG/Severity.png" />
```
Vogel612: considering that the InspectionFilterToBooleanConverters are only scoped to the ToolBar where they're used, would it make sense to move the GroupBy and FilterBy images into that `Toolbar.Resources` section as well?

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
```diff
@@ -22,6 +24,25 @@
 
 namespace Rubberduck.UI.Inspections
 {
+    [Flags]
+    public enum InspectionResultsFilter
+    {
+        None = 0,
+        Hint = 1,
```
Vogel612: the pattern would dictate this to be `1 << 1` which is obviously nonsense, because that's what `Suggestion` should be. There's a gap here.

Rubberduck.Resources/Rubberduck.Resources.csproj
```diff
@@ -12,6 +12,9 @@
     <Resource Include="**\*.bmp" Exclude="$(IntermediateOutputPath)\**" />
     <Resource Include="**\*.txt" Exclude="$(IntermediateOutputPath)\**" />
   </ItemGroup>
+  <ItemGroup>
+    <Resource Remove="Resources\light-bulb.png" />
+  </ItemGroup>
```
Vogel612: huh? what's with that? Why is the file added, but excluded here?

Rubberduck.Core/UI/Inspections/InspectionResultsControl.xaml
comintern: Maybe.  I left them where they were so all the images would be in the same place more than any other reason.  The converters are in the `Toolbar.Resources` so I could see them when I was writing the xaml.

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
comintern: That always throws me (I generally use `2 ^ 0`, `2 ^ 1`, `2 ^ 2`, ...) but I used what seems to be RD convention and face-planted it...  Will fix.

Rubberduck.Resources/Rubberduck.Resources.csproj
comintern: Grrr... I meant to remove that - I went with the existing one.  Will fix.

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
```diff
@@ -202,136 +209,151 @@ public IEnumerable<DisplayQuickFix> QuickFixes
...
+        private static readonly Dictionary<InspectionResultGrouping, PropertyGroupDescription> GroupDescriptions = new Dictionary<InspectionResultGrouping, PropertyGroupDescription>
         {
-            get => _groupByInspectionType;
-            set
-            {
-                if (_groupByInspectionType == value) { return; }
+            { InspectionResultGrouping.Type, new PropertyGroupDescription("Inspection", new InspectionTypeConverter()) },
+            { InspectionResultGrouping.Name, new PropertyGroupDescription("Inspection.Name") },
+            { InspectionResultGrouping.Location, new PropertyGroupDescription("QualifiedSelection.QualifiedName") },
+            { InspectionResultGrouping.Severity, new PropertyGroupDescription("Inspection.Severity") }
```
bclothier: Do they have to be literals? Const? Resource?

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
```diff
@@ -202,136 +209,151 @@ public IEnumerable<DisplayQuickFix> QuickFixes
...
 
-                _groupByLocation = value;
-                OnPropertyChanged();
+                _groupByInspection = value;
+                if (_groupByInspection)
+                {
+                    _groupByName = false;
+                    _groupByLocation = false;
+                    _groupBySeverity = false;
+                    SetGrouping(InspectionResultGrouping.Type);
```
bclothier: It seems to me that we can further cut down the amount of code and eliminate human errors by using a bitmask rather than 4 booleans. That would make it more consistent with filters which already uses a bitmask, as well.

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
comintern: `PropertyGroupDescription` takes a string for the first argument (it's the equivalent to the xaml `Path` on a binding). It seems silly to make a second lookup set with the same keys as this one.

Rubberduck.Core/UI/Inspections/InspectionResultsViewModel.cs
comintern: Agreed.  That's on the agenda for the next PR on this.  I just skipped it for the time being because they're mutually exclusive, and that means making a switch in the converter.

comintern: This is in the batch of commits I just pushed.

comintern:
Barring review comments (and assuming a successful build...), that's everything I intended to go into this one. Note that I had to disable group virtualization to handle the expand and collapse all buttons, but when I did the second round of performance testing it turned out to be "acceptable" for the ~3000 inspection result project I tested on. Not _instantaneous_ grouping changes, but sub 1 second. If this turns into a problem with enormous projects (10K+ results),  we can revisit that decision.

PR #4727, opened 2019-01-20
Code Explorer hotfix 3

comintern:
CodeExplorerTests.Count += 253

This also fixes (or greatly mitigates) an unrelated issue I stumbled across when I was live testing in Excel - there's apparently a race condition between the parser state and the disabling of the refresh button in response to the `EvaluateCanExecute` binding on the `ReparseCommand`. You could occasionally get it to execute while the parser was busy by mashing repeatedly on the refresh button in the CE. It's possible this is also an issue with other UI linked commands that check the parser state in their `EvaluateCanExecute` function.

More tests are coming, but this is can be merged.


Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerMemberViewModel.cs
```diff
@@ -9,9 +9,9 @@ namespace Rubberduck.Navigation.CodeExplorer
 {
     public class CodeExplorerMemberViewModel : CodeExplorerItemViewModel
     {
-        public CodeExplorerMemberViewModel(ICodeExplorerNode parent, Declaration declaration, IEnumerable<Declaration> declarations) : base(parent, declaration)
+        public CodeExplorerMemberViewModel(ICodeExplorerNode parent, Declaration declaration, List<Declaration> declarations) : base(parent, declaration)
```
bclothier: Does it have to be `List`? Can't it be `IList`?

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerMemberViewModel.cs
```diff
@@ -76,10 +76,14 @@ protected sealed override void AddNewChildren(List<Declaration> updated)
         {
             if (updated != null)
             {
-                AddChildren(updated
-                    .Where(item =>
-                        SubMemberTypes.Contains(item.DeclarationType) && item.ParentDeclaration.Equals(Declaration))
-                    .Select(item => new CodeExplorerSubMemberViewModel(this, item)));
+                var updates = updated.Where(item => SubMemberTypes.Contains(item.DeclarationType) && item.ParentDeclaration.Equals(Declaration)).ToList();
```
bclothier: I think the line is too wide and should be broke up.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
```diff
@@ -22,14 +22,14 @@ public class CodeExplorerProjectViewModel : CodeExplorerItemViewModel
 
         private readonly IVBE _vbe;
 
-        public CodeExplorerProjectViewModel(Declaration declaration, IEnumerable<Declaration> declarations, RubberduckParserState state, IVBE vbe, bool references = true) : base(null, declaration)
+        public CodeExplorerProjectViewModel(Declaration project, List<Declaration> declarations, RubberduckParserState state, IVBE vbe, bool references = true) : base(null, project)
```
bclothier: Same question RE: `List` vs. `IList`

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerViewModel.cs
```diff
@@ -237,34 +237,33 @@ private void HandleStateChanged(object sender, ParserStateEventArgs e)
...
             {
-                return;
+                Synchronize(_state.DeclarationFinder.AllUserDeclarations);
             }
-
-            Synchronize(_state.DeclarationFinder.AllUserDeclarations.ToList());
         }
 
-        private void Synchronize(List<Declaration> declarations)
+        private void Synchronize(IEnumerable<Declaration> declarations)
```
bclothier: Why change from `List` to `IEnumerable` here when you did the opposite earlier? Considering that it's a private procedure, it should be OK to use concretes?

RubberduckTests/CodeExplorer/CodeExplorerComparerTests.cs
```diff
@@ -0,0 +1,357 @@
...
+                var fieldNode = module.Children.Single(s => s.Name == "Bar");
+                var propertyGetNode = module.Children.Single(s => s.Name == "Foo (Get)");
+
+                Assert.AreEqual(-1, new CompareByDeclarationType().Compare(fieldNode, propertyGetNode));
+            }
+        }
+
+        [Category("Code Explorer")]
+        [Test]
+        public void CompareByType_ReturnsPropertyGetEqualToPropertyLet()
```
bclothier: This set of tests could have been done as a data-driven tests. Something like:

```[TestCase(0, "Foo (Get)", "Foo (Let)")]
[TestCase, 0, "Foo (Get)", "Foo (Set)")]
...
public void CompareByType_ReturnsPropertyEqual(int expected, string name1, string name2)
```

That would significantly cut down on the amount of copy'n'pasting and more importantly make the tests more maintainable.

RubberduckTests/CodeExplorer/CodeExplorerComponentViewModelTests.cs
```diff
@@ -0,0 +1,332 @@
...
+        [TestCase(CodeExplorerTestSetup.TestUserFormName)]
+        public void Constructor_NameWithSignatureIsSet(string name)
+        {
+            var declarations = CodeExplorerTestSetup.TestProjectOneDeclarations.TestComponentDeclarations(name);
+            var componentDeclaration = declarations
+                .First(declaration => declaration.DeclarationType.HasFlag(DeclarationType.Module) && declaration.IdentifierName.Equals(name));
+
+            var component = new CodeExplorerComponentViewModel(null, componentDeclaration, declarations, null);
+
+            Assert.IsFalse(string.IsNullOrEmpty(component.NameWithSignature));
```
bclothier: Do we have tests testing the content of the signature?

RubberduckTests/CodeExplorer/CodeExplorerComponentViewModelTests.cs
```diff
@@ -0,0 +1,332 @@
...
+        [TestCase(CodeExplorerTestSetup.TestUserFormName)]
+        public void Constructor_ToolTipIsSet(string name)
+        {
+            var declarations = CodeExplorerTestSetup.TestProjectOneDeclarations.TestComponentDeclarations(name);
+            var componentDeclaration = declarations
+                .First(declaration => declaration.DeclarationType.HasFlag(DeclarationType.Module) && declaration.IdentifierName.Equals(name));
+
+            var component = new CodeExplorerComponentViewModel(null, componentDeclaration, declarations, null);
+
+            Assert.IsFalse(string.IsNullOrEmpty(component.ToolTip));
```
bclothier: Same question RE: contents of the tool tip

RubberduckTests/CodeExplorer/CodeExplorerComponentViewModelTests.cs
```diff
@@ -0,0 +1,332 @@
...
+
+            var component = new CodeExplorerComponentViewModel(null, componentDeclaration, new List<Declaration> { componentDeclaration }, null);
+
+            var nonMatching = testCharacters.ToCharArray().Except(name.ToLowerInvariant().ToCharArray());
+
+            foreach (var character in nonMatching.Select(letter => letter.ToString()))
+            {
+                component.Filter = character;
+                Assert.IsTrue(component.Filtered);
+            }
```
bclothier: Ought there be an `Assert` for the matching? (I'm understanding the test to contain only one matching result for each iteration).

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerMemberViewModel.cs
comintern: Maybe? Changing that to use the interface would require changing it everywhere else in the object hierarchy (including on the interface) though, so I'm not horribly inclined to change it.  Given that it's fairly standard to call `ToList()` on an `IEnumerable` to materialize it, using the strong type makes a certain amount of sense.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerMemberViewModel.cs
comintern: Done. This is what happens after R# crashes. _Curse you Visual Studio!!!_

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerViewModel.cs
comintern: It allows the `IEnumerable` to be materialized on the UI thread instead of in the event handler.

RubberduckTests/CodeExplorer/CodeExplorerComparerTests.cs
comintern: Agreed. The old UTs could probably still use some more clean-up. This is low priority though, so I'll flag it with a TODO in case I forget to circle back to it.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerViewModel.cs
bclothier: Seems that merits a comment so a clown isn't tempted to change it in future?

comintern: Done.

Issue #4743, opened 2019-01-24
Parse after Compact & Repair leaves empty CE


daFreeMan:
**Rubberduck version information**
Version 2.3.1.4460
OS: Microsoft Windows NT 10.0.15063.0, x64
Host Product: Microsoft Office 2016 x64
Host Version: 16.0.4711.1000
Host Executable: MSACCESS.EXE

**Description**
CE pane is empty after an automatic parse on a Compact & Repair.

**To Reproduce**
1. Open Access
2. Open VBE & parse the project
2. Do C&R
3. RD automatically parses the project when it reopens the DB
4. The CE is completely blank in the tree view area - it doesn't even show the "RD doesn't see anything yet" message.
5. A manual parse gives me a populated CE, as expected

**Expected behavior**

**Screenshots**
![2019-01-24 08_53_32-microsoft visual basic for applications - roireport - genquarterly code](https://user-images.githubusercontent.com/11889733/51682969-be604580-1fb6-11e9-9177-6a516b4fedb5.png)

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2791990/RubberduckLog.txt)

**Additional context**
This is reproducible - I closed & opened Access then repeated the steps & got the exact same thing.
Possibly related to #3787

bclothier:
Further observations....

It looks like it's actually *toggling*.

For example, if we open file and do *not* parse it, C & R it, it will get parsed after the C & R. C & R again, then it is no longer loaded. C & R yet again, then it's parsed.

daFreeMan:
I'm seeing the toggle behavior too. As I did a 2nd C&R on the opened & parsed DB, I noticed that the ducks spun briefly on the CE, then went away. There was a pause - long enough for me to think I was seeing different behavior - then the ducks started spinning again and the CE was populated.

As a side note, the CI _is_ populated, even when the CE isn't.

comintern:
Just to capture some of the findings from chat on this, I was able to reproduce this with a build that did not incorporate the recent CE changes. When I tested this with a random Access project, the C & R closed and reopened the VBE, and on the reopen it displayed the empty UI refresh control.

My suspicion is that Access unloads the project to perform the C & R, and checks the state of the VBE to see if it needs to reopen it when it's done. During this process, it either catches RD in a state where it isn't connected to the event sinks *or* the VBE doesn't fire the event at all. 

[...]

PR #4748, opened 2019-01-26
Show empty UI message in code explorer if no project nodes are found.

comintern:
Ref #4743


MDoerner:
Is this really a good idea? This will be confusing if there really is no project. This can happen in Access.

Issue #4405, opened 2018-10-03
Bang notation appears to be broken in the SLL parse


comintern:
I decided to track down some of the output spam from the parser:

```
[10/2/2018 8:56:39 PM Warning] line 4:9 extraneous input 'SomeIdentifier' expecting {'=', WS, LINE_CONTINUATION}
[10/2/2018 8:56:39 PM Warning] line 4:9 extraneous input 'SomeIdentifier' expecting {'=', WS, LINE_CONTINUATION}
```
It turns out to be this code in the `MemberNotOnInterface_DoesNotReturnResult_BangNotation` test:
```
Sub Foo()
    Dim dict As Dictionary
    Set dict = New Dictionary
    dict!SomeIdentifier = 42
End Sub
```
Parsing that code in the VBE with logging results in this:
```
2018-10-03 18:46:05.5888;WARN-2.2.6849.38216;Rubberduck.Parsing.VBA.Parsing.TokenStreamParserBase;SLL mode failed while parsing the CodePaneCode version of module Sheet1 at symbol SomeIdentifier at L4C10. Retrying using LL.;
2018-10-03 18:46:05.5888;DEBUG-2.2.6849.38216;Rubberduck.Parsing.VBA.Parsing.TokenStreamParserBase;Rubberduck.Parsing.Symbols.ParsingExceptions.MainParseSyntaxErrorException: extraneous input 'SomeIdentifier' expecting {'=', WS, LINE_CONTINUATION}
   at Rubberduck.Parsing.Symbols.ParsingExceptions.MainParseExceptionErrorListener.SyntaxError(IRecognizer recognizer, IToken offendingSymbol, Int32 line, Int32 charPositionInLine, String msg, RecognitionException e) in C:\Rubberduck\Rubberduck.Parsing\Symbols\ParsingExceptions\MainParseExceptionErrorListener.cs:line 17
   at Antlr4.Runtime.ProxyErrorListener`1.SyntaxError(IRecognizer recognizer, Symbol offendingSymbol, Int32 line, Int32 charPositionInLine, String msg, RecognitionException e)
   at Antlr4.Runtime.Parser.NotifyErrorListeners(IToken offendingToken, String msg, RecognitionException e)
   at Antlr4.Runtime.DefaultErrorStrategy.SingleTokenDeletion(Parser recognizer)
   at Antlr4.Runtime.DefaultErrorStrategy.Sync(Parser recognizer)
[...]

PR #4760, opened 2019-01-30
Fix bang notation in the SLL mode for the most common case

MDoerner:
This PR tries to address the problem that the SLL parser fails to properly resolve bang notation, as noted in issue #4405.

The basic problem is that the SLL parser is not context-aware. Consequently, each subrule matches as much as it can. Unfortunately, an exclamation mark is the type hint for the type `Single`. So, the identifer rule descends into the `typedIdentifier` branch and consumes the exclamation mark still required by the surrounding rule. 

To fix this, this PR introduces a sematic predicate in the `identifier` rule that deactivates the `typedIdentifier` alternative whenever there is an IDENTIFIER token two tokens further down. (A typehint cannot immediately be followed by an identifier.)

Unfortunately, the RHS of the bang notation is an `unrestrictedIdentifier`, which can be more than just an `IDENTIFER`; it allows any keyword and foreign identifiers. Moreover, the LHS identifer can also be a foreign name, in which case it can consit of an arbitrary number of tokens. These edge-cases are not covered by the fix. (Note that predicates must be at the start of an alternative.)


MDoerner:
Now, bang notation with a foreign name on the RHS is covered as well.

Rubberduck.Parsing/Grammar/VBAParser.g4
```diff
@@ -579,7 +579,8 @@ subscript : (expression whiteSpace TO whiteSpace)? expression;
 
 unrestrictedIdentifier : identifier | statementKeyword | markerKeyword;
 legalLabelIdentifier : { !(new[]{DOEVENTS,END,CLOSE,ELSE,LOOP,NEXT,RANDOMIZE,REM,RESUME,RETURN,STOP,WEND}).Contains(_input.La(1))}? identifier | markerKeyword;
-identifier : typedIdentifier | untypedIdentifier;
+identifier : {_input.La(3) != IDENTIFIER && _input.La(3) != L_SQUARE_BRACKET || _input.La(1) == L_SQUARE_BRACKET}? typedIdentifier
```
bclothier: I think this merits a comment explaining the logic because in first part, we look at the third token(?) but in the 2nd part, we look at the first token.

RubberduckTests/Grammar/VBAParserTests.cs
```diff
@@ -2953,6 +2953,50 @@ End Sub
...
+        {
+            const string code = @"
+Sub Test()   
+Dim dict As Scripting.Dictionary
+
+Dim x
+x = dict![a]
+End Sub
+";
+            var parseResult = Parse(code);
```
bclothier: Should we also test that the `a` is correctly parsed as an identifier? That may be more important for other cases where we have a expression like `dic![[Test]![Run]]`; the `[Test]![Run]` in this case would be an identifier, not a piece of code.

Rubberduck.Parsing/Grammar/VBAParser.g4
MDoerner: Males sense; I will add a comment explaining the logic.

RubberduckTests/Grammar/VBAParserTests.cs
MDoerner: I think that would be a different test and actually be of a different nature than the other tests. None of those test specific parts of the code, just that it parses and that some elements appear the correct number of times. Here, we could test the correct number of `unrestrictedIdentifier`.

BTW, we ne er parse `a` as an identifier: the `unrestrictedIdentifier` in question is `[a]`.

bclothier: That is fine by me - I just want to be sure we were covering the supported edge cases correctly. One more thought - do we need a separate test to assert the count of `unrestrictedIdentifer` for a chained bangs expression like `Forms![Main]![SomeControl]`/`Forms!Main!SomeControl` ?

MDoerner: I will add a test for chained bang notation without brackets and an ignored one with brackets. The version with brackets on the left cannot work the way we do this right now.

Rubberduck.Parsing/Grammar/VBAParser.g4
bclothier: Typo corrections...

```
//The predicate in the following rule has been introduced to lessen the problem that VBA uses the same characters used as type hints in other syntactical constructs, 
//e.g. in the bang notation (see withDictionaryAccessExpr). Generally, it is not legal to have an identifier with a opening bracket follow immediately after a type hint.
//The first part of the predicate tries to exclude these two situations. Unfortunately, predicates have to be at the start of a rule. S, an assumption about the number 
//of tokens in the identifier is made. All untypedIdentifers not a foreignNames consist of exactly one token and a typedIdentifier is an untypes one followed by a typeHint,
//again a single token. So, in the majority of situations, the third token is the token following the potential type hint. 
//For foreignNames, no assumption can be made because they consist of a pair of brackets containing arbitrarily many tokens. 
//That is why the second part of the predicate looks at the first character in order to determine whether the identifier is a foreignName.
```

Issue #4704, opened 2019-01-10
Enable running tests by location


bclothier:
In addition to category and selection, it should be possible (and hopefully trivial) to run tests based on their location (e.g. module they are in). 

_Originally posted by @bclothier in https://github.com/rubberduck-vba/Rubberduck/issues/4698#issuecomment-453072604_

Vogel612:
This is somewhat abstractable to running tests based on their current grouping in the grouping grid. As such we'd need a "Run Tests in Selected Group", where group would be any single of:

 - Location
 - Outcome
 - Category
 - ...?

Irubataru:
I think that is what I suggested when I talked about it that it would be more versatile if you could run issues in the selected group. In that case it could also replace "Run selected category" (and invalidate #4703) as that functionality would be there if you chose to group by category.

comintern:
The problem inherent in this UI is that it kind of conflicts with "Run selected tests" if we're allowing multi-selection in the test grid (see #4701). If the user Ctrl selects multiple row in different groupings, then what exactly is the "Selected Group"? Both of them?

The menu on the toolbar currently gives the following options:

![screenshot from 2019-02-02 12-54-46](https://user-images.githubusercontent.com/8944005/52167999-d5571400-26e9-11e9-9a78-c049daf0a97d.png)

Compare this to the options available in the VS "Run..." menu:

![screenshot from 2019-02-02 12-56-35](https://user-images.githubusercontent.com/8944005/52168023-10594780-26ea-11e9-89ad-5488ea35ee6b.png)

The main difference is that VS _also_ has context menus in its TreeView for running a selected node, whether or not it has sub-nodes. If it does have sub-nodes, it runs all of them. 

The compromise here might be to just remove the "Selected Category" option from the toolbar completely (ref #4703) and change the "Selected Test" option to "Selected Tests" (plural).  We could then add context menus on the grid expanders more fine grained control. Maybe something like:

```
Run Group
---------
Collapse
Expand
```
Individual tests *could* have context menus also, but I'm not sure what to put in them other than "Run Test" (although VS has an "Add to playlist option..." :grinning: )

Thoughts?

bclothier:
FWIW as long we gain a right-click menu to run a test in a group (regardless of what the grouping is set), I think losing the `Selected Category` and having `Run selected tests` makes more sense and as a bonus simplify the UI. I'm for it.

comintern:
Closed via #4769.

Issue #4703, opened 2019-01-10
Run Selected Category in test explorer does not work


bclothier:
Ref #4698

Though the category parameter categorizes the test methods just fine but unfortunately does not actually enable the `Run Selected Category` option. That's a bug. 

_Originally posted by @bclothier in https://github.com/rubberduck-vba/Rubberduck/issues/4698#issuecomment-453072604_

Issue #4702, opened 2019-01-10
Test Explorer needs to surface the category better


bclothier:
Ref #4698

Test explorer has supported categorization of tests but the feature is not very discoverable. We need to make it easier to find how to create a category, update it, and so forth.

A trivial implementation would be to update the template to include `@TestMethod("Uncategorized")` which would provide users with hints on how to use the category. A more fancy implementation would provide some kind of UI for labeling methods which would then rewrite the annotation.

An additional enhancement would be to enable categorization at the test module level, which would then apply to all test methods within the module. 

_Originally posted by @bclothier in https://github.com/rubberduck-vba/Rubberduck/issues/4698#issuecomment-453072604_

Issue #4658, opened 2018-12-18
Test engine needs to be updated to use private methods


bclothier:
Currently, it is possible to execute private procedures. Since test modules aren't meant to be run outside well, *tests*, it seems nice to be able to make all procedures as private, as to not pollute the intellisense / global namespace, especially in projects where it's not practical to have a separate test project. 

The change would be mainly be in the UI as it currently filters out any private procedures when discovering tests.

PR #4769, opened 2019-02-04
Refresh Testing (Pun Intended)

### LINKED ISSUE #3738: Rubberduck>UnitTest>Run All Tests doesn't update the grouping headers
Below is what is shown when run through Rubberduck>Unit Test>Run All Tests
rubberduck unittests runalltests

Below the grouping headers update appropriately if from the TestExplorer window Run button
runbutton on testexplorer

### LINKED ISSUE #3879: Smarten up the templates for unit testing
We support both early and late bound testing but what I don't like is that they derive from two different templates and make it hard to flip between the states. (You gonna help me enable my intellisense addiction, mmkay?)

My suggested template is provided below:

What it brings:

1) To change between early/late binding, we change *exactly* only one line -- The `#Const LateBind = ...` and add references accordingly.

2) Only one default template to maintain instead of 2. The setting used for binding when adding a test module would be then used to determine what default value of the `#LateBind` should be. WE always use `CreateObject` regardless of the mode, so we do not need to write `Set Assert = New Rubberduck.AssertClass`. 

3) This can be tied into a project-level compilation constant in case we need to support project-wide switch, rather than per-module. This is achieved with `#Const LateBind = LateBind` (assuming the `LateBind` is defined in the project's conditional compilation arguments). This approach also makes it easy for us to conditionally enable only early binding on a particular test module, rather than for all test modules. 

```
Option Compare Database
Option Explicit
Option Private Module

'@TestModule
'@Folder("Tests")

#Const LateBind = 0

#If LateBind Then
Private Assert As Object
Private Fakes As Object
#Else
Private Assert As Rubberduck.AssertClass
Private Fakes As Rubberduck.FakesProvider
#End If

'@ModuleInitialize
Public Sub ModuleInitialize()
    'this method runs once per module.
    Set Assert = CreateObject("Rubberduck.AssertClass")
    Set Fakes = CreateObject("Rubberduck.FakesProvider")
End Sub

'@ModuleCleanup
Public Sub ModuleCleanup()
    'this method runs once per module.
    Set Assert = Nothing
    Set Fakes = Nothing
End Sub

'@TestInitialize
Public Sub TestInitialize()
    'this method runs before every test in the module.
End Sub

'@TestCleanup
Public Sub TestCleanup()
    'this method runs after every test in the module.
End Sub
```

### LINKED ISSUE #4220: Localize the names of Test Modules
Test Modules are added with a number-suffixed, English module name like TestModule1, but when adding standard, class and document modules in different locales, the VBE uses a localized name.

We don't rely upon the name to identify test modules, so to be user friendly, we should be using localized names.

see #3879

### LINKED ISSUE #4365: Get UnitTesting internals out of Rubberduck.Core
There's some stuff in Rubberduck.Core that does no longer belong there, but which was too disjoint from #4146 to be pushed into there.

As such this is a follow-up issue to deal with the following more or less gaping issues:

 - IVBEInteraction is used outside of UnitTesting
 - AddTestModule command builds it's own template
 - AddTestMethod and AddTestMethodExpectedError don't call into UnitTesting to determine the next test name

Generally the internal API to Rubberduck should be consistent and cleanly exposed and separated into the Rubberduck.UnitTesting assembly

### LINKED ISSUE #4698: Refresh resets all test statuses
In the latest release (2.3.1.4308) a refresh will reset the status of all of my tests regardless of whether or not they have been affected. I remember this not being an issue with the previous version I used, although I can't remember which version that was. In general it makes writing tests and adding new tests a real chore as you have to either find the test you are working on in the menu and run that, or run every test for every change you make to either your code or your tests.

Ideally I want how it used to work as I remember that working "as expected".

**To reproduce**
1. Make an empty sheet with a test module with one empty test.
2. Run the test
3. Click refresh

_Edit (added the version/host info)_

Version 2.3.1.4308
Host Version: 16.0.11029.20108
Host Executable: EXCEL.EXE

### LINKED ISSUE #4701: Test explorer should support multiple selections
Ref #4968

Selection seems to be single and doesn't allow for multiple selections. I thought that was addressed in a previous PR but it does not look like it's the case. That would be an enhancement and would make `Run Selected` feature more useful

_Originally posted by @bclothier in https://github.com/rubberduck-vba/Rubberduck/issues/4698#issuecomment-453072604_

### LINKED ISSUE #4705: Invalidate test results on an individual basis rather than all results when running tests
Originally from https://github.com/rubberduck-vba/Rubberduck/issues/4698#issue-397357974

> In the latest release (2.3.1.4308) a refresh will reset the status of all of my tests regardless of whether or not they have been affected. I remember this not being an issue with the previous version I used, although I can't remember which version that was. In general it makes writing tests and adding new tests a real chore as you have to either find the test you are working on in the menu and run that, or run every test for every change you make to either your code or your tests.
> 
> Ideally I want how it used to work as I remember that working "as expected".
> 
> **To reproduce**
> 
>     1. Make an empty sheet with a test module with one empty test.
> 
>     2. Run the test
> 
>     3. Click refresh
> 
> 
> _Edit (added the version/host info)_
> 
> Version 2.3.1.4308
> OS: Microsoft Windows NT 10.0.17134.0, x64
> Host Product: Microsoft Office x86
> Host Version: 16.0.11029.20108
> Host Executable: EXCEL.EXE

The idea is that the test explorer should be more smart about invalidating only results for tests that actually got ran. The way it is, it will reset all tests' status, which is different from how most unit testing framework behaves. That would make RD's framework more consistent with say, NUnit or MSTests.

comintern:
This PR addresses several UI and structural concerns with Unit Testing.

screenshot from 2019-02-03 23-08-58

- The toolbar in the Test Explorer was refreshed to more closely match the updated CE and IR windows. This fixes the grouping and closes #3738.
- The Test Explorer is now multi-select capable, with options to run single tests via a context menu or all selected tests via either the context menu or the toolbar drop-down. This closes #4701, #4703, and #4704.
- Reparsing the project will now only reset the status of tests that have not been changed (any changes to the test procedure's source code will invalidate it).  In addition, there is now a button on the toolbar to reset all test statuses.  This closes #4705 and by extension closes #4698.
- All of the non-VB text in the test modules is now localizable, as is the module name. This closes #4220.
- There is a new "dual binding" option for added test modules. Given the upcoming addition of user templates and the redesign of the templating (see below), I'd say that this closes #3879.
- This introduces the `ITestCodeGenerator` interface, which moves all test code generation out of `Rubberduck.Core` and into `Rubberduck.UnitTesting`. 
- The `UnitTestSettings` and `UnitTestConfigProvider` are now in `Rubberduck.UnitTesting` where they belong.  This was necessary because of the dependency relationships between the two assemblies. That means they are no longer tied to the application settings, because well, that requires strong coupling and prevented separation of other more meaningful concerns.  Between this and the above item, this closes #4365.
- Test methods are now generated with a default category, addressing #4702. I may add an option prompt to select the category when a new method is added or put that in the context menu in the Test Explorer, but I'm up in the air about the best way to handle that at this point, so I'll leave it to @bclothier to decide if this PR is sufficient to close that issue.

Tagging this as WIP for now in that I still need to re-write a ton of unit test setup that the refactoring broke.  I also discovered that the test coverage for the view models is _really_ scanty, so I'll be adding more tests as appropriate. There's also still some low-hanging fruit in the open unit test issues that may get rolled into this during that process.

Finally, if #4717 merges first, I'm intending on adding collapse all and expand all buttons on the toolbar for the test groups.


Rubberduck.UnitTesting/UnitTesting/TestCodeGeneratorStatics.cs
```diff
@@ -0,0 +1,140 @@
+﻿using System;
+using Rubberduck.Resources.UnitTesting;
+
+namespace Rubberduck.UnitTesting.UnitTesting
+{
+    public partial class TestCodeGenerator
```
bclothier: IMPOV, all of the elements here should be coming from a resource file, not hard-coded within a `.cs` file. (ref #4622 for an example).  While making it implement the template (ref #4549 ) may be out of scope for the PR, it should be easy enough to have a resource file contain all the needed elements, which will then aid any future changes to get it to conform to the template model, and thus make the code less hard-coded. We want to avoid writing code that basically takes a bunch of elements and concatenate/interleave them because if the content changes, the code almost invariably must change, too.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
```diff
@@ -0,0 +1,231 @@
...
+            if (project == null || project.IsWrappingNullReference)
+            {
+                return;
+            }
+
+            var settings = _settings.Create();
+
+            if (settings.BindingMode == BindingMode.EarlyBinding)
+            {
+                // FIXME: Push the actual adding of TestModules into UnitTesting, which sidesteps VBEInteraction being inaccessble here
```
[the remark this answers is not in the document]
comintern: Nope. Missed that one. Will remove.

Rubberduck.UnitTesting/UnitTesting/ITestCodeGenerator.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using Rubberduck.Parsing.Symbols;
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+
+namespace Rubberduck.UnitTesting.UnitTesting
```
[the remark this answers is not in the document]
comintern: That was auto-generated when I added the class. `Rubberduck.UnitTesting.CodeGeneration` works for me. Will change.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
```diff
@@ -0,0 +1,231 @@
...
+                using (var component = components.Add(ComponentType.StandardModule))
+                using (var module = component.CodeModule)
+                {
+                    component.Name = GetNextTestModuleName(project);
+
+                    // Test modules always have appropriate options so remove any pre-generated code.
+                    if (module.CountOfLines > 0)
+                    {
+                        module.DeleteLines(1, module.CountOfLines);
+                    }
```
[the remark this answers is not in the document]
comintern: That deletes all code in the module.  The component is newly added, so the only code it can possibly contain at that point are the auto-generated options from the VBE.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        private string GetNewTestModuleCode(IVBComponent component, List<Declaration> stubs)
+        {
+            if (stubs is null || !stubs.Any())
+            {
+                return GetNewTestModuleCode(component);
+            }
+
+            var module = string.Join(Environment.NewLine + Environment.NewLine, new [] { GetBaseTestModule() }.Concat(stubs.Select(GetNewTestStubMethod)));
```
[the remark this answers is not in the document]
comintern: Will do. The settings get refreshed at the start of `GetBaseTestModule` if they aren't provided.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
```diff
@@ -0,0 +1,231 @@
...
+        private string GetBaseTestModule(UnitTestSettings settings = null)
+        {
+            if (settings is null)
+            {
+                settings = _settings.Create();
+            }
+
+            string declaration;
+            string initialization;
+            var asserts = settings.AssertMode == AssertMode.PermissiveAssert ? "PermissiveAssertClass" : "AssertClass";
```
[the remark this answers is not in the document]
comintern: We can't use `nameof` here because the types aren't referenced by `Rubberduck.UnitTesting`. I'll comment it.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            return $"{TestMethodBaseName}{index}";
+        }
+
+        private IEnumerable<Declaration> GetDeclarationsToStub(Declaration parentDeclaration)
+        {
+            return _state.DeclarationFinder.Members(parentDeclaration)
+                .Where(d =>
+                            Equals(d.ParentDeclaration, parentDeclaration) && d.Accessibility == Accessibility.Public &&
```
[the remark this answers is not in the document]
comintern: Agreed. That change will need to happen when #4658 is addressed.

Rubberduck.UnitTesting/UnitTesting/TestCodeGeneratorStatics.cs
```diff
@@ -0,0 +1,140 @@
...
+        private static string LateBindingDeclarations =>
+@"    Private Assert As Object
+    Private Fakes As Object";
+
+        private static string EarlyBindingDeclarations =>
+@"    Private Assert As Rubberduck.{0}
+    Private Fakes As Rubberduck.FakesProvider";
+
+        private static string DualBindingDeclarations =>
+$@"#Const {LateBindConstName} = 0
```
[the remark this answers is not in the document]
comintern: The only potential issue that I see with that is that it could potentially leave the project in a state where it couldn't be compiled after adding a test module, and we always do a reparse on a module add. I do need to verify that the RD reference gets added as needed when this template is used though.

Rubberduck.UnitTesting/UnitTesting/TestCodeGeneratorStatics.cs
comintern: This has to be done at run-time because the locale can be changed in the settings (this is why they're static properties instead of static fields). The difference between the code generation here and a predefined template is that it makes the localization much easier to maintain.

Rubberduck.UnitTesting/UnitTesting/TestCodeGenerator.cs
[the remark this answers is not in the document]
comintern: I'm an idiot - this _doesn't_ need to be addressed with #4658 - this is looking for methods to generate test stubs for, _not_ tests.  We're creating the tests in a different module, so they have to be public - otherwise the test stub won't be able to call it.  RD don't have to respect the accessibility, but VBA does...

Rubberduck.Core/UI/GridViewSort.cs
```diff
@@ -1,40 +1,40 @@
-﻿using System.Collections.Generic;
-using System.Linq;
+﻿//using System.Collections.Generic;
```
bclothier: Did you mean to leave this commented-out file in?

Rubberduck.Core/UI/UnitTesting/TestExplorerModel.cs
```diff
@@ -125,13 +86,198 @@ public bool IsBusy
...
+        private static readonly Dictionary<TestOutcome, Color> OutcomeColors = new Dictionary<TestOutcome, Color>
+        {
+            { TestOutcome.Unknown, Colors.DimGray },
+            { TestOutcome.Succeeded, Colors.LimeGreen },
+            { TestOutcome.Inconclusive, Colors.Gold },
+            { TestOutcome.Ignored, Colors.Orange },
+            { TestOutcome.Failed, Colors.Red }
+        };
+
+        private TestOutcome _bestOutcome = TestOutcome.Unknown;
```
bclothier: "Best" implies that if a single test passes with 99 failures, we get green. I'm pretty sure that's not the "best" outcome. Maybe `_worstOutcome`? `_lcdOutcome`?

Rubberduck.Core/UI/UnitTesting/TestExplorerModel.cs
```diff
@@ -125,13 +86,198 @@ public bool IsBusy
...
+                return;
+            }
+
+            IsBusy = true;
+
+            try
+            {
+                _testEngine.Run(tests.Select(test => test.Method));
+            }
+            catch (Exception ex)
```
bclothier: Would the cancellation get propagated to here? If so, shouldn't we have a `ThreadAbortException` catch block?

Rubberduck.Core/UI/GridViewSort.cs
comintern: Nope, I commented it out to see if it was used and forgot to remove it.  Will fix.

Rubberduck.Core/UI/UnitTesting/TestExplorerModel.cs
comintern: Doh! I somehow missed that when the logic flipped.  Will fix.

Rubberduck.Core/UI/UnitTesting/TestExplorerModel.cs
comintern: The cancellation won't raise a `ThreadAbortException` - the thought of mixing a task cancellation on the UI thread with our own custom message pumping tossed in the mix made me cringe.  It's handled by a simple state flag that just lets the invoke exit normally. Pumping the messages lets the cancel button call the `RequestCancellation()` method directly on the `TestEngine`. The initial discussion on the issue #3158 was **_way_** overthinking this.

Rubberduck.Parsing/UIContext/UiDispatcher.cs
```diff
@@ -53,16 +41,40 @@ public void Invoke(Action action)
...
+                    NativeMethods.TranslateMessage(ref message);
+                    NativeMethods.DispatchMessage(ref message);
+                }
+
+                handle.Free();
+            }
+
+            if (_contextProvider.UiContext == SynchronizationContext.Current)
+            {
+                PumpMessages();
```
bclothier: FWIW, I tripped over this as at first read, I thought I was reading a recursive method but now I see that it's a local function. I assume you elected for a local so it wouldn't be used anywhere else but this method. Maybe consider moving it to be the very first or the very last thing within the method and provide a comment?

Rubberduck.Resources/Settings/UnitTestingPage.resx
```diff
@@ -129,6 +129,9 @@
   <data name="UnitTestSettings_BindingMode" xml:space="preserve">
     <value>Binding mode:</value>
   </data>
+  <data name="UnitTestSettings_DualBinding" xml:space="preserve">
```
bclothier: This might be beyond the scope of the PR but now that we're splitting up the resx files, the prefix seems redundant and probably should be gotten rid of at some point in future.

Rubberduck.UnitTesting/CodeGeneration/ITestCodeGenerator.cs
```diff
@@ -0,0 +1,13 @@
...
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+
+namespace Rubberduck.UnitTesting.CodeGeneration
+{
+    public interface ITestCodeGenerator
+    {
+        void AddTestModuleToProject(IVBProject project);
+        void AddTestModuleToProject(IVBProject project, Declaration stubSource);
+        string GetNewTestMethodCode(IVBComponent component);
+        string GetNewTestMethodCodeErrorExpected(IVBComponent component);
```
bclothier: FWIW - those 2 methods could be consolidated with a 2nd parameter -- perhaps an enum -- to make it easier to extend the stubs beyond the default & expected error templates.

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
```diff
@@ -34,16 +34,26 @@ internal class TestEngine : ITestEngine
...
+
+        public IReadOnlyList<TestMethod> LastRunTests => _lastRun;
+
         public bool CanRun => AllowedRunStates.Contains(_state.Status) && _vbe.IsInDesignMode;
-        public bool CanRepeatLastRun => LastRun.Any();
+        public bool CanRepeatLastRun => _lastRun.Any();
         
-        private bool refreshBackoff;
-
+        private bool _refreshBackoff;
```
bclothier: TBH the name of the variable doesn't make sense - "turn refresh back off" is what I think of. Is that what it means?

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
```diff
@@ -229,22 +274,26 @@ private void RunWhileSuspended(IEnumerable<TestMethod> tests)
                                     Logger.Trace(trace, "Unexpected COMException when running TestInitialize");
                                     continue;
                                 }
+
+                                _uiDispatcher.FlushMessageQueue();
```
bclothier: This seems to be the only one place in the `RunWhileSuspended` method where we pump the message but we've handled several other events. Shouldn't we pump before/after those other events?

Rubberduck.UnitTesting/UnitTesting/TestResult.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Rubberduck.UnitTesting
 {
-    public class TestResult
+    public struct TestResult
```
bclothier: Why not `readonly struct`?

Rubberduck.Core/UI/UnitTesting/TestExplorerViewModel.cs
```diff
@@ -1,222 +1,280 @@
...
         public RunFailedTestsCommand RunFailedTestsCommand { get; set; }
         public RunSucceededTestsCommand RunPassedTestsCommand { get; set; }
-        public CommandBase RunSelectedTestCommand { get; }
-        public CommandBase RunSelectedCategoryTestsCommand { get; }
+        public CommandBase RunSingleTestCommand { get; }
+        public CommandBase RunSelectedTestsCommand { get; }
+        public CommandBase RunSelectedGroupCommand { get; }
 
+        public CommandBase CancelTestRunCommand { get; }
+        public CommandBase ResetResultsCommand { get; }
```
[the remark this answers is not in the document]
comintern: The readonly commands assigned in the ctor are specific to the VM (i.e. use VM assets), whereas the injected commands are used throughout the UI.  We're using the same pattern in the other tool window controls, so I was just going with it.  I can add a comment if you want though.

Rubberduck.Core/UI/UnitTesting/TestExplorerViewModel.cs
```diff
@@ -249,28 +307,6 @@ private void ExecuteCopyResultsCommand(object parameter)
...
-        {
-            if (Model.IsBusy || SelectedItem == null)
-            {
-                return false;
-            }
-
-            return ((TestMethod) SelectedItem).Category.Name != string.Empty;
-        }
-
         //KEEP THIS, AS IT MAKES FOR THE BASIS OF A USEFUL *SUMMARY* REPORT
```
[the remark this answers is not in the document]
comintern: On quick inspection, probably both. :-D  TBH, all of the copy to clipboard stuff needs to be abstracted a bit (DRY).  It's on my radar...

Rubberduck.Parsing/UIContext/UiDispatcher.cs
comintern: You hit the nail on the head re why it's a local function. I'll move it to the bottom (and comment) - I think you're right about it looking "less like recursion" there.

Rubberduck.Resources/Settings/UnitTestingPage.resx
comintern: Agreed.  The resx files are generally in bad need of clean-up and rearranging. That's a fairly monumental task in and of itself though. Maybe it needs an issue?

Rubberduck.UnitTesting/CodeGeneration/ITestCodeGenerator.cs
comintern: I could see that, although a two member enum would strike me as a bit weird.  IIR this was the existing naming (or close), but I think if this were extended an enum might be in order.

Rubberduck.Resources/Settings/UnitTestingPage.resx
bclothier: [Left a comment](https://github.com/rubberduck-vba/Rubberduck/issues/4541#issuecomment-464053990) on #4541 .

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
comintern: I must have looked at that a hundred times and and said "huh"? It means "don't refresh from the parser if this flag is set and it's otherwise ready".  Any suggestions? Maybe `_deferRefresh` or `_refreshUnsafe`?

Rubberduck.Parsing/UIContext/UiDispatcher.cs
MDoerner: At least, with the new build setup, they seem to auto-generated the appropriate code again.

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
bclothier: I probably would flip the logic and call it `_reactToParserStates` or more simply (but less descriptively), `_listening`.

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
comintern: That was intentional.  I'm pumping here to catch a possible cancel button press, and by my read that was the only safe place to do it given the requirements for running the test setup and teardown procedures. The templates are probably safe to cancel in the middle of those, but we run every setup method we find so I'm thinking it's best to treat those as a unit (and not to cancel those steps). The rest of the pumping is done prior to raising status events from the `TestEngine`.  In those instances, we _know_ the apartment boundary is in a good state to pump.

Rubberduck.UnitTesting/UnitTesting/TestResult.cs
comintern: Because I let R# do that and the IDE complains that it's a 7.2 feature?  :-P

Will fix - it indeed should be readonly.

Rubberduck.UnitTesting/UnitTesting/TestEngine.cs
bclothier: Then that needs to be put in a comment (xmldoc?). That intention is not apparent from reading the source code alone and we'll need to know that when we modify this in future.

comintern: OK, I'll actually comment _all_ uses of `FlushMessageQueue` to be safe. There's already an extensive xmldoc warning on the interface itself about safe usage (you should see the tooltip).

Rubberduck.Core/UI/UnitTesting/TestExplorerViewModel.cs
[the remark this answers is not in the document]
bclothier: That's what I suspected and reading it again, I'm torn - we would have to choose between grouping the injected delegates and homebrewn delegates versus listing the delegates based on their function/names. Both approaches have their merits so we'll have to pick one or other.

Issue #4588, opened 2018-12-01
Wrong Interface implementation


Stonovsky:
Sometime it happens, that there is wrong interface implementation. I tried to figure it out but I'm not sure when (it may be after renaming Interface class using Rubberduck). Please see following code with wrong implementation:

```
Option Compare Database
Option Explicit
Implements clsIWstawTowarDoFrm

Private Property Get clsIWalidacja_CzyWalidacjaPozytywna() As Boolean
    Err.Raise 5 'TODO implement interface member
End Property

Private Property Get clsIWalidacja_PowodBrakuWalidacji() As String
    Err.Raise 5 'TODO implement interface member
End Property
```
Please help me figure it out because I cannot work with rubberduck interface implementation as it's always implelements only one (wrong) interface. Thank you in advance for help.

comintern:
I was able to (kind of) get a repro on this. If you implement an interface, then delete the implementation, renaming the interface will re-implement it (with the new names) in the implementing class. This is apparently due to the rename refactoring not checking to see if a module is dirty before rewriting it with the new naming.

So @Stonovsky, just to be clear - did the implement interface command result in the wrongly named implementation, or did the rename add them?

I think part of the issue here might be in the UI of the refactoring itself.  If you open a new class module and add an `Implements` statement to the top like this...

```
Option Explicit
Implements IFoo
```
...that will result in un-compilable code.  The parser will warn about it, but you have to allow the parse to progress in order to run the refactoring - otherwise, RD can't see the `Implements` statement or find the declaration for `IFoo` in order to implement it. Not sure if that's relevant to this issue though, because we are explicitly checking to see if either the interface or target is dirty in `RefactorImplementInterfaceCommand.EvaluateCanExecute`. I'm wondering if we should simply have the user select from a list of user defined interfaces in a dialog instead of requiring them to type the `Implements` statement and parse un-compilable code before the refactoring is available. The only reason we should *really* be concerned with a `Implements` statement that already exists in parsed code would be to complete implementations that are missing members of the interface.

Stonovsky:
Thank you for the explanation. Do I understand you right: When I create class, save it with needed name, then type Implements Ifoo, then I need to parse code in order to detect refactor implements possibilities by RB and implement the Interface. When I do all above I cannot change name of interface class? What is the best way to create implementation in class incl. parsing?

Stonovsky:
[...]

PR #4777, opened 2019-02-08
Recover selection after rewrite

### LINKED ISSUE #4524: RD should recover the selection after each rewrite in a consistent way
Now that PR #4465 is merged we have one place all rewriting has to go through, the `IRewriteSession`. This presents an opportunity to consistently recover the previous selection after every rewrite.

Currently, each refactoring implements its own way of recovering the selection in the active code pane, wherease the quickfixes make no attempt at all to recover the selection.

My suggestion is to introduce some `ISelectionService` with the capability to save the selection for specified modules and recover it on request and to wire this up in the implementations of `IRewriteSession`.

Recovering the selection would include:
1. Set the selection to the selection prior to the rewrite for all affected modules.
2. Activate the previously active code pane, if there was any.

For attribute rewrites, we should make sure the same code panes are open after the rewrite. (Attributes get rewritten via exporting, removing and then reimporting the modules.)

In a later stage members to add an offset to the selection or to specify the selection could be added to the `IModuleRewriter` in order to allow for adjustments considering the code altered during the rewrite.

### LINKED ISSUE #4773: Refactor renaming via VBA IDE project explorer does not remame in the code
Version 2.4.0.4511
Host Version: 16.0.11231.20130
Host Executable: WINWORD.EXE

**Description**
If I click on a class name in the VBA IDE and use RubberDuck.Refactor.Rename then the class name is changed in the project explorer but this change is not propagated into the code base.

**Logfile**

[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2836360/RubberduckLog.txt)

MDoerner:
This PR closes #4524. 
This PR introduces the `ISelectionRecoverer` that allows saving a later recovering the selection of delected modules, either at am explicit time or on the next parse. This is then wired up in the `RewriteSessionBase` to recover the selections after each rewrite.

Moreover, there are two new mwmbers on the `IModuleRewriter` that allow to add a selection offset (`SelectionOffset`) or replace the selection (`Selection`) on the rewrite. The latter has precedence, i.e. any offset is ignored if the selection is set.

Further steps could include removing the selection recovery from the refactorings and setting up useful selection offsets/changes in the refactorings ans quickfixes.

Update:

This PR now also introduces recovery of the active code pane.
Furthermore, explicit selection recovery has been removed from the `RenameRefactoring` because it is redundant now. This closes  #4773.  

Further update:

Reopeneing opened code panes on attribute rewrites is now included as well.

Testing in the VBE, I found a problem: all modules that get their selection recovered also get opened.
This is caused because we force focus to the selection for some reason when we set it, in contrast to the VBE's behaviour.

Further Update:

Now, this PR includes moving all users of the `ICodePane.Selection` property to the `ISelectionService` that do not handle SCWs anyway. An additional exception is `ExtractMethod`, which I left alone as much as possible.

This also led to a rework of `MoveCloserToUsage`.

Unfortunately, the issue of opening code paned on rewrite still persists. So, more investigation is required. However, I would appreciate comments on whatt is already there. 

Final Update:

Opening the code pane is a side effect of accessing it, provided it is not open already. This is generally ture. So, selection recovery had to be restricted to open code panes. (I guess nobody will dicover this restriction anyway.)

The last commit adds a fix for `DeclarationFinder.FindInterface`, which should fix some strange issues with the `ImplementInterfaceRefactoring`. Ref #4588


MDoerner:
This PR is no longer WIP. I would appreciate any feedback.
In particular, I would be interested to know whether I overlooked something important in the redesign of the `MoveCloserToUsageRefactoring`.

Rubberduck.VBEEditor/SafeComWrappers/VB/Abstract/ICodeModule.cs
```diff
@@ -6,6 +6,10 @@ public interface ICodeModule : ISafeComWrapper, IEquatable<ICodeModule>
     {
         IVBE VBE { get; }
         IVBComponent Parent { get; }
+        /// <summary>
+        /// Returns the code pane associated with the cod module.
+        /// Accessing this property will open the code pane if it is not open already.
+        /// </summary>
         ICodePane CodePane { get; }
```
bclothier: Just a thought - instead of exposing the property straight, we could consider wrapping it in `TryGetOpenCodePane()` and `OpenCodePane()`. That would provide the benefit of separating the action of getting a code pane from opening a code pane, and require the caller to be explicit about what it wants to do with the code pane.

Rubberduck.Core/UI/Command/Refactorings/RefactorEncapsulateFieldCommand.cs
```diff
@@ -5,40 +5,41 @@
...
             _indenter = indenter;
             _factory = factory;
         }
 
         protected override bool EvaluateCanExecute(object parameter)
         {
-            Declaration target;
-            using (var pane = Vbe.ActiveCodePane)
+            //This should come first because it does not require COM access.
+            if (_state.Status != ParserState.Ready)
```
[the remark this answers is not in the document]
MDoerner: I agree with the idea, but this is out of scope for this PR.

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
```diff
@@ -0,0 +1,181 @@
...
+            _savedOpenModules.Clear();
+            var openModules = _selectionService.OpenModules();
+            _savedOpenModules.UnionWith(modules.Where(module => openModules.Contains(module)));
+        }
+
+        public void RecoverOpenState()
+        {
+            foreach (var module in _savedOpenModules)
+            {
+                _selectionService.TryActivate(module);
```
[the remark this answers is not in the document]
MDoerner: I do not think that the z-ordering is really important. Moreover, I have no idea where that would be readily exposed.

Rubberduck.Refactorings/MoveCloserToUsage/MoveCloserToUsageRefactoring.cs
```diff
@@ -155,31 +224,52 @@ private void InsertNewDeclaration(IRewriteSession rewriteSession)
...
+                {
+                    return $"{declarationText}{Environment.NewLine}{precedingWhitespaces[0]?.GetText()}";
+                }
+
+                return $"{declarationText}{Environment.NewLine}";
+            }
+            //This is the first statement, which actually should not happen.
+            if (blockStmtContext.TryGetPrecedingContext(out VBAParser.WhiteSpaceContext startingWhitespace))
+            {
+                return $"{declarationText}{Environment.NewLine}{startingWhitespace?.GetText()}";
```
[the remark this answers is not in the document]
MDoerner: We are in the situation that we move something into a method. So, the moved thing cannot really be the first statement.

The string the method returns is exactly the string appropriate for the first statement in a module.

Rubberduck.Refactorings/RemoveParameters/RemoveParametersRefactoring.cs
```diff
@@ -9,32 +9,31 @@
...
 namespace Rubberduck.Refactorings.RemoveParameters
 {
     public class RemoveParametersRefactoring : IRefactoring
     {
-        private readonly RubberduckParserState _state;
-        private readonly IVBE _vbe;
-        private readonly IRefactoringPresenterFactory _factory;
+        private readonly IDeclarationFinderProvider _declarationFinderProvider;
+        private readonly ISelectionService _selectionService;
+        private readonly Func<RemoveParametersModel, IDisposalActionContainer<IRemoveParametersPresenter>> _presenterFactory;
```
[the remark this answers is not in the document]
MDoerner: We should actually have a `RefactoringBase` in the first place. All refactorings need an `IRewritingManager` and an `ISelectionService`. Moreover, after changing the call sequence of the overloads of `Refactor` to a sensible one, the parameterless overload would be identical for all refactorings.

As you mentioned already, this is out of scope for this PR, though.

Rubberduck.VBEEditor/SafeComWrappers/VB/Abstract/ICodeModule.cs
MDoerner: Currently, the SCWs copy and extend the VBE API. Removing this property would destroy that.

Moreover, this would require each access to a code pane to make a lot of COM calls: validating whether a code pane is open requires iterating over the open code panes exposed via `VBE.CodePanes`. 

So, although this separates getting the code pane from opening it, I am not a huge fan of this idea.

Rubberduck.Refactorings/ReorderParameters/ReorderParametersRefactoring.cs
```diff
@@ -77,14 +77,9 @@ public void Refactor()
...
             {
-                if (pane == null || pane.IsWrappingNullReference)
-                {
-                    return;
-                }
-
-                pane.Selection = target.Selection;
+                return;
             }
             Refactor();
```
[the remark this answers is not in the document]
MDoerner: Everywhere else we also follow the convention to bail out on error states and then to do the processing for the good state, independent of the work required for the good state.

Rubberduck.Refactorings/MoveCloserToUsage/MoveCloserToUsageRefactoring.cs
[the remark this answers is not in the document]
bclothier: May I suggest rephrasing the comment: `//This is the first statement after declarations so no padding is needed` and `//This seems to be at the start of module, which shouldn't be possible so we will preserve declarations to be safe`? That should make clear that those are not "errors" but rather preventing potential errors due to edge cases.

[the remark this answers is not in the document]
MDoerner: I can change the comment a bit. 

However, note that this case has nothing to do with declarations. It only applies to the very first statement in the module. The difference in the two cases is that in the first case the very first statement in the module is indented and in the second it is not.

Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
```diff
@@ -233,18 +236,8 @@ public CommandBase OpenTodoSettings
...
-            get
-            {
-                if (_navigateCommand != null)
-                {
-                    return _navigateCommand;
-                }
-                return _navigateCommand = new NavigateCommand(_state.ProjectsProvider);
-            }
-        }
+        private readonly Lazy<NavigateCommand> _navigateCommand;
```
comintern: Is there a reason to make this instantiate lazily? If it's bound to the view it would be forced to materialize as soon as the control its bound to does, which would be immediately after the ctor exited.

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Rubberduck.Parsing.VBA;
+using Rubberduck.VBEditor;
+using Rubberduck.VBEditor.Extensions;
+using Rubberduck.VBEditor.Utility;
+
+namespace Rubberduck.Parsing.Rewriter
+{
+    public sealed class SelectionRecoverer : ISelectionRecoverer, IDisposable
```
comintern: Is this guaranteed to only be called from a single thread?

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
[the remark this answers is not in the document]
comintern: @bclothier If we really want the z-ordering, I could probably track and expose that through either the subclass provider or the `VbeEvents`. If you think this is important enough, open an enhancement issue for it.

Rubberduck.Refactorings/Rename/RenameRefactoring.cs
```diff
@@ -483,19 +476,23 @@ private void RenameModule(IRewriteSession rewriteSession)
...
-                            // VBA - rename code module
-                            using (var codeModule = component.CodeModule)
+                            if (vbe.Kind == VBEKind.Hosted)
                             {
-                                Debug.Assert(!codeModule.IsWrappingNullReference, "input validation fail: Attempting to rename an ICodeModule wrapping a null reference");
-                                codeModule.Name = _model.NewName;
+                                // VBA - rename code module
+                                using (var codeModule = component.CodeModule)
+                                {
+                                    Debug.Assert(!codeModule.IsWrappingNullReference,
```
comintern: This should probably log rather than assert.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBE.cs
```diff
@@ -29,11 +29,42 @@ public VBE(VB.VBE target, bool rewrapping = false)
...
             {
-                if (!IsWrappingNullReference)
+                return;
+            }
+
+            Target.ActiveCodePane = pane.Target;
+            ForceFocus(codePane);
+        }
+
+        private void ForceFocus(ICodePane codePane)
```
comintern: Can you do me a favor and drop a TODO on this? The subclass manager should have all of the window handles associated with code panes now - I can probably do this without using FindChildWindow at this point.

MDoerner: Sure, I can add a todo. Just one question, is the subclass manager in an assembly we can reference from here?

Rubberduck.Refactorings/Rename/RenameRefactoring.cs
MDoerner: Although I agree, I will not change this in this PR; I will make a note for my next/new one.

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
MDoerner: No. Why do you ask?

Currently, this is only used by the rewrite sessions.

Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
MDoerner: I have no idea, I just changed the hand crafted one to one using the `Lazy<T>` type.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBE.cs
bclothier: The [subclass manager](https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.VBEEditor/WindowsApi/SubclassManager.cs) is in the `Rubberduck.VBEEditor` assembly; you should be able to reference it from VBA/VB6 assembly.

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
comintern: My main concern was the thread safety of the caches vis-a-vis `SaveSelections` and other callers. If it's only being called by rewrite sessions, that _should_ be fine.

Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
comintern: This might be better just loading it in the ctor then - the instantiation of the window is wiring up all of the event handlers to deal with the wpf bindings,  and I'm not sure what expectations the framework has regarding the immediate availability of bound controls.

Rubberduck.Parsing/Rewriter/SelectionRecoverer.cs
MDoerner: Note thatit is registered in transient scope. So, all users will get their own cache. Well, the rewrite sessions all get the same one via their fatcory, but they are guaranteed to never execute in parallel.

PR #4827, opened 2019-02-26
TODO Explorer refresh

comintern:
Refresh Todo Explorer to match other UI elements, refactor grouping to get rid of duplicated grid control.

For those of us who have difficultly reading raw PNG data, it looks like this:

screenshot from 2019-02-25 18-06-15

Not much to it, just added the expand and collapse commands to match similar controls and changed the grouping to toggle buttons instead of a dropdown.

Feel free to suggest different icons for the marker grouping and remove (or in the case above - "complete") icons.  Sticky notes were a first thought, but I'm not sure if I'm sold on them.

WIP pending some fixes to a couple other things that I found broken.


Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
```diff
@@ -123,122 +119,109 @@ private void HandleStateChanged(object sender, EventArgs e)
...
+            }
 
-                    RefreshCommand.Execute(null);
-                }
-                );
+            var component = _state.ProjectsProvider.Component(_selectedItem.Selection.QualifiedName);
+            using (var module = component.CodeModule)
+            {
+                var oldContent = module.GetLines(_selectedItem.Selection.Selection.StartLine, 1);
+                var newContent = oldContent.Remove(_selectedItem.Selection.Selection.StartColumn - 1);
```
Vogel612: This looks weird. We clear the matching line up to the start of the selection?

RubberduckTests/TodoExplorer/TodoExplorerTests.cs
```diff
@@ -219,5 +226,12 @@ private Configuration GetTodoConfig(string[] markers)
             var userSettings = new UserSettings(null, null, null, todoSettings, null, null, null, null);
             return new Configuration(userSettings);
         }
+
+        private IUiDispatcher GetMockedUiDispatcher()
+        {
+            var dispatcher = new Mock<IUiDispatcher>();
+            dispatcher.Setup(m => m.Invoke(It.IsAny<Action>())).Callback((Action argument) => argument.Invoke());
+            return dispatcher.Object;
+        }
```
Vogel612: This mock setup will definitely be used in other tests. Do we not have a central place for this yet?

Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
```diff
@@ -123,122 +119,109 @@ private void HandleStateChanged(object sender, EventArgs e)
...
+                foreach (var item in _state.AllComments.SelectMany(GetToDoMarkers))
+                {
+                    _items.Add(item);
+                }
+            });
         }
 
-        private ToDoItem _selectedItem;
-        public INavigateSource SelectedItem
+        public INavigateCommand NavigateCommand { get; }
```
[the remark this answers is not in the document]
comintern: Until #4777, that was instantiated lazily in the `get`. 4777 moved it to the ctor and made it explicitly `Lazy<NavigateCommand>`.  The only real change here is that it's not lazy any more (that was preventing it from binding correctly until the user interacted with the UI). I didn't look into the possibility of injecting it, but if it's possible it seems reasonable - it's just not really in scope for this PR.

Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
comintern: The selection there is the comment (this also only shown in the diff because I moved it).  It's had that behavior as far back as I can go in the commit history.

RubberduckTests/TodoExplorer/TodoExplorerTests.cs
comintern: Not yet.  I think? it's currently used by two other test suites (the Code Explorer and unit testing). In those cases, the setup is basically just a one liner in the constructor of the test setup object though.

Issue #4757, opened 2019-01-29
Unhandled Exception when using PPT


daFreeMan:
**Rubberduck version information**
Version 2.4.0.4491
OS: Microsoft Windows NT 10.0.15063.0, x64
Host Product: Microsoft Office 2016 x64
Host Version: 16.0.4266.1001
Host Executable: POWERPNT.EXE

**Description**
I had a PPTM open with a code module opened in the VBE & parsed by RD. I opened a 2nd PPT (without code), closed it and opened a 3rd PPT (again without code). Instead of opening the 3rd PPT, I got the error message.

**To Reproduce**
See "Description"

**Expected behavior**
No `Unhandled exception` should occur.

**Screenshots**
![2019-01-29 15_16_13-quarterly reports - master slide list pptm - powerpoint](https://user-images.githubusercontent.com/11889733/51937873-ad11a180-23d9-11e9-94bf-f5cbe8090aaf.png)

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2809520/RubberduckLog.txt)

**Additional context**
Here's the .NET debug details:

```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.Runtime.InteropServices.InvalidComObjectException: COM object that has been separated from its underlying RCW cannot be used.
   at System.StubHelpers.StubHelpers.GetCOMIPFromRCW(Object objSrc, IntPtr pCPCMD, IntPtr& ppTarget, Boolean& pfNeedsRelease)
   at Microsoft.Vbe.Interop._VBProject.get_References()
   at Rubberduck.VBEditor.SafeComWrappers.VBA.VBProject.get_References() in C:\projects\rubberduck\Rubberduck.VBEditor.VBA\SafeComWrappers\VB\VBProject.cs:line 45
   at Rubberduck.Navigation.CodeExplorer.CodeExplorerProjectViewModel.GetProjectReferenceModels() in C:\projects\rubberduck\Rubberduck.Core\Navigation\CodeExplorer\CodeExplorerProjectViewModel.cs:line 150
   at Rubberduck.Navigation.CodeExplorer.CodeExplorerProjectViewModel.SynchronizeReferences() in C:\projects\rubberduck\Rubberduck.Core\Navigation\CodeExplorer\CodeExplorerProjectViewModel.cs:line 117
[...]

Issue #4489, opened 2018-11-04
The tests should more follow the DRY principle


MDoerner:
We have a lot of tests that basically do the same thing: they define input code, set up a `Mock<IVBE>` using that, use the `MockParser` to get a `RubberduckParserState` and parse, then generate a result based on this and assert on some property of the result.

In several domains (inpections, quickfixes,...) or as least within the same test class, the result generated is nearly always identical, e.g. the inspection tests generally new up the inspection based on the state and assert on some property of the inspection results. Accordingly, there is no need to repeat the implementation of how the result is produced over and over.

One major problem with this repetition is that it makes changes to the parsing setup rather painful since all individual tests may need to be updated.

Because of this, I think it would be good to have one abstract base class for each domain containing a general method to generate the result to assert on, e.g. `IEnumerable<IInspectionResult> InspectionResults(string code)`  for the inspections, or at least one such setup within the individual classes, in case a general setup for the domain is too complicated. In PR #4465, this has already been realized for the quickfixes.

bclothier:
In the specific scenarios of where arguments are provided via constructor injection, the long-delayed PR #4072 introduces a Mock IoC to help with abstracting out this setup and also offer better fidelity to the actual production use in cases where features of CW are used (e.g. magic factories).

PR #4833, opened 2019-03-01
Introduce ComCommand classes

### LINKED ISSUE #4269: Fix RCW already released warnings at shutdown
CommandBarControls currently emit the following warning at shutdown:

`2018-08-12 19:41:55.4990;WARN 2.2.6798.35429;Rubberduck.VBEditor.SafeComWrappers.SafeComWrapper1;Released COM wrapper of type Rubberduck.VBEditor.SafeComWrappers.Office12.CommandBarControl whose underlying RCW has already been released from outside the SafeComWrapper. New reference count is -1.;`

indicating that too many disposals are occurring. Affects both VBA and VB6.

bclothier:
Closes #4269

The PR means to help address concerns identified from the issues #4767 and or #4757 -- the trend being that we are trying to access COM after the shutdown has started. Those issues are actually with the code explorer but the assumption is that it need not be just the code explorer; any subsystems that does COM access needs to be prevented once the shutdown has started.

For a while we've had `VbeEvents` which provides a `EventsTerminated` event to broadcast the shutdown but it has been unused... up to now.

This first stage of work focuses on controlling the commands. This required 2 changes:

1. Because the commands are transient, it does no good to have them listening to an event which may have already happened. Thus, `VbeEvents` now exposes a property `Terminated` to ensures that any new instances created after the shutdown inadvertently will be disabled accordingly. Accordingly, the `VbeEvents` has been cleaned up a bit. Note that I opted to use `Interlocked` to provide very basic locking. I considered a more elaborate scheme but decided it was best to first demonstrate that there is in fact a need before going to the trouble.

2. Commands that touches a COM object now derive from the `ComCommandBase` abstract class. The main difference between the `ComCommandBase` and `CommandBase` is simply that `ComCommandBase` takes `IVBEEvents` as a dependency and wraps checks around the `CanExecute` and `Execute` which prevents the commands from being used once we are shutting down. 

3. As part of refactoring, the tests has been updated in accordance with #4489 - namely the numbers of `new`s has been cut down and converted into a set of overloaded static factory methods. 

TODO:

1. ~Update the commands outside the `UI\Command` (e.g. `AutoComplete`, `CodeExplorer` and `UnitTesting`~ -- there is also `Refactoring` but the refactorings will be skipped over at MDoerner's request).

2. ~Add more unit tests? More live testing~

NOTE:

There are subsystems that will need similar treatments. A good example is the test engine which performs several COM accesses. Such subsystem will need its own implementation of _something_ to cut off the access after the shutdown. It is more likely that those will be addressed in separate PRs. For now, this PR deals with the commands. 

Addendum: The PR also addresses the issues with the ~`ShowIntellisenseCommand`~ `ShowQuickInfoCommand` as identified in #4110, suppressing the beep message when executing. Thus we now have a working implementation. Furthermore, it has been tested in VB6 as well as VBA, and verified to be working in both environment so AutoCompletion is now able to trigger the quick info without annoying the users.


Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using NLog;
+using Rubberduck.VBEditor.Events;
+
+namespace Rubberduck.UI.Command.ComCommands
```
comintern: Nitpick - Is there any reason that this needs to be in its own namespace? The folder obviously makes sense organizationally, but sub-dividing the namespace doesn't seem like it does much other than adding more noise to the usings - I'm having difficulty seeing this as a qualitatively different enough implementation to warrant that.

Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public abstract class ComCommandBase : CommandBase
+    {
+        private readonly IVBEEvents _vbeEvents;
+
+        protected ComCommandBase(ILogger logger, IVBEEvents vbeEvents) : base(logger)
```
comintern: Probably out of scope, but wouldn't it be simpler to just grab the class logger in the ctor instead of having it in every single derived class's `base` call? I'm guessing this was initially done because `GetCurrentClassLogger()` returns the abstract base class logger (I consider that a bug, [NLog disagrees](https://github.com/NLog/NLog/issues/2141)), but there's [a work-around here](https://stackoverflow.com/a/35119353/4088852).

Rubberduck.VBEEditor/Events/IVbeEvents.cs
```diff
@@ -18,5 +18,6 @@ public interface IVBEEvents
         event EventHandler<ReferenceEventArgs> ProjectReferenceAdded;
         event EventHandler<ReferenceEventArgs> ProjectReferenceRemoved;
         event EventHandler EventsTerminated;
+        bool Terminated { get; }
```
comintern: Am I missing something, or is the diff just screwed up?  I don't see this implemented in VbeEvents.cs.

Rubberduck.VBEEditor/Events/VbeEvents.cs
```diff
@@ -53,6 +53,11 @@ private VBEEvents(IVBE vbe)
             
             _projects = vbe.VBProjects;
 
+            if (_projects == null)
```
comintern: This is deterministically never null - `VBProjects` is returned wrapped in a SCW. It should be `if(_projects.IsWrappingNullReference)`

Rubberduck.VBEEditor/Events/VbeEvents.cs
```diff
@@ -227,51 +232,34 @@ private void ProjectReferenceRemovedHandler(object sender, ReferenceEventArgs e)
...
+                UnregisterProjectHandlers(projectid);
             }
-        }
 
-        ~VBEEvents()
-        {
-            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
-            Dispose(false);
-        }
+            if (_projects == null)
```
comintern: See above. The only way this will ever be null is if something kills it after the ctor runs.  Should probably be something more like ``if(_projects?.IsWrappingNullReference ?? true)`

Rubberduck.VBEEditor/Events/VbeEvents.cs
bclothier: I added that for the unit tests' sake since that was returning `null` as I didn't set it up and I got worried that there might be a weird edge case where the `IVBE` might decide to dispense a `null` for the `VBProjects` collection so figured an extra guard wouldn't hurt.

comintern: In that case, wouldn't you want the test to fail?

Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
bclothier: The primary reasoning was to make it easy to write a code analysis rule that permits COM access only in certain namespaces. That doesn't exist yet but I have had in mind to write one in future.

Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
bclothier: This is an pre-existing thing, and I think it's best that this is addressed in its own PR - I'm sure it's not just the commands that does this -- we have lot of inconsistencies in how we initialize the logger all over the place, I think.

Rubberduck.VBEEditor/Events/VbeEvents.cs
```diff
@@ -29,15 +30,14 @@ public static VBEEvents Initialize(IVBE vbe)
...
+            lock (_instanceLock)
             {
                 if (_instance == null)
                 {
                     return;
                 }
 
-                _instance.Dispose();
-                _instance = null;
+                _instance.TerminateInternal();
```
comintern: Doesn't `_instance` need to be set to `null` here?

bclothier: No, because as per the comments in the PR's OP, commands are transient, so we need to persist the `Terminated` state for those that get created after the `EventsTerminated` event has been fired. Those commands would never receive the event. Furthermore, note that the `VBEEvents` is basically a singleton and is used by CW as such -- we pass in the instance to the CW during the registration, among with the `vbe` and other instances. Therefore during the host application's lifetime, there will be only one `VBEEvents` instance that will provide a consistent answer to all the commands (and in short future, other subsystems) that need to know if it's safe to proceed or not.

comintern: The problem with that is that it breaks the class as a singleton and effectively makes it something more like a "single-use class". `Initialize` is static, and it depends on whether or not `_instance` is null - *anything* can call `Initialize(IVBE vbe)`.  If the class is _internally_ torn down in response to the `Terminate()` call it leaves it in a state where it can never be useful again until the assembly is unloaded.  If that is the intention, the lifespan and instancing should probably be re-thought entirely.

bclothier: For the record while I agree this needs to be reworked, I I think it's best in its own PR, as `IVbeEvents` isn't the only faux single-use singleton class that needs to be reworked and it would be better to handle all those fauxes together.

Rubberduck.Core/AutoComplete/ShowQuickInfoCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+            using (var commandBars = _vbe.CommandBars)
+            {
+                using (var command = commandBars.FindControl(showIntelliSenseId))
+                using (HookVbaBeep())
+                {
+                    command.Execute();
+
+                    // Ensures that the queued beep message (if any) is cleared
+                    // while we have it hooked
+                    _pumper.PumpMessages();
```
bclothier: Memo to myself - from discussion w/ Comintern, probably better off just having a long-lived hook instead of pumping.

MDoerner:
I am a bit confused here. Where exactly is it relevant that the commands are transient? As far as I know, they are all injected or constructed inside the constructor of a view model. I do not think that we use factories for them.

bclothier:
Commands are explicitly marked as transient during the registration for example here:

https://github.com/rubberduck-vba/Rubberduck/blob/9afe3bb26688ebbd107edd54e4e51a1cbe0ba9c9/Rubberduck.Main/Root/RubberduckIoCInstaller.cs#L733-L745

Other objects such as `ParentMenuItems` are similarly registered with same lifecycle.

MDoerner:
My point is that being registered as transient does not mean that the object gets created later than other objects or even once after the initial resolution of the object graph. All it means is that a separate object gets injected at each injection point.

So, if there is no automagic factory providing commands, all injected commands are there after the initial resolution.

Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        protected ComCommandBase(ILogger logger, IVbeEvents vbeEvents) : base(logger)
+        {
+            _vbeEvents = vbeEvents;
+        }
+        
+        protected override bool EvaluateCanExecute(object parameter)
+        {
+            return !_vbeEvents.Terminated && base.EvaluateCanExecute(parameter);
+        }
```
MDoerner: This will not work as intended because subclasses override this method.

Btw, this implementation is equivalent to `protected override bool EvaluateCanExecute(object parameter) => !_vbeEvents.Terminated`.

Rubberduck.Core/UI/Command/ComCommands/ComCommandBase.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public new void Execute(object parameter)
```
MDoerner: This will not work as intended because when cast to `CommandBase` the base class implementation will be used.

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -424,11 +426,13 @@ End Enum
             }
         }
 
+        // TODO: Remove the temporal copuling with other Arrange*
         private (Mock<IVBE> Vbe, RubberduckParserState State) ArrangeAndParseTestCode(ComponentType type, string name, string code)
         {
             return ArrangeAndParseTestCode(TestProjectName, new List<(ComponentType type, string name, string code)> {(type, name, code)});
         }
 
+        // TODO: Remove the temporal copuling with other Arrange*
```
MDoerner: typo - coupling

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -456,6 +460,57 @@ End Enum
             return (vbe, state);
         }
 
+        // TODO: Remove the temporal copuling with other Arrange*
```
MDoerner: dito

bclothier:
Apologies for the long delay. The changes requested should be incorporated. There were a number of tests that I had to revise or ignore due to the fact that the test setups were directly invoking the `Execute` method, which now indirectly invokes the `CanExecute` method. As such, it is not valid in all test setups (e.g. some test just passed in `null` parameter where the `CanExecute` prevents that). There should be no more changes needed to the PR but review will be appreciated to ensure everything is sane.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTestComponentCommand.cs
```diff
@@ -19,24 +21,37 @@ public class AddTestComponentCommand : AddTestModuleCommand
...
         {
             AddToCanExecuteEvaluation(SpecialEvaluateCanExecute);
         }
 
         private bool SpecialEvaluateCanExecute(object parameter)
         {
-            if (parameter == null || 
-                !ApplicableNodes.Contains(parameter.GetType()) ||
-                !(parameter is CodeExplorerItemViewModel node))
+            if (parameter == null)
```
Hosch250: Missing braces.

Rubberduck.Main/VbeProvider.cs
```diff
@@ -16,19 +17,23 @@ namespace Rubberduck
...
+            BeepInterceptor = beepInterceptor;
         }
 
         internal static void Terminate()
         {
             Vbe = null;
-            VbeRuntime = null;
+            VbeNativeApi = null;
+            BeepInterceptor?.Dispose();
+            BeepInterceptor = null;
```
Hosch250: I hope the beep interceptor is configured in transient scope.

Vogel612: It's configured as a Singleton, but that's just fine, considering the instance injected through CW is obained from here (and not the other way round)

RubberduckTests/Commands/ComCommandTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            return mockComCommand.Object;
+        }
+    }
+
+    internal static class ComCommandExtensions
+    {
+        internal static void VerifyOnExecute(this ComCommandBase comCommand, Times times)
+        {
+            var mock = Mock.Get(comCommand);
+            mock.Protected().Verify("OnExecute", times, ItExpr.IsAny<object>());
```
Vogel612: this is a bit of a hack, but considering that it's in a unit-test I'm very happy about having such a clean verification :)

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -50,7 +52,7 @@ public void AddsTest(Type command)
             var (vbe, state) = ArrangeAndParseTestCode(ComponentType.StandardModule, TestModuleBaseName, TestModuleHeader);
             using (state)
             {
-                var addTestMethodCommand = (ICommand)Activator.CreateInstance(command, vbe.Object, state, ArrangeCodeGenerator(vbe.Object, state));
+                var addTestMethodCommand = ArranageAddTestMethodCommand(command, vbe, state);
```
Vogel612: typo: `Arranage`

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -456,6 +461,57 @@ End Enum
...
+        }
+
+        // TODO: Remove the temporal copuling with other Arrange*
+        private AddTestModuleWithStubsCommand ArrangeAddTestModuleWithStubsCommand(Mock<IVBE> vbe, AddTestModuleCommand addTestModuleCommand, Mock<IVbeEvents> vbeEvents)
+        {
+            return new AddTestModuleWithStubsCommand(vbe.Object, addTestModuleCommand, vbeEvents.Object);
+        }
+
+        // TODO: Remove the temporal copuling with other Arrange*
+        private ICommand ArranageAddTestMethodCommand(Type command, Mock<IVBE> vbe, RubberduckParserState state)
```
Vogel612: see above

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -456,6 +461,57 @@ End Enum
...
+        {
+            return ArranageAddTestMethodCommand(command, vbe, state, ArrangeCodeGenerator(vbe.Object, state));
+        }
+
+        // TODO: Remove the temporal copuling with other Arrange*
+        private ICommand ArranageAddTestMethodCommand(Type command, Mock<IVBE> vbe, RubberduckParserState state,
```
Vogel612: dito

RubberduckTests/Commands/UnitTestCommandTests.cs
```diff
@@ -456,6 +461,57 @@ End Enum
...
+            ITestCodeGenerator testCodeGenerator)
+        {
+            return ArranageAddTestMethodCommand(command, vbe, state, testCodeGenerator, MockVbeEvents.CreateMockVbeEvents(vbe));
+        }
+
+        // TODO: Remove the temporal copuling with other Arrange*
+        private ICommand ArranageAddTestMethodCommand(Type command, Mock<IVBE> vbe, RubberduckParserState state,
```
Vogel612: dito

Rubberduck.Core/UI/CodeExplorer/Commands/CodeExplorerCommandBase.cs
```diff
@@ -1,13 +1,14 @@
...
 using System.Collections.Generic;
 using System.Linq;
-using Rubberduck.UI.Command;
+using Rubberduck.UI.Command.ComCommands;
+using Rubberduck.VBEditor.Events;
 
 namespace Rubberduck.UI.CodeExplorer.Commands
 {
-    public abstract class CodeExplorerCommandBase : CommandBase
+    public abstract class CodeExplorerCommandBase : ComCommandBase
```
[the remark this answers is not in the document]
bclothier: If I understand correctly, the `ExportCommand` in the referenced PR still does COM access, so it's still appropriate that it's derived from `CodeExplorerCommandBase` command. Note that "COM access" can be quite broad - simply taking a reference to a `IVBE` or the project repository is IMO a COM access and must be managed accordingly. 

Also, IIRC, I decided that all CE commands should be `ComCommand` because CE is a UI tool and it doesn't make sense to use it if everything is going down anyway.

Rubberduck.VBEEditor/Events/VbeEvents.cs
MDoerner: Could you remind me who calls `Terminate`? (I am currently very far from an IDE.) Depending on the answer, it might be a very bad idea not to set `_instance` to `null`. More precisely, if `Terminate` gets called on every add-in shutdown, not doing it breaks reloading RD. (RD does not get unloaded on deactivation via the dialog.)

Anyway, why should not setting it to `null` should be required for commands created after termination? CW already has a fixed instance it uses completely independent of the current value of `_instance`. It only queries `Initialize` on the next load of RD.

Rubberduck.Core/UI/Command/CommandBase.cs
```diff
@@ -53,10 +53,20 @@ public bool CanExecute(object parameter)
...
 
+        protected virtual bool RequireReEvaluationOnExecute => false;
+
         public void Execute(object parameter)
         {
             try
             {
+                if (RequireReEvaluationOnExecute)
+                {
+                    if (!CanExecute(parameter))
```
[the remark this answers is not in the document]
bclothier: I like that idea - in reality we only need to test the `Terminated` and not necessarily other conditions because it is possible to have that changed between the check and the execution, and we'd then get disconnection errors anyway. 

I will refactor the functions to have the optional bool parameter.

Rubberduck.VBEEditor/Events/VbeEvents.cs
bclothier: There are two places where it is called:

https://github.com/rubberduck-vba/Rubberduck/blob/b57c05ab52cf5a23ba084a89597117e01427aa19/Rubberduck.Parsing/VBA/RubberduckParserState.cs#L1109-L1130

https://github.com/rubberduck-vba/Rubberduck/blob/e7e27d409078fe68094d5bdb4cbdd366be6366fe/Rubberduck.VBEEditor/Events/VBENativeServices.cs#L41-L54

I need to check but IIRC, the VbeNativeService.UnhookEvents gets called first before RPS's `Dispose`. RPS's `Dispose` is mainly precautionary. 

I don't think CW's instance is independent because we inject it ourselves:
https://github.com/rubberduck-vba/Rubberduck/blob/4ded6497ce34fae399fa86bc7f0749f96cdc2204/Rubberduck.Main/Root/RubberduckIoCInstaller.cs#L843

The idea is that the `VBEEvents` needs to be linked to VBIDE's lifetime - if it's shutting down, then we need to show that. But VBIDE gives us precious little about its state besides the addin events via the `Extension` class from where we reach the `VbeNativeServices.UnhookEvents`:
https://github.com/rubberduck-vba/Rubberduck/blob/61fd569e76965342379caf2b896123739d6f0f4c/Rubberduck.Main/Extension.cs#L251-L281

As I commented to Comintern earlier, I am not happy with faux-singleton classes and I do want to fix them, but I really want to do this in its own PR. I will test again but I believe I verified that unloading & reloading works for RD.

bclothier: I'm glad that this was reviewed - I tested again and to my surprise, it did persist the state between loads, which causes the 2nd load to be non-functional. I've restored the `null` setting and revised the `_terminated` property instead, so that it is no longer single-use. This is not best as I still have in mind to improve the design of this class and other classes where their functionality are dependent on the VBIDE's lifetime.

Rubberduck.Core/UI/Command/CommandBase.cs
```diff
@@ -22,16 +22,42 @@ protected CommandBase(ILogger logger = null)
...
+        private void AddToCanExecuteEvaluation(Func<object, bool> furtherCanExecuteEvaluation)
+        {
+            var currentCanExecuteCondition = CanExecuteCondition;
+            CanExecuteCondition = (parameter) =>
+                currentCanExecuteCondition(parameter) && furtherCanExecuteEvaluation(parameter);
+        }
+
+        private void AddToOnExecuteEvaluation(Func<object, bool> furtherCanExecuteEvaluation)
+        {
+            if (OnExecuteCondition == null)
```
MDoerner: Why not simply set `OnExecuteEvaluation` to `() => true` in the constructor like for the `CanExecuteEvaluation`. That removes the need for all those `null` checks.

Rubberduck.Core/UI/Command/FormDesignerFindAllReferencesCommand.cs
```diff
@@ -10,7 +12,8 @@ public class FormDesignerFindAllReferencesCommand : CommandBase
     {
         private readonly FindAllReferencesCommand _findAllReferences;
 
-        public FormDesignerFindAllReferencesCommand(FindAllReferencesCommand findAllReferences)
+        public FormDesignerFindAllReferencesCommand(
+            FindAllReferencesCommand findAllReferences, IVbeEvents vbeEvents)
```
MDoerner: Why the line break?

Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestMethodCommand.cs
```diff
@@ -3,25 +3,29 @@
...
-    public class AddTestMethodCommand : CommandBase
+    public class AddTestMethodCommand : ComCommandBase
     {
         private readonly IVBE _vbe;
         private readonly RubberduckParserState _state;
         private readonly ITestCodeGenerator _codeGenerator;
 
-        public AddTestMethodCommand(IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator)
+        public AddTestMethodCommand(
+            IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator,
```
MDoerner: All on one or one on each

Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestMethodExpectedErrorCommand.cs
```diff
@@ -3,25 +3,29 @@
...
-    public class AddTestMethodExpectedErrorCommand : CommandBase
+    public class AddTestMethodExpectedErrorCommand : ComCommandBase
     {
         private readonly IVBE _vbe;
         private readonly RubberduckParserState _state;
         private readonly ITestCodeGenerator _codeGenerator;
 
-        public AddTestMethodExpectedErrorCommand(IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator)
+        public AddTestMethodExpectedErrorCommand(
+            IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator,
```
MDoerner: dito

Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestModuleCommand.cs
```diff
@@ -1,23 +1,27 @@
...
     [ComVisible(false)]
-    public class AddTestModuleCommand : CommandBase
+    public class AddTestModuleCommand : ComCommandBase
     {
         private readonly RubberduckParserState _state;
         private readonly ITestCodeGenerator _codeGenerator;
 
-        public AddTestModuleCommand(IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator)
+        public AddTestModuleCommand(
+            IVBE vbe, RubberduckParserState state, ITestCodeGenerator codeGenerator,
```
MDoerner: dito

Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestModuleWithStubsCommand.cs
```diff
@@ -1,24 +1,27 @@
...
 {
-    public class AddTestModuleWithStubsCommand : CommandBase
+    public class AddTestModuleWithStubsCommand : ComCommandBase
     {
         private readonly IVBE _vbe;
         private readonly AddTestModuleCommand _newUnitTestModuleCommand;
 
-        public AddTestModuleWithStubsCommand(IVBE vbe, AddTestModuleCommand newUnitTestModuleCommand) : base(LogManager.GetCurrentClassLogger())
+        public AddTestModuleWithStubsCommand(
+            IVBE vbe, AddTestModuleCommand newUnitTestModuleCommand, IVbeEvents vbeEvents)
```
MDoerner: dito

PR #4844, opened 2019-03-07
Implement DocumentModuleDeclaration as subclass of ClassModuleDeclaration

### LINKED ISSUE #4724: DeclarationType is not correct for documents
That was discovering while fixing #4708 -- for documents, they are given `DeclarationType.ClassModule`. That is because of the document falling into the `default` case:

https://github.com/rubberduck-vba/Rubberduck/blob/dda930087260194b2e3935fac1171143b6f5cdf4/Rubberduck.Parsing/VBA/DeclarationResolving/DeclarationResolveRunnerBase.cs#L186-L213

The fix is trivial; we can just add a `isDocument = false` to the `ClassModuleDeclaration` and set the declaration type accordingly. The fix does not break any unit tests which is a red flag that there is insufficient coverage so we may end up breaking more things than fixing. We will need to do a careful review of the code to determine the risk and pay this technical debt. There are likely callers out there that are unnecessarily checking both declaration types and component types because of that.

comintern:
This is basically just a stub ATM, and is intended to house document specific functionality (mainly to allow caching of properties).  

Closes #4724

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/HungarianNotationInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/HungarianNotationInspection.cs
@@ -85,6 +85,7 @@ namespace Rubberduck.Inspections.Concrete
             DeclarationType.Constant,
             DeclarationType.Control,
             DeclarationType.ClassModule,
+            DeclarationType.Document,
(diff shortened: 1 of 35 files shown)


comintern:
Closed to flatten.

PR #4857, opened 2019-03-15
czech language for v2019-03-xx

SonGokussj4:
Should be alright :-)


Vogel612:
Thank you very much :)

PR #4884, opened 2019-04-01
Fix line special form grammar

Vogel612:
Note that this doesn't enforce the same constraints as the VBE when parsing the lineSpecialForm. Most notably, the form `foo.Line -(x, y),` does not provoke a parser error, even though it does not compile.

The grammar should correctly deal with all cases that do compile.


Vogel612:
Since #4888 is now about the second part of the issue this references, this PR is now considered a fix for #4875 instead of only partially fixing it.

PR #4890, opened 2019-04-02
balanced settings headers

SonGokussj4:
Fixed balancing paddings, margins of the "Settings headers" (don't know how to specify it more). Here's a screenshot :) 

http://www.noirgoku.mablog.eu/screenshots/2019-04-02_235744.jpg


SonGokussj4:
Latest commit fixed the `#exclamation` text within Settings window so it shows some text. In this case, warning that the user doesn't have any modifiers for hotkeys. Which would be problematic.

[img from](https://i.imgur.com/VHbsrYa.png) --> [img result](https://i.imgur.com/m4tmgRW.png)

Issue #4628, opened 2018-12-09
Create HostApplicationModel in resources for storing stringified metadata


bclothier:
Originally from the discussion:

https://github.com/rubberduck-vba/Rubberduck/pull/4601#discussion_r240046688

It seems that with upcoming PR #4601, we will need metadata about host applications in multiple places in different projects. 

Though we already have a `IHostApplication`, this is a SCW, and we don't really need one just to get the file name, the allowable list of extensions and other stringified metadata that should be kept uniform for all potential consumers. Thus, the idea is that we create a dumb POCO model that provides all the metadata and update the consumers to use that instead of hard-coding those metadata about host applications.

bclothier:
Ref #2799 and #4784 

The bugfix introduces a `HostAutoMacro` struct that provides a description of signatures that different hosts may treat specially. It is still attached to a `IHostApplication`, however. It is known that the rename refactoring has some code that checks the `auto_` naming. There may be other similar check scattered in the RD codebase which should be consolidated. 

The original reasoning behind attaching the structs to the `IHostApplication` was to remove the need to determine the host application using stringified name or something like that. Further thoughts may be needed to better architecture this, considering that the data in the `HostAutoMacro` don't require any SCWs.

PR #4896, opened 2019-04-03
Command hierarchy can execute conditions

MDoerner:
This PR does two things aroud commands.

First, it removes the need to pass a logger to the `CommandBase` contructor, but retains the possibility for the sake of the `DelegateCommand`.

Second, it changes how `EvaluateCanExecute` is handled to the way it is alrady done for commands deriving from `ReactorCommandBase`: derived classes can only add to the `CanEvaluateExecute` evaluation, but not overwrite. This allows to enforce the base classe's can evaluate contract.

This is still WIP because it breaks the `AddTemplateCommand` because this class does not honor its base class's contract; it takes a different type of parameter. This will need to be separated somehow.

I am aware that this PR conflicts heavily with #4833. So, feel free to merge that first. I will then clean up the conflicts. However, this PR might help to actually enforce the can evaluate condition ion subclasses of `ComCommandBase`.


Rubberduck.Core/UI/Command/CommandBase.cs
```diff
@@ -12,20 +12,33 @@ public abstract class CommandBase : ICommand
...
-            Logger = logger;
+            Logger = logger ?? LogManager.GetLogger(GetType().FullName);
+            CanExecuteEvaluation = (parameter => true);
         }
 
         protected ILogger Logger { get; }
-        protected virtual bool EvaluateCanExecute(object parameter) => true;
         protected abstract void OnExecute(object parameter);
 
+        protected Func<object, bool> CanExecuteEvaluation { get; private set; }
```
bclothier: This is a nitpick but the name doesn't sound right - it's asking if it can execute an evaluation, which isn't really what we mean. Maybe `EvaluateAllExecuteConditions`?

Rubberduck.Core/UI/CodeExplorer/Commands/AddComponentCommand.cs
```diff
@@ -38,10 +40,9 @@ protected override void OnExecute(object parameter)
             AddComponent(parameter as CodeExplorerItemViewModel);
         }
 
-        protected override bool EvaluateCanExecute(object parameter)
+        private bool SpecialEvaluateCanExecute(object parameter)
         {
-            if (!base.EvaluateCanExecute(parameter) || 
-                !(parameter is CodeExplorerItemViewModel node) ||
+            if (!(parameter is CodeExplorerItemViewModel node) ||
```
bclothier: I'm inclined to think that this class is doing too much. The primary reason for `AddTemplateCommand` to derive from this class was because that class already has the implementation for creating a VB component in the project. It probably now should be the most derived class (e.g. `AddStdModuleCommand`, `AddClassModuleCommand`, etc.) responsibility to filter for `parameter is CodeExplorerItemViewModel`, or at least push them further out with a new abstract class that provides this specific filtering behavior. 

OTOH, most of the implementations has little to do with the `ICommand` at all; that probably could be factored out into a specialized service for which the derived classes can request and use to create VBComponent as they need.

Rubberduck.Core/UI/Command/CommandBase.cs
MDoerner: I will change the name.

MDoerner:
I could not keep myself from adding some things not directly related to the main aim of this PR:

1. `IRewriteSessionTryRewrite` has been extracted to `IExecutableRewriteSession`, which extends `IRewriteSession`. This allows to pass an `IRewriteSession` around with the guarantee that it does not get executed. (Unless someone casts it.)
2. There is a symptomatic fix for #4897. This issue seems to be caused by the visual tree not being reliable.
3. Fixed false warning that a quick fix could not be applied for refactoring quickfixes. (`TryRewrite` returned false if nothing was to be rewritten. That is reversed now, i.e. it will return `true` if no rewriter has been checked out.)
4. Refactorings now communicate a failure to rewrite by throwing a `RewriteFailedException`, which is used by the `QuickFixProvider` to trigger the failure notification.

MDoerner:
I have extracted adding components out of the component `AddComponentCommend` and changed the version adding code with attributes to actually have a chance to work for VB6. However, I am not completely sure whether it works. (I cannot test this myself.)

Moreover, I made the templates actually work when the UI language is soemthing else than English.

Rubberduck.Core/UI/CodeExplorer/CodeExplorerAddComponentService.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            return projectName;
+        }
+
+        private string OptionCompareStatement()
+        {
+            using (var hostApp = _vbe.HostApplication())
+            {
+                return hostApp?.ApplicationName == "Access" 
+                    ? "Option Compare Database"
```
bclothier: Referencing #4628 -- this probably ought to be cleaned up when a PR is opened to address that issue.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
```diff
@@ -59,6 +66,25 @@ private bool SpecialEvaluateCanExecute(object parameter)
...
+            if (!(node is CodeExplorerItemViewModel)
+                || node.Declaration == null)
+            {
+                return false;
+            }
+
+            try
+            {
+                var project = node.Declaration.Project;
+                return AllowableProjectTypes.Contains(project.Type);
```
bclothier: Shouldn't the `Project` be wrapped in a `using`?

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
```diff
@@ -59,6 +66,25 @@ private bool SpecialEvaluateCanExecute(object parameter)
...
+            }
+            catch (COMException)
+            {
+                return false;
```
bclothier: I'm inclined to think that it should at least get logged to `Debug` or `Trace` rather than silently swallowing.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
```diff
@@ -4,11 +4,12 @@
 using System.Runtime.InteropServices;
 using Rubberduck.Navigation.CodeExplorer;
 using Rubberduck.Templates;
+using Rubberduck.UI.Command;
 using Rubberduck.VBEditor.SafeComWrappers;
 
 namespace Rubberduck.UI.CodeExplorer.Commands
 {
-    public class AddTemplateCommand : CodeExplorerCommandBase
+    public class AddTemplateCommand : CommandBase
```
bclothier: TBH, it feels a bit weird to make it a generic Command when it's a part of the code explorer. 

I think it can be made compatible with the `CodeExplorerCommandBase` by changing `ApplicableNodes` to something else and have it contain only one member for the tuple type. The additional conditions from the original `ApplicableNodes` (renamed to something else) can be then added instead.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
```diff
@@ -42,7 +42,8 @@ public AddTemplateCommand(ICodeExplorerAddComponentService addComponentService,
 
         public IEnumerable<ProjectType> AllowableProjectTypes => ApplicableProjectTypes;
 
-        public ComponentType ComponentType => ComponentType.Undefined;
+        //We need a valid component type to add the component in the first place. Then the module content gets overwritten.
+        public ComponentType ComponentType => ComponentType.ClassModule;
```
bclothier: I think we need to pass it from the resource or something; it might not necessarily be a class module in future. At least change the comment to a `TODO` or `FIXME` so it's marked.

MDoerner: For VBA, the `ComponentType` used to create the component is entirely irrelevant bacause changing the contents via the file will change it to the appropriate one.

However, I am not sure whether this works for VB6. It would be good if somebody who has the IDE could test whether this works for templates with a different component type.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
MDoerner: No, the project on declarations is ultimately managed by the projects repository. Disposing it will break things.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
MDoerner: Will log ad `Debug`.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
MDoerner: This will require some investigation with the debugger. After all the _Node_ in `ApplicableNodeTypes` is not really true. I will just have to find out the appropriate type and maybe change slightly how the check works.

Rubberduck.Core/UI/CodeExplorer/Commands/AddTemplateCommand.cs
MDoerner: I would like to revert my position. The entire `try` should go. Handling exceptions by logging them is already done in `CommandBase`.

Rubberduck.Core/UI/CodeExplorer/CodeExplorerAddComponentService.cs
```diff
@@ -0,0 +1,122 @@
...
+            var folderAnnotation = FolderAnnotation(node);
+            var optionCompare = OptionCompareStatement();
+
+            var basePrefixInModule = optionCompare != null
+                ? $"{folderAnnotation}{Environment.NewLine}{optionCompare}"
+                : folderAnnotation;
+
+            var prefixInModule = additionalPrefixInModule != null
+                ? $"{basePrefixInModule}{Environment.NewLine}{additionalPrefixInModule}"
+                : basePrefixInModule;
```
Vogel612: StringBuilder should be able to simplify this a bit...

```
var modulePrefix = new StringBuilder(folderAnnotation);
if (optionCompare != null) {
  modulePrefix.Append(Environment.NewLine).Append(optionCompare);
}
if (additionalPrefixInModule != null) {
  modulePrefix.Append(Environment.NewLine).Append(additionalPrefixInModule);
}
...
```

Rubberduck.Core/UI/CodeExplorer/CodeExplorerAddComponentService.cs
```diff
@@ -0,0 +1,122 @@
...
+                ? folder.FolderAttribute 
+                : $"'@Folder(\"{Folder(node)}\")";
+        }
+
+        private string Folder(CodeExplorerItemViewModel node)
+        {
+            var declaration = node?.Declaration;
+            if (declaration == null)
+            {
+                return ActiveProjectName();
```
Vogel612: Shouldn't this be `ActiveProjectFolder`?

Rubberduck.Core/UI/CodeExplorer/Commands/CodeExplorerFindAllReferencesCommand.cs
```diff
@@ -24,6 +24,15 @@ public CodeExplorerFindAllReferencesCommand(RubberduckParserState state, FindAll
...
             _state = state;
             _finder = finder;
+
+            AddToCanExecuteEvaluation(SpecialEvaluateCanExecute);
+        }
+
+        private bool SpecialEvaluateCanExecute(object parameter)
+        {
+            return ((ICodeExplorerNode)parameter).Declaration != null &&
+                   (!(parameter is CodeExplorerReferenceViewModel reference) || !reference.IsDimmed) &&
```
Vogel612: could we de-morgan this into `!(parameter is CodeExplorereReferenceViewModel reference && reference.IsDimmed)`? Or am I missing something because I can't think straight?

Vogel612: Also I think we should pull the parser-state check to an earlier position, that is the cheapest check and it probably applies a lot...

Rubberduck.Core/UI/CodeExplorer/CodeExplorerAddComponentService.cs
MDoerner: Yes, it should.

Rubberduck.Core/UI/CodeExplorer/Commands/CodeExplorerFindAllReferencesCommand.cs
MDoerner: TBH, I have blindly copied from the original `EvaluateCanExecute`.

I will adjust this.

PR #4914, opened 2019-04-13
Invert project order for RD.CodeAnalysis and make RD.SettingsProvider more useful

Vogel612:
This cleans up Settings and their loading behaviour a bit and moves most of the interfaces relating to it into Rubberduck.SettingsProvider. 
The Defaults are now loaded differently, registration is probably utterly broken, though.

Furthermore this extracts the CodeInspectionSettings defaults to RD.CodeAnalysis to make them accessible for the loader there.
It also inverts the project interdependence between RD.Core and RD.CodeAnalysis in preparation for a new large feature involving dependency graphs.

As it stands the Defaults need a bit of cleanup and revisiting of the loading mechanism they use. This should unify how settings work and make specialized SettingsProviders superfluous (or at least significantly easier to implement)


Rubberduck.SettingsProvider/ConfigurationServiceBase.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
+namespace Rubberduck.SettingsProvider
+{
+    public class ConfigurationServiceBase<T> : IConfigurationService<T>
+        where T : class, new()
+    {
+        protected readonly IPersistanceService<T> persister;
```
MDoerner: Protected members should start with a capital letter.

Vogel612:
This PR adresses the I/O stall observed in #4905

PR #4924, opened 2019-04-18
Remove dymanic calls from resolver

MDoerner:
This PR replaces the dynamic call in the binding stage of the references resolver.

This has two main advantages: it is easier to navigate the code when debugging (no need to find some code and run it through the debugger to follow the execution flow) and it is slightly faster.


MDoerner:
I just did a search for uses of `dynamic` adn I seem to have missed a few. So, WIP for now.

PR #4939, opened 2019-04-28
Hotfix to TodoExplorer 

bclothier:
From an email reporting, it was observed that decompile do not work on Access. Investigating, it's not really the decompile but in fact the reparse command being executed during the startup which also compiles the codebase, giving an appearance of not working. This is further confirmed by the fact that if a project has broken code, dialog will be shown asking to parse code anyway at the startup. We do not want to parse code during startup as this is usually too premature. This seemed to be caused by the change of `RefreshCommand` which is invoked during the load. The hot fix converts the RefreshCommand into a CommandBase, so it is property-injected rather than newed up directly.

### DIFF
--- a/Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
+++ b/Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
@@ -49,6 +49,12 @@ namespace Rubberduck.UI.ToDoItems
             _uiDispatcher = uiDispatcher;
             _state.StateChanged += HandleStateChanged;
 
+            RefreshCommand = new DelegateCommand(LogManager.GetCurrentClassLogger(),
+                _ =>
+                {
+                    _state.OnParseRequested(this);
+                },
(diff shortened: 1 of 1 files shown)


Rubberduck.Core/UI/ToDoItems/ToDoExplorerViewModel.cs
```diff
@@ -49,6 +49,12 @@ public sealed class ToDoExplorerViewModel : ViewModelBase, INavigateSelection, I
             _uiDispatcher = uiDispatcher;
             _state.StateChanged += HandleStateChanged;
 
+            RefreshCommand = new DelegateCommand(LogManager.GetCurrentClassLogger(),
+                _ =>
+                {
+                    _state.OnParseRequested(this);
+                },
+                _ => _state.IsDirty());
```
MDoerner: This is the wrong can execute condition. The command should not be activated in most parser states. Moreover, it should always be available in `Ready`, `Error`, `ResolverError` and `Pending`.

PR #4958, opened 2019-05-14
Newest Czech translations

SonGokussj4:
Added missing Czech translations for next release and fixed some existing typos/wording.


Rubberduck.Resources/Templates.cs.resx
```diff
@@ -127,18 +127,7 @@
...
-  MultiUse = -1  'True
-END
-Attribute VB_GlobalNameSpace = False
-Attribute VB_Creatable = False
-Attribute VB_PredeclaredId = True
-Attribute VB_Exposed = False
-Attribute VB_Ext_KEY = "Rubberduck", "Predeclared Class Module"
-
-Option Explicit
-'@PredeclaredId</value>
```
Vogel612: does the predeclared class module template work if you have your locale set to czech now?

SonGokussj4: If you mean:
 - Code Explorer
 - Add Modules - Template - Predeclared Class Module (.cls)

It shows czech without a problem.
image

But I noticed that it fails to compile afterward. But that's, I think, not a bug caused by me?

Upon clicking on the template, it adds a `Module1`  with

```
'@Folder("MatigueProject")
VERSION 1.0 CLASS
BEGIN
  MultiUse = -1  'True
End

Option Explicit
```

Which fails. The `VERSION 1.0 CLASS` is highlighted in RED, when compiling

```
Compile error:

Expected: end of statement
```

retailcoder: @SonGokussj4 I'm 99% certain an issue was opened for that bug, but I can't find it ATM. But no, it's not something you accidentally introduced - looks like we're adding the '@Folder` annotation in the "attributes pass" (exported) code, but it should only be added at the very top of the module off the "code pane pass" (in-editor) code.

retailcoder: That said given the `@PredeclaredId` annotation, I'm not convinced the "predeclared class" template has any good reason to stay anymore.

bclothier: One reason to keep it is to avoid the multiple passes that adding `@PredeclaredId` would require to actually apply the behavior; adding the template would only need to be parsed only once. Adding the `@PredeclaredId` would be more useful for an existing class module with code already added, so that is two good different ways to quickly whipping the codebase into shape.

PR #4967, opened 2019-05-21
Correct exporting settings file

### LINKED ISSUE #4966: Export command in Settings window doesn't export settings file
Version 2.4.1.24543
Host Version: 15.0.5023.1000
Host Executable: EXCEL.EXE

**Description**
Export option for settings window doesn't export settings.

**To Reproduce**
1. Display settings window from IDE Rubberduck>Settings
2. Click on `Export` in top right of window to choose directory and filename
3. Navigate to directory and there is no file

**Expected behavior**
Settings actually exported.

**Screenshots**
RD Export button not exporting

IvenBach:
Closes #4966. Previously was exporting to %appdata%\Rubberduck\ and overwriting
rubberduck.config. Now exports to chosen directory and uses chosen filename.

### DIFF
--- a/Rubberduck.SettingsProvider/Persistence/XmlPersistenceService.cs
+++ b/Rubberduck.SettingsProvider/Persistence/XmlPersistenceService.cs
@@ -61,9 +61,9 @@ namespace Rubberduck.SettingsProvider
                 }
             }
 
-            EnsurePathExists();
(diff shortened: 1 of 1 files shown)


IvenBach:
[Linking chat](https://chat.stackexchange.com/transcript/message/50382102#50382102) for minor edit still needed.

PR #5034, opened 2019-07-01
Changed InnoSetup graphics to use wildcards

mansellan:
Probably a long shot, but might fix the failing AV build?

PR #5036, opened 2019-07-02
Use version 5.6.1 of InnoSetup

mansellan:


Vogel612:
This supersedes #5024 and #5034

PR #5089, opened 2019-08-17
Resolve recursive and parameterless default member accesses

MDoerner:
This PR aims to enhance the resolver to actually resolve default member accesses due to Let coersion and to generate identifier references for intermediate default member accesses in recursive default member resolutions.

Moreover, this PR includes a few fixes to the resolver.

So far, the identifier references for the intermediate default member accesses are generated. So, if somebody was interested to see how that is incorporated, I would appreciate any feedback.

### DIFF
--- a/Rubberduck.Parsing/Binding/Bindings/DictionaryAccessDefaultBinding.cs
+++ b/Rubberduck.Parsing/Binding/Bindings/DictionaryAccessDefaultBinding.cs
@@ -62,15 +62,17 @@ namespace Rubberduck.Parsing.Binding
             return Resolve(_lExpression, _argumentList, _expression);
         }
 
-        private IBoundExpression Resolve(IBoundExpression lExpression, ArgumentList argumentList, ParserRuleContext expression)
+        private static IBoundExpression Resolve(IBoundExpression lExpression, ArgumentList argumentList, ParserRuleContext expression)
         {
-            if (lExpression.Classification == ExpressionClassification.ResolutionFailed)
+            if (lExpression.Classification == ExpressionClassification.ResolutionFailed
+                || !(expression is VBAParser.LExpressionContext lExpressionContext))
             {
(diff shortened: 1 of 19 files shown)


MDoerner:
The main features of the PR are in now, i.e. resolution of default member calls due to Let coercion and procedure coercion and references for intermediate default member accesses on recursive default member resolution. 

To make it possible to distinguish the different kinds/levels of default member calls, the `IdentifierReference` got a few new properties. There are now `IsIndexedDefaultMemberAccess` and `IsNonIndexedDefaultMemberAccess` to distinguish those accesses with an argument list present, and thus an indication that a default member is used` and those completely implicit. Moreover, there is the new `DefaultMemberRecursionDepth`, which indicates how many default member calls on the same expression were required to get to the default member referenced. (For technical reasons, default member accesses on array accesses count from the  depth of the array access.) thus information is also used in the `DeclarationFinder` to order enumerables of identifiers with the same selection; the last one is the final reference.

I would appreciate any review on this, especially since I am not too happy with how some things turned out. In particular, I am not too happy about the two new optional parameters on `IBindingContext` only used for the Let coercion.

Although the main part of this PR is done, I will still add a bit to it by fixing two other resolver bugs. Any improvements or additions of inspections will be part of later PRs.

MDoerner:
Now, I have fixed the two remaining resolver issues I wanted to deal with in this PR, the `MemberNotOnInterfaceInspection` not working for procedure arguments (and, as I had to realize, in a lot more places) and generating non-undeclared array declarations instead of undeclared non-array declarations for ReDim statements on arrays without a previous declaration.

I also took the liberty to strip the parentheses from the `AsTypeName` of functions and properties returning an array. This was inconsistent with variables and parameters, which never had the parentheses and we know anyway wether it retuns an array by looking at the `IsArray` flag.

Since I am done with this PR from my side, I would appreciate featback on this PR.

PR #5190, opened 2019-10-06
Resolve issue 5088

### LINKED ISSUE #5088: False negative "Variable is not used" inspection
Version 2.4.1.4819
Host Version: 16.0.4873.1000
Host Executable: MSACCESS.EXE

**Description**
I have a variable that's _assigned_ but never again referenced, but no inspection result.

**To Reproduce**
```
Public Function UpdateTableLinks() As Boolean

  Const CONNECTION_TYPE As String = "ODBC;"
  Const CONNECTION_DESCRIPTION As String = "Description=Development;"
  Const CONNECTION_DRIVER As String = "DRIVER=SQL Server Native Client 11.0;"
  Const CONNECTION_TRUSTED As String = "Trusted_Connection=Yes;"
  
  On Error GoTo ErrorHandler
  
  Dim databaseHandle As Database
  Set databaseHandle = CurrentDb

  Dim TDF As TableDef
  Dim DBLoc As Long
  LogManager.Log TraceLevel, "Updating Table Links"
  For Each TDF In CurrentDb.TableDefs
    If TDF.connect <> vbNullString And InStr(1, TDF.connect, "sharepoint") = 0 Then
      DBLoc = InStr(1, TDF.connect, "DATABASE=") - 1
      Dim connectionDatabase As String
      connectionDatabase = IIf(TempVars.Item("DevMode").Value, "DATABASE=Dev;", "DATABASE=Prod;")
      TDF.connect = Strings.Join(Array(CONNECTION_TYPE, CONNECTION_DESCRIPTION, CONNECTION_DRIVER, TempVars.Item("dbServer").Value, CONNECTION_TRUSTED, connectionDatabase), vbNullString)
      TDF.RefreshLink
    End If
  Next

  UpdateTableLinks = True
  
ExitPoint:
  Exit Function
  
ErrorHandler:
  LogManager.Log FatalLevel, "Error in UpdateTableLinks Num: " & Err.Number & " Description: " & Err.Description & " Table Name: " & TDF.Name
  LogManager.Log FatalLevel, "Aborting."
  UpdateTableLinks = False
  Resume ExitPoint
  
End Function
```

Does _not_ generate an `Assignment is not used` inspection on `databaseHandle`. However, in this MCVE, an inspection result _is_ generated:
```
Private Sub foo()
  Dim databaseHandle As Database
  Set databaseHandle = CurrentDb
End Sub
```

**Expected behavior**
Inspection results should be generated no matter how simple or complex the procedure is.

**Logfile**
_GitHub is still being obnoxious and not letting me upload logs. I'll add it later._

BZngr:
Resolves #5088.  The existing inspection looked for declarations with zero references, but did not evaluate the case(s) where all references were assignments only.  Also, did some refactoring of tests to reduce duplicate boilreplate code.

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
@@ -39,15 +39,24 @@ namespace Rubberduck.Inspections.Concrete
     /// </example>
     public sealed class VariableNotUsedInspection : InspectionBase
     {
+        /// <summary>
+        /// Inspection results for variables that are never referenced.
+        /// </summary>
+        /// <returns></returns>
         public VariableNotUsedInspection(RubberduckParserState state) : base(state) { }
 
+        /// <summary>
+        /// VariableNotUsedInspection override of InspectionBase.DoGetInspectionResults()
+        /// </summary>
+        /// <returns>Enumerable IInspectionResults</returns>
         protected override IEnumerable<IInspectionResult> DoGetInspectionResults()
         {
             var declarations = State.DeclarationFinder.UserDeclarations(DeclarationType.Variable)
                 .Where(declaration =>
                     !declaration.IsWithEvents
                     && !declaration.IsIgnoringInspectionResultFor(AnnotationName)
-                    && !declaration.References.Any());
+                    && (!declaration.References.Any() 
+                        || declaration.References.All(rf => rf.IsAssignment)));
 
             return declarations.Select(issue => 
                 new DeclarationInspectionResult(this,
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
@@ -55,8 +55,7 @@ namespace Rubberduck.Inspections.Concrete
                 .Where(declaration =>
                     !declaration.IsWithEvents
                     && !declaration.IsIgnoringInspectionResultFor(AnnotationName)
-                    && (!declaration.References.Any() 
-                        || declaration.References.All(rf => rf.IsAssignment)));
+                    && declaration.References.All(rf => rf.IsAssignment));
 
             return declarations.Select(issue => 
                 new DeclarationInspectionResult(this,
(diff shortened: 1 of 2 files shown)


Vogel612:
minor nit: the commit message could be more expressive, but that's reeeally nitpicky. Thanks for this :)

Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
```diff
@@ -39,15 +39,24 @@ namespace Rubberduck.Inspections.Concrete
...
+        /// <returns>Enumerable IInspectionResults</returns>
         protected override IEnumerable<IInspectionResult> DoGetInspectionResults()
         {
             var declarations = State.DeclarationFinder.UserDeclarations(DeclarationType.Variable)
                 .Where(declaration =>
                     !declaration.IsWithEvents
                     && !declaration.IsIgnoringInspectionResultFor(AnnotationName)
-                    && !declaration.References.Any());
+                    && (!declaration.References.Any() 
+                        || declaration.References.All(rf => rf.IsAssignment)));
```
MDoerner: The second condition alone should suffice. If there are no elements, `All` will always return `true`.

RubberduckTests/Inspections/VariableNotUsedInspectionTests.cs
```diff
@@ -93,79 +68,77 @@ End Sub
...
             {
 
                 var inspection = new VariableNotUsedInspection(state);
                 var inspectionResults = inspection.GetInspectionResults(CancellationToken.None);
 
-                Assert.AreEqual(0, inspectionResults.Count());
+                resultCount = inspectionResults.Count();
             }
+            return resultCount;
         }
```
MDoerner: Thanks for making this a bit more dry.
Could I ask you to further refactor this and base this test class on the recently introduced `InspectionTestBase`.

Rubberduck.CodeAnalysis/Inspections/Concrete/VariableNotUsedInspection.cs
BZngr: Wow...that result for `All` on an empty collection is really not intuitive to me...but you're right.  Done.

RubberduckTests/Inspections/VariableNotUsedInspectionTests.cs
BZngr: Done

PR #5193, opened 2019-10-06
ImplementedInterfaceMemberInspection evals Annotated-only and implemented interfaces

### LINKED ISSUE #5143: False negative `ImplementedInterfaceMember`
```
Version 2.4.1.4930
Host Version: 16.0.4873.1000
Host Executable: MSACCESS.EXE
```

**Description**
In converting the beginnings of a regular class to an Interface, I added the `'@Interface` annotation and compiled. I still have actual code instead of just method headers, and I do _not_ get the error.

**To Reproduce**
1. Create new class
2. Add `'@Interface` annotation
3. Add a `Public Property Get/Let` with actual code
4. See error, or, rather lack of error

**Expected behavior**
This situation should generate an inspection warning me to remove the actual code from my Interface

**Screenshots**
2019-09-11 14_28_39-Microsoft Visual Basic for Applications - ROIReport -  IRestDriver (Code)

2019-09-11 14_29_36-Code Inspections

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/3602221/RubberduckLog.txt)

**Additional context**
At this point, I do _not_ have any `Implementation`s of this `Interface`. However, I'd expect that the `'@Interface` annotation should be enough to trigger the inspection.

Note: the examples on the RD site of [how this works](http://rubberduckvba.com/Inspections/Details/ImplementedInterfaceMember) are reversed - the "No error" example should actually produce an error, while the "Error" example should not.

BZngr:
Resolves #5143.  `ImplementedInterfaceMemberInspection` now inspects class modules with the `@Interface` annotation even if the interface has yet to be implemented by another class.  Incorporates `InspectionTestsBase` into the test class.

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/ImplementedInterfaceMemberInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/ImplementedInterfaceMemberInspection.cs
@@ -7,11 +7,12 @@ using System.Linq;
 using Rubberduck.Parsing.Symbols;
 using Rubberduck.Inspections.Inspections.Extensions;
 using Rubberduck.Common;
+using System;
 
 namespace Rubberduck.Inspections.Concrete
 {
     /// <summary>
-    /// Identifies members of class modules that are used as interfaces, but that have a concrete implementation.
+    /// Identifies class modules that define an interface with one or more members containing a concrete implementation.
     /// </summary>
     /// <why>
     /// Interfaces provide an abstract, unified programmatic access to different objects; concrete implementations of their members should be in a separate module that 'Implements' the interface.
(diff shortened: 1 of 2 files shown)


Rubberduck.CodeAnalysis/Inspections/Concrete/ImplementedInterfaceMemberInspection.cs
```diff
@@ -43,7 +44,13 @@ public ImplementedInterfaceMemberInspection(Parsing.VBA.RubberduckParserState st
 
         protected override IEnumerable<IInspectionResult> DoGetInspectionResults()
         {
-            return State.DeclarationFinder.FindAllUserInterfaces()
+            var annotatedAsInterface = State.DeclarationFinder.Classes
+                .Where(cls => cls.Annotations.Any(an => an.Annotation.Name.Equals("Interface"))).Cast<ClassModuleDeclaration>();
```
Vogel612: The predicate here should be `an.Annotation is InterfaceAnnotation`. Also I wonder why the Cast is necessary... `DeclarationFinder.Classes` should preerrably already return `ClassModuleDeclaration`s, but currently doesn't. That seems to be something of an oversight, though @MDoerner might correct me there...

BZngr: Yes - definitely a better predicate. Done.

PR #5195, opened 2019-10-06
Extract Interface from Code Explorer 

Mrblackey:
Partially applies to #2702

Added Extract Interface refactoring to Code Explorer context menu.
I didn't really know how to write proper tests for it.

Moved `CanExecute `logic to `ExtractInterfaceRefactoring`, since it was needed in two separate command implementations.

I have noticed that ExtractInterface is already available in Project Explorer context menu.

Issue #2702, opened 2017-02-17
Extract Interface should be available on the Project Explorer / Code Explorer context menus


ThunderFrame:
Why should I need to open a class' codepane just to extract an interface?

Vogel612:
FWIW this seems like a bad idea, because as a user I can't reasonably judge what side-effects I might incur by extracting an interface there ... the point I'm trying to make is: how can I judge what I want to extract and what not?

ThunderFrame:
the Extract interface shows you (and lets you choose) the list of members to extract. If your existing class has a more than 40-50 lines of code, you're unlikely to be able to see more than 2-3 members in the codepane anyway.

Issue #4338, opened 2018-08-31
VB6: Code Explorer does not show Related Documents


mansellan:
Title says it all.

mansellan:
Might be best to fix this in conjuction with #3697.

comintern:
Can you check to see if this is still an issue after the CE refactor? I'm not sure I changed much that would effect this, but the relevant code also didn't strike me as being "wrong" either.

mansellan:
Will do. IIUC its because we no longer treat codeless components as declarations, so related documents and resource files don't get collected.

mansellan:
No, still borked.

mansellan:
Linking #4852 - component types icons in the CE are broken in general atm.

PR #5202, opened 2019-10-10
Extract Interface from Code Explorer

Mrblackey:
Replaces PR #5195 
Partially applies to #2702


RubberduckTests/CodeExplorer/CodeExplorerViewModelTests.cs
```diff
@@ -245,6 +245,31 @@ public bool AddUserDocument_CanExecuteBasedOnProjectType(ProjectType projectType
...
+        [Test]
+        [TestCase(ComponentType.ActiveXDesigner, ExpectedResult = true)]
+        [TestCase(ComponentType.ClassModule, ExpectedResult = true)]
+        [TestCase(ComponentType.ComComponent, ExpectedResult = true)]
+        [TestCase(ComponentType.DocObject, ExpectedResult = true)]
+        [TestCase(ComponentType.Document, ExpectedResult = true)]
+        [TestCase(ComponentType.MDIForm, ExpectedResult = true)]
+        [TestCase(ComponentType.PropPage, ExpectedResult = true)]
+        [TestCase(ComponentType.RelatedDocument, ExpectedResult = false, Ignore = "Module doesn't contain members")]
+        [TestCase(ComponentType.ResFile, ExpectedResult = false, Ignore = "Module doesn't contain members")]
```
Mrblackey: IDK if it's common to comment on my own commit... 😁

These test are ignored since they throw an exception. I can't tell how will that affect the actual CE in IDE. Will it prevent execution? Will it crash?

retailcoder: Make sure `Rubberduck.Deployment` is the solution's "StartUp Project" (should be bold): 

image

You can right-click the `Rubberduck.Deployment` project, select "Properties", and in the "Debug" tab you want "Launch" set to "Executable" and "Executable" set to "C:\Program Files (x86)\Microsoft Office\root\Office16\EXCEL.EXE" to build & launch Excel when you hit F5 in Visual Studio 😃

Mrblackey: Right... But how do I add Related Document or Res File? Aren't they avilable only for VB6 IDE?
These are the two test cases I lack.

retailcoder: What is the exception? Is it handled, can/should it be caught & logged? We wouldn't want to ignore tests if they are actually telling us that there's a problem somewhere...

Mrblackey: The exception is when the `MockedCodeExplorer` is trying to select the first member of `RelatedDocument` or `ResFile` component types, in order to perform the test:
```
using (var explorer = new MockedCodeExplorer(ProjectType.HostProject, componentType, @"Public Sub Foo():  MsgBox """":End Sub ")
                .ImplementExtractIntercaceCommand().SelectFirstModule())
```
Which in my mind could mean one of two things:
1. RD's support for `RelatedDocument` and `ResFile` is not full. (at least in the CE, they don't appear)
2. It is in fact intentionally impossible for the user to select `RelatedDocument` or `ResFile` in CE.

In both cases, the exception has nothing to do with the test itself, but with the setup. i.e. the test fails because it's unable to select the first module.
I think we can safely assume that the tests are fine, since the only thing we really are testing here is `.Where(item => ModuleTypes.Contains(item.DeclarationType))` provided with
```
private static readonly DeclarationType[] ModuleTypes =
        {
            DeclarationType.ClassModule,
            DeclarationType.Document,
            DeclarationType.UserForm
        };
```
so `RelatedDocument` and `ResFile` are excluded...

Determining how to proceed with them in the CE is a matter for a different PR.

RubberduckTests/CodeExplorer/CodeExplorerViewModelTests.cs
```diff
@@ -245,6 +245,31 @@ public bool AddUserDocument_CanExecuteBasedOnProjectType(ProjectType projectType
...
+        [Category("Code Explorer")]
+        [Test]
+        [TestCase(ComponentType.ActiveXDesigner, ExpectedResult = true)]
+        [TestCase(ComponentType.ClassModule, ExpectedResult = true)]
+        [TestCase(ComponentType.ComComponent, ExpectedResult = true)]
+        [TestCase(ComponentType.DocObject, ExpectedResult = true)]
+        [TestCase(ComponentType.Document, ExpectedResult = true)]
+        [TestCase(ComponentType.MDIForm, ExpectedResult = true)]
+        [TestCase(ComponentType.PropPage, ExpectedResult = true)]
+        [TestCase(ComponentType.RelatedDocument, ExpectedResult = false, Ignore = "Module doesn't contain members")]
```
Mrblackey:
```suggestion
        [TestCase(ComponentType.RelatedDocument, ExpectedResult = false, Ignore = "Project doesn't contain selectable modules")]
```

RubberduckTests/CodeExplorer/CodeExplorerViewModelTests.cs
Mrblackey:
```suggestion
        [TestCase(ComponentType.ResFile, ExpectedResult = false, Ignore = "Project doesn't contain selectable modules")]
```

bclothier: Pinging @mansellan for opinions on the VB6 stuff above.

mansellan:
So for the time being, it's not an issue because [neither resources nor related documents are showing up in the VB6 Code Explorer](https://github.com/rubberduck-vba/Rubberduck/issues/4338). That said, you should probably check the component type in `CanExecute` and exclude both of these component types.

Rubberduck.Core/UI/Command/Refactorings/RefactorExtractInterfaceCommand.cs
```diff
@@ -32,41 +32,7 @@ private bool SpecializedEvaluateCanExecute(object parameter)
...
-            }
-
-            var parseTree = _state.GetParseTree(interfaceClass.QualifiedName.QualifiedModuleName);
-            var context = ((ParserRuleContext)parseTree).GetDescendents<VBAParser.ImplementsStmtContext>();
-
-            // true if active code pane is for a class/document/form module
-            return !context.Any()
-                   && !_state.IsNewOrModified(interfaceClass.QualifiedModuleName)
-                   && !_state.IsNewOrModified(activeSelection.Value.QualifiedName);
+            return ((ExtractInterfaceRefactoring)Refactoring).CanExecute(_state, activeSelection.Value.QualifiedName);
```
bclothier: Explicitly casting doesn't feel right here. 

I'd rather that it was on an interface or at least an abstract method on the `InteractiveRefactorBase`. This would require modifying other refactoring. If we don't want to address this in this PR, I would at least ask that a `TODO` be added to mark that this needs further development so we don't forget about this tech-debt.

Rubberduck.Core/UI/CodeExplorer/Commands/CodeExplorerExtractInterfaceCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+            AddToOnExecuteEvaluation(FurtherCanExecuteEvaluation);
+        }
+
+        public sealed override IEnumerable<Type> ApplicableNodeTypes => ApplicableNodes;
+
+        private bool SpecialEvaluateCanExecute(object parameter)
+        {
+            return _state.Status == ParserState.Ready &&
+                   parameter is CodeExplorerComponentViewModel node &&
+                   _refactoring.CanExecute((RubberduckParserState)_state, node.QualifiedSelection.Value.QualifiedName);
```
bclothier: Instead of casting, I'd rather that it used the interface or modify the class to depend on `RubberduckParserState` instead of the interface.

MDoerner: Yes, please do not cast an injected interface to an implementation. Which implementation we inject may change at any time and such a change is not expected to break anything.

Rubberduck.Refactorings/ExtractInterface/ExtractInterfaceRefactoring.cs
```diff
@@ -135,5 +136,36 @@ private string GetInterfaceModuleBody(ExtractInterfaceModel model)
             DeclarationType.Document,
             DeclarationType.UserForm
         };
+
+        public bool CanExecute(RubberduckParserState state, QualifiedModuleName qualifiedName)
```
MDoerner: Is this really a responsibility of the refactoring?

MDoerner:
@Mrblackey If you do not mind, I will pick this up and give it some finishing touches.

Mrblackey:
> @Mrblackey If you do not mind, I will pick this up and give it some finishing touches.

Sure! Feel free to do so, as I am a bit occupied with stuff at work lately.

PR #5207, opened 2019-10-11
Make PermissiveAssertClass more permissive

### LINKED ISSUE #4050: PermissiveAssertClass not permissive enough for tests
Setup
```
Dim Assert As Object
Set Assert = CreateObject("Rubberduck.PermissiveAssertClass")
Dim foo As Long
foo = 10
```

A test that has
```Assert.AreEqual "10", foo``` <-- Passes
```Assert.AreEqual 10, foo``` <-- Fails

I'd expect that anything that VBA considers as the same would pass: ```"10" = 10```, ```cint(10) = clng(10)```, ```cdbl(10) = 10```, etc...

I created a test in VS and got a passing result. I'm not understanding something or the `PermissiveAssertClass` needs a little work.
```
[Category("Unit Testing")]
[Test]
public void TestVBAIntAgainstLong()
{
    var assert = new PermissiveAssertClass();
    const short foo = 10;

    assert.AreEqual(foo, Convert.ToInt32(10));
}
```

### PR DESCRIPTION
**Pull request by bclothier**
Closes #4050 

Also unignore some of unit tests by providing a COM object.

### CODE REVIEW

File: RubberduckTests/VBEditor/Variants/VariantConverterTests.cs [hunk 1]
```diff
@@ -0,0 +1,456 @@
...
+        [TestCase(-1, typeof(DateTime), ExpectedResult = "1899-12-29 00:00:00")]
+        [TestCase(-657434.00001157413d, typeof(DateTime), ExpectedResult = "0100-01-01 00:00:01")]
+        [TestCase(-657434.000011574d, typeof(DateTime), ExpectedResult = "0100-01-01 00:00:01")]
+        [TestCase(2958465.999988426d, typeof(DateTime), ExpectedResult = "9999-12-31 23:59:59")]
+        [TestCase(2958465.99998843d, typeof(DateTime), ExpectedResult = "9999-12-31 23:59:59")]
+        [TestCase(0.5d, typeof(DateTime), ExpectedResult = "1899-12-30 12:00:00")]
+        [TestCase(-0.5d, typeof(DateTime), ExpectedResult = "1899-12-30 12:00:00")]
+        public string Test_ObjectConversion_SimpleValues_To_Date(object value, Type targetType)
+        {
+            var culture = new CultureInfo("en-US");
```
Vogel612: While the `en-US` culture gives a defined result, can we not pass `CultureInfo.InvariantCulture` here to achieve the same behaviour?

bclothier: I'm not sure that is an appropriate format to use. For one thing, we are using the overload that doesn't take a localization, so the results will be in `en-US` because that's how VBA swings. I do not think VBA has any concept of "invariant culture". 

I can't find a good entry in MS-VBAL; looked at:

https://docs.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/MS-VBAL/65097495-6498-414f-8af8-8e6f481f831f
https://docs.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/MS-VBAL/c86480b2-aef2-4488-b177-f55e13cc51f2

However, in VBA, the dates are pretty-printed to US formats,even if you enter it using non-US formats. So `someDate = #2019-10-31#` will end up as `someDate = #10/31/2019#`. For that reason, the test should behave the same way, right?

MDoerner: In .Net, the invariant culture is 'en-us` and we use it everywhere else in the code to prevent comparisons of VBA code from using a local culture.

bclothier: The [CultureInfo.InvariantCulture](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=netframework-4.8) does not suggest it's `en-us`, only that it's generally associated with English but no particular conventions. 

Furthermore, neither are perfect representative of the behavior in VBA. If we put in a date of `#13/4/2019#`, VBA tolerates it and interprets it as `2019-04-13`, whereas in .NET, neither invariant nor `en-US` will accept that date. Note that above, we have [test cases handling that problem](https://github.com/rubberduck-vba/Rubberduck/pull/5207/files/67e205e5cd2f78afa2dfbd2145a89f1f7607eee6#diff-e131315fbb415d2233e01b678439c46aR216-R218):

```
        [TestCase("04/12/2000", typeof(DateTime), ExpectedResult = "2000-04-12 00:00:00")]
        [TestCase("12/04/2000", typeof(DateTime), ExpectedResult = "2000-12-04 00:00:00")]
        [TestCase("13/04/2000", typeof(DateTime), ExpectedResult = "2000-04-13 00:00:00")]
```

Because we are parsing using the oleaut function, the cultureInfo doesn't actually contribute anything to the test case. I think I originally I needed it to parse the `DateTime` but that's not done here. So, I removed that. Thanks for pointing that out!

bclothier: Actually now I'm doubting myself again. The problem is that we are passing in `object value`, with dates as strings. Those dates are in US format and is passed to the `ChangeType` function. I think without any localization, it may use the current culture. Can anyone with non-US VBA install kindly confirm for me how VBA will interpret the 3 cases (and also `"04/13/2000"` as well) and tell me what is the output?

bclothier: In the test, it seems that this is sensitive to localization. With `fr-CA`, I can get same results that I would get with `en-US`, but not if it's `de-DE`. I assume that is due to the fact that in `fr-CA`, the default date format is `yyyy-mm-dd` (how sane!), whereas in `de-DE`, the default is `dd.mm.yyyy`, so that blurs the dates like `04/12/2000` and `04/13/2000`. Therefore, the date strings are locale-sensitive and tests need to reflect this. I put the `en-US` back, lacking any explicit documentation that `InvariantCulture` = `en-US` (it only says that it's generally English but does not get into details on how date formats are handled), I prefer to err on the side of explicit.

MDoerner: I have a simple question regarding the `VariantConverter`. Is it supposed to work the same way no matter the culture of the host?
In that case, its culture parameter is superfluous as every user will have to pass in the same culture.

bclothier: No. If you use the overload without the culture, it will take on the user's current locale settings and thus behave accordingly. 

However, if we want to enable users to write locale-neutral unit tests, they'll need to be able to pass in the LCID indirectly somehow (that'd be a separate PR, BTW) as to not deal with tests failing just because it was run on a different computer with different locale settings. Note that within VBA there is no way to explicitly pass in the LCID - one is at the mercy of the OS's regional setting.

PeterMTaylor: Will we end up having some inspections confirming/preventing users local settings behaviour affecting VBA experience or the other way around?

bclothier: We could create inspections to warn the users about locale-sensitive conversions like the example of converting strings into dates implicitly but that would be a separate issue still from the aforementioned suggestion of being able to specify the locale for the test's purpose. That suggestion is to address the current limitation of being unable to write any kind of localization tests because VBA does not provide a mechanism for explicitly specifying the locale short of changing the OS's Region & Language settings.

PR #5248, opened 2019-10-27
Resolve expression in Next of For loop

### LINKED ISSUE #5236: "Refactor | Rename" ignores "Next LoopCounter" statements and leaves uncompilable code
Version 2.4.1.5105
Host Version: 16.0.4888.1000
Host Executable: WINWORD.EXE (pretty sure it's host agnostic)

**Description**
Using the "Refactor | Rename" functionality on a loop variable that is used in any `Next counter` statement will rename the variable everywhere but in those `Next` statements, effectively breaking the code. See below for a brief example.

**To Reproduce**
``` vba
Option Explicit
' start here
Public Sub BeforeRefactorRename()
    Dim I As Long
    For I = 1 To 10
        Debug.Print I
    Next I
End Sub

' Rename: I --> TooFancyForALoopCounter
Public Sub RefactorRenameLeavesBrokenCode()
    Dim TooFancyForALoopCounter As Long
    For TooFancyForALoopCounter = 1 To 10
        Debug.Print TooFancyForALoopCounter
    Next I
End Sub
```

**Expected behavior**
The refactoring should also recognize and rename the variable in the `Next counter` statement.

### LINKED ISSUE #5252: Enhance ExpandDefaultMemberQuickFix to substitute hidden default member with equivalent member
**Justification**
Currently, we expand the hidden default member `[_Default]` for `Range`, which is not very helpful to the user. 

**Description**
Instead, for a clearly defined collection of combinations of default member and number of arguments, we should expand with the equivalent non-hidden member. 

**Additional context**
In case of `Range.[_Default]`, implicit calls are equivalent to `Range.Value` and indexed calls to `Range.Item`.

### LINKED ISSUE #5253: Code Inspection - Indexed default member access fix does not work correctly for default members with underscores
Version 2.4.1.5132
Host Version: 16.0.9126.2282
Host Executable: EXCEL.EXE

**Description**
Code Inspection - Indexed default member access fix does not work

Example sub prior to fixes being applied will trigger 3 Indexed default member access 

Sub test()
Columns("M:M").Select
Cells(4, 6).Value = "x"
Sheets("Sheet1").Range("A1").Value = Range("H1").Value  
End Sub

After applying the 3 fixes, all three lines will show a syntax error as it has added **._Default** :

Sub test()
Columns._Default("M:M").Select
Cells._Default(4, 6).Value = "x"
Sheets._Default("Sheet1").Range("A1").Value = Range("H1").Value
End Sub

**Expected behavior**
Not sure, but this does not work

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/3792842/RubberduckLog.txt)

MDoerner:
Closes #5236 
Closes #5253
Closes #5252

This PR fixes the oversight that we did not resolve the expressions after `Next` in For loops. For For Each loops everything was already in place.


RubberduckTests/Grammar/ResolverTests.cs
```diff
@@ -6858,6 +6858,115 @@ End Function
...
+        public void ExpressionInNextHasReference_DoubleFor()
+        {
+            var moduleCode = $@"
+Private Sub Foo()
+   Dim loopIndex As Long
+   Dim otherLoopIndex As Long
+   For loopIndex = 0 To 42
+      For otherLoopIndex = 0 To 23
+         'DoSomething
+   Next otherLoopIndex, loopIndex
```
retailcoder: :nauseated_face: this is criminal.... excellent test case!

Rubberduck.Parsing/VBA/ReferenceManagement/IdentifierReferenceResolver.cs
```diff
@@ -659,6 +659,11 @@ public void Resolve(VBAParser.ForNextStmtContext context)
             {
                 Resolve(stepStatement);
             }
+
+            for (var exprIndex = 2; exprIndex < context.expression().Length; exprIndex++)
+            {
+                ResolveDefault(context.expression()[exprIndex], false);
```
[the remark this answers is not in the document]
retailcoder: also, unclear why `exprIndex` needs to start at `2`... Wouldn't the index be dependent on whether there's a `Step` clause?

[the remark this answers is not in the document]
MDoerner: I can optimize that. I will do it as well for the for each, from where I copied it.

No, the step is in the step clause. Index 1 is the value after `To`.

MDoerner:
I took the liberty to also pack a fix for expanding default members with illegal characters into this PR.

RubberduckTests/Grammar/ResolverTests.cs
```diff
@@ -6858,6 +6858,115 @@ End Function
...
+                var elementReferences = element.References;
+
+                Assert.AreEqual(2, elementReferences.Count());
+            }
+        }
+
+        [Category("Grammar")]
+        [Category("Resolver")]
+        [Test]
+        public void ExpressionInNextHasReference_DoubleForEach()
```
bclothier: Do we need one more test to avoid ambiguity in those cases?

```
For i = 0 to 4
  For j = 0 to 23
    For k = 0 to 42
  Next j, k
Next i
```

as the variant where the final `Next` contains no reference to the `i`.

MDoerner:
Since this PR already contained the fix for the default member expansion for default members that need to be enclosed on brackets, I also added the enhancemant to map special default membert to other equivalent members based on the number of arguments provided.

To enable this for a default member in the `ExpandDefaultMemberQuickfix`, it has to be added to the dictionary of base overrides. Should a further distinction be made for specific numbers of arguments, these have to be added to the argument number based dictionary of overrides in addition.
For the `ExpandBangNotationQuickfix`, there is ony one dictionary, since there will alwys be exactly one argument.

Currently, there are only overrides for `Excel.Range._Default`, but more can be added easily.

RubberduckTests/Grammar/ResolverTests.cs
MDoerner: I could add such a test, but I do not see the added benefit. Actually, I also do not see any ambiguity here. 

Moreover, could you please clarify your last sentence? I have a really hard time interpreting it.

PR #5257, opened 2019-10-31
Update GeneralSettings header instructions

IvenBach:
Subclassed GeneralSettingsView to allow header instructions and button text to remain synchronized for further updates.

Instructions were a bit off previously with a recent PR.
General settings out of sync

### DIFF
--- a/Rubberduck.Resources/Settings/SettingsUI.resx
+++ b/Rubberduck.Resources/Settings/SettingsUI.resx
@@ -199,7 +199,7 @@
     <value>Configure which autocompletions are enabled.</value>
   </data>
   <data name="PageInstructions_GeneralSettings" xml:space="preserve">
-    <value>Click [Ok] to close the window and apply changes, or [Cancel] to discard them.</value>
+    <value>Click [{0}] to close the window and apply changes, or [{1}] to discard them.</value>
   </data>
   <data name="PageInstructions_IndenterSettings" xml:space="preserve">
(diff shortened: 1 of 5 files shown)


Rubberduck.Resources/Settings/SettingsUI.resx
```diff
@@ -199,7 +199,7 @@
     <value>Configure which autocompletions are enabled.</value>
   </data>
   <data name="PageInstructions_GeneralSettings" xml:space="preserve">
-    <value>Click [Ok] to close the window and apply changes, or [Cancel] to discard them.</value>
+    <value>Click [{0}] to close the window and apply changes, or [{1}] to discard them.</value>
```
bclothier: I like the implementation but I am wondering why would we need a "click x to do y" instruction that's not on the other pages. I mean, "click x to do y" is true for all pages, isn't it? So why are we calling it out only for the general page? Seems to me that the general page should be just "Configure how Rubberduck should behave, whether it should modify VBIDE and what hotkeys to use", which is more in line with other page instructions.

IvenBach: This was a direct update addressing the new text. I didn't think more than than when doing the update. Thinking it through [Instructions](https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.Core/UI/Settings/SettingsView.cs#L9) should be updated to a more semantic wording like `Description`.

PR #5274, opened 2019-11-09
Follow up on SC style import commands

### LINKED ISSUE #5293: Code Explorer Sync>Replace aborts silently
Version 2.4.1.20898
Host Version: 15.0.5023.1000
Host Executable: EXCEL.EXE

**Description**
When choosing Code Explorer>Context Menu>Sync Project>Replace contents from files... if there is a missing document the replace proceeds up until the missing document is encountered.

**To Reproduce**
1. Have a workbook with 3 sheets named Sheet1, Sheet2, Sheet3.
2. Sync Project>Replace contents - to show that it replaces everything
3. Clear out all code modules ensuring they are empty
4. Delete Sheet2
5. Sync Project>Replace contents - to show only those files listed before Sheet2 are replaced.

**Expected behavior**
The replacement should either continue importing everything thereafter or, as is done with Update components, should abort with a warning.

**Screenshots**
When sorting is done on name since Sheet2 does not exist files listed before are imported. Those after are not.
RD replace by name sort

When sorting is done on type only files listed before Sheet2 are imported.
RD replace by type sort

**Additional context**
Code to test with
Class1.cls
```
VERSION 1.0 CLASS
BEGIN
  MultiUse = -1  'True
END
Attribute VB_Name = "Class1"
Attribute VB_GlobalNameSpace = False
Attribute VB_Creatable = False
Attribute VB_PredeclaredId = False
Attribute VB_Exposed = False
Option Explicit

Public Sub Bazz()
    Debug.Print "Bazz was invoked"
End Sub
```

Module1.bas
```
Attribute VB_Name = "Module1"
Option Explicit

Public Sub Foo()
    Debug.Print "Foo was invoked"
End Sub

Public Sub InstantiateClassModuleAndInvoke()
    Dim bar As Class1
    Set bar = New Class1
    bar.Bazz
End Sub

Public Sub InvokeWorksheetCode()
    Sheet1.DocClsCode
End Sub
```

Sheet1.doccls
```
Option Explicit

Public Sub DocClsCode()
    Debug.Print "DocClsCode was invoked."
End Sub
```

Sheet3.doccls
```
Option Explicit

Public Sub Duk()
	Debug.Print "Duk was invoked"
End Sub
```

ThisWorkbook.doccls
```
Option Explicit

Public Sub WorkbookSub()
	Debug.Print "WorkbookSub was invoked"
End Sub
```
---

MDoerner:
Closes #5293

This PR extends the work done for the SC style import commands to handle the remaining VB6 component types.

In addition, the `UpdateFromFilesCommand` is enhanced to be more extensible. Now, we allow to import component types that require a binary file if and only if there is an implementation of `IRequiredBinaryFileNameExtractor` supporting the component type. The implementation of this interface (Currently there is only one for user forms.) provide the names of all required binary files. If these are not found in the same directory as the file to import, the import is aborted.  
There is an exception however for files for which we have a special import to existing components.Currently this applies to user forms.  

Since I have no VB6, I cannot investigate and later test how to extract the required binary names from VB6 types. Accordingly, somebody else will have to implement the corresponding extractors, if updating from file should be supported for VB6 component types with binary files as well.


mansellan:
I've not had a chance yet to fully review this PR, but it seems that the import assumes that certain component types must always provide binary files alongside the primary (declarations and code) file. This is not the case - binary files will only be present where there is a need for them based on the properties of their child controls. If a component does not have at least one child control which has a binary-content property set, a binary file will not be produced on export and will not be incuded on import.

I appreciate this makes life difficult for us - it may be that we need to parse the component declarations section more deeply in order to determine whether a binary file is referenced by the primary file.

MDoerner:
The assumption is not that some components always need binary files (except for user forms, which always have a binary, even when empty). The assumption is that there are components that might need binary files. These types are excluded from update unless an implementation of `IRequiredBinaryFileExtractor` exists that supports them.

mansellan:
Ah ok cool, that was my misunderstanding from a quick scan of the code in a web browser.

Rubberduck.CodeAnalysis/Inspections/Abstract/QuickFixBase.cs
```diff
@@ -3,6 +3,7 @@
 using System.Diagnostics;
 using System.Linq;
 using NLog;
+using Rubberduck.JunkDrawer.Extensions;
 using Rubberduck.Parsing.Inspections.Abstract;
 using Rubberduck.Parsing.Rewriter;
 using Rubberduck.Parsing.VBA.Extensions;
```
bclothier: If I understand correctly, we moved the `EnumerableExtensions` from `Rubberduck.Parsing.VBA.Extensions` to `Rubberduck.JunkDrawer.Extensions`.

Why isn't this removed? If it can't be removed because it uses other extensions, then why move only one extension and adding yet another `using`?

Same comment applies to several other files that only have the `using` added but not a corresponding removal.

MDoerner: I think the simple answer is: because the move refactoring has not removed the now obsolete using.

mansellan:
linking [chat](https://chat.stackexchange.com/transcript/message/52529208#52529208)

In theory, these commands shouldn't be necessary for VB6, as the source files, binary-content files and project files are already stored on the filesystem. Unfortunately, VB6 has no awareness of changes made externally. Not only does it not put filewatchers in place, it won't even check the filesystem for changes when compiling. The version in the editor is considered canonical at all times.

I will need to put a VB6 project under source control in git and simulate multi-user development.

mansellan:
Potential follow-on: #5282.

MDoerner:
I will extend this PR and move the validation logic for documents and user forms to the import command. However, I will not perform the exclusion based on possibly required and possibly not existing binary files for VB6 since we should not cripple the ability to import these just because we have not implemented a way to get the right binary names.

MDoerner:
Now, most of the checks introduces for `UpdateFromFilesCommand` are executed for all import commands. We check that the component names for the files are unique (Otherwise the import is non-deterministic since the VBE  numbers the later imported components and the order cannot be determined by the user.), we check that binary files are present (VBA only) and if they are not abort unless there is a different import path into existing modules (currently for user forms). In this case we check whether a corresponding component already exists and has the correct component type. If not, we abort. If yes, we import to the component by replacing the contents. The behaviour for documents is the same.

The `UpdateFromFilesCommand` executes the additional check that all files with existing modules have consistent component types based on the extension.

The `ReplaceFromFilesCommand` asks for final confirmation by the user before executing.

The final change since last week is that the replace command is no longer visible for VB6.

MDoerner:
I would appreciate if this PR could be reviewed rather sooner than later as it is the last PR for the release containing changes in the base resources.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -26,32 +32,43 @@ public class ImportCommand : CodeExplorerCommandBase
...
+            IEnumerable<IRequiredBinaryFilesFromFileNameExtractor> binaryFileExtractors,
+            IFileExistenceChecker fileExistenceChecker,
             IMessageBox messageBox)
             : base(vbeEvents)
         {
             _vbe = vbe;
             _dialogFactory = dialogFactory;
             _parseManager = parseManager;
+            _projectsProvider = projectsProvider;
+            DeclarationFinderProvider = declarationFinderProvider;
```
Vogel612: this sticks out like a sore thumb here. Can we move it down to the other protected and public properties?

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -106,6 +123,25 @@ private IVBProject TargetProjectFromVbe()
                 : null;
         }
 
+        private IDictionary<ComponentType, IRequiredBinaryFilesFromFileNameExtractor> BinaryFileExtractors(IEnumerable<IRequiredBinaryFilesFromFileNameExtractor> extractors)
+        {
+            var dict = new Dictionary<ComponentType, IRequiredBinaryFilesFromFileNameExtractor>();
+            foreach (var extractor in extractors)
+            {
+                foreach (var componentType in extractor.SupportedComponentTypes)
```
Vogel612: if an extractor can support multiple componentTypes, overlaps seem very much possible. This code does not account for that (or rather explicitly only uses the first extractor). A clarifying comment as to why that is sensible might be in order here.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -128,7 +164,8 @@ protected virtual ICollection<string> FilesToImport(object parameter)
 
                 var fileNames = dialog.FileNames;
                 var fileExtensions = fileNames.Select(Path.GetExtension);
-                if (fileExtensions.Any(fileExt => !_importableExtensions.Contains(fileExt)))
+                var importableExtensions = ImportableExtensions;
```
Vogel612: why the local reference copy?

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -142,31 +179,116 @@ protected virtual ICollection<string> FilesToImport(object parameter)
...
         {
-            var suspensionResult = _parseManager.OnSuspendParser(this, new[] {ParserState.Ready}, () => ImportFiles(filesToImport, targetProject));
-            if (suspensionResult != SuspensionResult.Completed)
+            var suspendResult = _parseManager.OnSuspendParser(this, new[] {ParserState.Ready}, () => ImportFiles(filesToImport, targetProject));
+            var suspendOutcome = suspendResult.Outcome;
+            if (suspendOutcome != SuspensionOutcome.Completed)
             {
+                if (suspendOutcome == SuspensionOutcome.UnexpectedError || suspendOutcome == SuspensionOutcome.Canceled)
+                {
+                    ExceptionDispatchInfo.Capture(suspendResult.EncounteredException).Throw();
```
Vogel612: this is really cool, but I had to look up the documentation for it. An explanatory comment along the lines of `// pretend the exception was thrown through this codepath instead of the parse manager` might be useful.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -180,6 +302,202 @@ protected virtual void ImportFiles(ICollection<string> filesToImport, IVBProject
...
+                .All(moduleNameGroup => moduleNameGroup.Count() == 1);
+        }
+
+        private void NotifyUserAboutAbortDueToDuplicateComponent(IDictionary<string, string> moduleNames)
+        {
+            var firstDuplicateModuleName = moduleNames
+                .GroupBy(kvp => kvp.Value)
+                .First(moduleNameGroup => moduleNameGroup.Count() > 1)
+                .Key;
+            var message = string.Format(RubberduckUI.ImportCommand_DuplicateModule, firstDuplicateModuleName);
```
Vogel612: this makes for a mediocre user experience. I'd want to know about all conflicts. That's something for a future PR though. A follow-up issue and a comment would be useful :)

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -180,6 +302,202 @@ protected virtual void ImportFiles(ICollection<string> filesToImport, IVBProject
...
+                }
+            }
+
+            return null;
+        }
+
+        private void NotifyUserAboutAbortDueToNonExistingDocument(ICollection<string> documentFiles, IDictionary<string, string> moduleNames, IDictionary<string, QualifiedModuleName> existingModules)
+        {
+            var firstNonExistingDocumentFilename = documentFiles.First(filename => !existingModules.ContainsKey(filename));
+            var firstNonExistingDocumentModuleName = moduleNames[firstNonExistingDocumentFilename];
```
Vogel612: same UX issue as above, same remedy :)

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -180,6 +302,202 @@ protected virtual void ImportFiles(ICollection<string> filesToImport, IVBProject
...
+            var message = string.Format(
+                RubberduckUI.ImportCommand_DocumentDoesNotExist,
+                firstNonExistingDocumentModuleName,
+                firstNonExistingDocumentFilename);
+            MessageBox.NotifyWarn(message, DialogsTitle);
+        }
+
+        private void NotifyUserAboutAbortDueToNonExistingBinaryFile(ICollection<string> filesWithoutBinary, IDictionary<string, ICollection<string>> missingBinaries)
+        {
+            var firstFilenameForFileWithoutBinaryAndComponent = filesWithoutBinary.First();
```
Vogel612: and again :)

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -180,6 +302,202 @@ protected virtual void ImportFiles(ICollection<string> filesToImport, IVBProject
...
+            var message = string.Format(
+                RubberduckUI.ImportCommand_BinaryDoesNotExist,
+                firstFilenameForFileWithoutBinaryAndComponent,
+                missingBinariesOfFirstFilenameWithoutBinaryAndComponent);
+            MessageBox.NotifyWarn(message, DialogsTitle);
+        }
+
+        private void NotifyUserAboutAbortDueToNonExistingBinaryFileAndComponent(ICollection<string> filesWithoutBinary, IDictionary<string, ICollection<string>> missingBinaries, IDictionary<string, string> moduleNames, IDictionary<string, QualifiedModuleName> existingModules)
+        {
+            var firstFilenameForFileWithoutBinaryAndComponent = filesWithoutBinary.First(filename => !existingModules.ContainsKey(filename));
```
Vogel612: *~signpost* :black_flag:

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -204,6 +522,37 @@ protected override void OnExecute(object parameter)
             }
         }
 
-        protected IDictionary<string, ComponentType> ComponentTypeForExtension { get; }
+
+        //We only allow extensions to be imported for which we might be able to determine that the conditions are met to actually import the file.
+        //The exception are specif exceptions to the rule.
```
Vogel612: that line is a bit redundant, maybe reformulating around "There are specific exceptions to this rule stored in AlwaysImportableExtensions"

Rubberduck.Parsing/VBA/IParseManager.cs
```diff
@@ -31,4 +31,55 @@ public interface IParseManager : IParserStatusProvider
...
+    {
+        public SuspensionResult(SuspensionOutcome outcome, Exception encounteredException = null)
+        {
+            Outcome = outcome;
+            EncounteredException = encounteredException;
+        }
+
+        public SuspensionOutcome Outcome { get; }
+        public Exception EncounteredException { get; }
+    }
```
Vogel612: :+1: love these

Rubberduck.Resources/RubberduckUI.de.resx
```diff
@@ -1532,13 +1532,18 @@ Import abgebrochen.</value>
...
-  <data name="UpdateFromFilesCommand_DocumentDoesNotExist" xml:space="preserve">
+  <data name="ImportCommand_DocumentDoesNotExist" xml:space="preserve">
     <value>Es existiert kein Dokumentenmodul namens '{0}', s.d. die Datei '{1}' nicht importiert werden kann.
 
 Import abgebrochen.</value>
   </data>
-  <data name="UpdateFromFilesCommand_UserFormDoesNotExist" xml:space="preserve">
-    <value>Es existiert keine Userform namens '{0}', s.d. kein Codebehind einer Userform aus der Datei '{1}' überschrieben werden kann.
+  <data name="ImportCommand_BinaryAndComponentDoNotExist" xml:space="preserve">
+    <value>Für die Datei '{0}' fehlen die benötigte(n) Binary-Datei(en) '{1}' und die Zielkomponente '{2}'existiert noch nicht.
```
Vogel612: Binary -> Binär

Rubberduck.VBEEditor/Utility/FileExistenceChecker.cs
```diff
@@ -0,0 +1,17 @@
...
+        bool FileExists(string filename);
+    }
+
+    public class FileExistenceChecker : IFileExistenceChecker
+    {
+        public bool FileExists(string filename)
+        {
+            return File.Exists(filename);
+        }
+    }
```
Vogel612: feels a bit overkill to me, TBQH, but if it makes unit-testing easier, I'm all for it

mansellan:
LGTM

Rubberduck.VBEEditor/Utility/FileExistenceChecker.cs
MDoerner: We can later grow this to a more useful abstraction of the file system. Without this, testing the behaviour regarding missing corresponding binaries is simply impossible.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
MDoerner: I think this was used more than once before. 
BTW, it is actually not a reference copy since the property returns a new collection each time.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
MDoerner: I changed it to use all extractors. This should not be too much of a performance problem.

Rubberduck.Resources/RubberduckUI.de.resx
MDoerner: Danke

Rubberduck.Core/UI/CodeExplorer/Commands/ReplaceProjectContentsFromFilesCommand.cs
```diff
@@ -16,21 +20,28 @@ public class ReplaceProjectContentsFromFilesCommand : ImportCommand
...
-            }
+            return DeclarationFinderProvider.DeclarationFinder
+                .UserDeclarations(DeclarationType.Module)
+                .Select(decl => decl.QualifiedModuleName)
+                .ToHashSet();
+        }
 
-            RemoveReimportableComponents(targetProject);
-            base.ImportFiles(filesToImport, targetProject);
+        protected override bool UserDeniesExecution(IVBProject targetProject)
```
bclothier: The usage of `Denies` doesn't flow here. Did you mean `Cancelled` or maybe `Declines`?

Rubberduck.Core/UI/CodeExplorer/Commands/UpdateFromFileCommand.cs
```diff
@@ -30,194 +27,31 @@ public class UpdateFromFilesCommand : ImportCommand
...
-                    {
-                        //We have to dispose the return value.
-                        using (components.Import(filename)) { }
-                    }
-                }
-            }
-        }
-
-        private Dictionary<string, string> ModuleNames(ICollection<string> filenames)
+        protected override bool ExistingModulesAreGenerallyOk(IDictionary<string, QualifiedModuleName> existingModules)
```
bclothier: could we rename to `ExistingModulesMayBeImported` or maybe `ExistingModulesAreEligible`?

Rubberduck.Parsing/Rewriter/AttributesRewriteSession.cs
```diff
@@ -32,9 +32,13 @@ protected override bool TryRewriteInternal()
             PrimeOpenStateRecovery();
 
             var result = _parseManager.OnSuspendParser(this, new[] {ParserState.Ready, ParserState.ResolvedDeclarations}, ExecuteAllRewriters);
-            if(result != SuspensionResult.Completed)
+            if(result.Outcome != SuspensionOutcome.Completed)
             {
                 Logger.Warn($"Rewriting attribute modules did not succeed. suspension result = {result}");
```
bclothier: The `s` in the `Suspension` should be capitalized.

Rubberduck.Parsing/VBA/IParseManager.cs
```diff
@@ -31,4 +31,55 @@ public interface IParseManager : IParserStatusProvider
         SuspensionResult OnSuspendParser(object requestor, IEnumerable<ParserState> allowedRunStates, Action busyAction, int millisecondsTimeout = -1);
         void MarkAsModified(QualifiedModuleName module);
     }
+
+
+
```
bclothier: Why the 3 line whitespace?

Rubberduck.Parsing/VBA/IParseManager.cs
```diff
@@ -31,4 +31,55 @@ public interface IParseManager : IParserStatusProvider
...
+    public enum SuspensionOutcome
```
bclothier: I wonder if we need a separate flag for where the suspension itself failed, as opposed to the underlying action failing?

Rubberduck.Parsing/VBA/ParseCoordinator.cs
```diff
@@ -142,16 +143,23 @@ public void SuspendRequested(object sender, RubberduckStatusSuspendParserEventAr
...
+            catch (OperationCanceledException ex)
             {
-                e.Result = SuspensionResult.UnexpectedError;
+                e.Result = SuspensionOutcome.Canceled;
+                e.EncounteredException = ex;
+                throw;
+            }
+            catch (Exception ex)
+            {
+                e.Result = SuspensionOutcome.UnexpectedError;
```
bclothier: As noted earlier in my previous comment regarding the enum, the `UnexpectedError` would be returned even if the exception originated from other place besides the `e.BusyAction.Invoke()` on the line 151. Does it make sense to separate the exception thrown in other places from the one that we may get as a result of `e.BusyAction.Invoke()`?

Rubberduck.Resources/RubberduckUI.de.resx
```diff
@@ -1532,13 +1532,18 @@ Import abgebrochen.</value>
...
-    <value>Es existiert keine Userform namens '{0}', s.d. kein Codebehind einer Userform aus der Datei '{1}' überschrieben werden kann.
+  <data name="ImportCommand_BinaryAndComponentDoNotExist" xml:space="preserve">
+    <value>Für die Datei '{0}' fehlen die benötigte(n) Binärdatei(en) '{1}' und die Zielkomponente '{2}' existiert noch nicht.
+
+Import abgebrochen.</value>
+  </data>
+  <data name="ImportCommand_BinaryDoesNotExist" xml:space="preserve">
+    <value>Für die Datei '{0}' fehlen die benötigte(n) Binärdatei(en) '{1}'.
 
 Import abgebrochen.</value>
```
bclothier: Ought the resource contain line breaks?  (Note: my comment applies to other resource keys with line breaks; I will not repeat it for those)

Rubberduck.VBEEditor/Extensions/ComponentTypeExtensions.cs
```diff
@@ -1,76 +1,103 @@
...
 
-        //TODO: double check whether the guesses below are correct.
+        //VB6 only
         public const string UserControlExtension = ".ctl";
         public const string PropertyPageExtension = ".pag";
         public const string DocObjectExtension = ".dob";
+        public const string ActiveXDesignerExtension = ".dsr";
+        public const string ResourceExtension = ".res";
+
+        //Binary resources
```
bclothier: This might be pedantic but I'd call it `VB6 Binary resources` to remind that it's still VB6 only.

Rubberduck.VBEEditor/Utility/FileExistenceChecker.cs
bclothier: Just to be aware - there are already packages that provides IO abstraction (e.g. `System.IO.Abstraction`) -  we should be careful to not end up reinventing wheel and having to pay additional tech debt as result of reinvented wheel being not to the standard specs.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
```diff
@@ -186,9 +187,16 @@ private void NotifyUserAboutAbortDueToUnsupportedFileExtensions(IEnumerable<stri
...
+            var suspendOutcome = suspendResult.Outcome;
+            if (suspendOutcome != SuspensionOutcome.Completed)
             {
+                if (suspendOutcome == SuspensionOutcome.UnexpectedError || suspendOutcome == SuspensionOutcome.Canceled)
+                {
+                    ExceptionDispatchInfo.Capture(suspendResult.EncounteredException).Throw();
+                    return;
+                }
+
                 Logger.Warn("File import failed due to suspension failure.");
```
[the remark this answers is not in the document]
MDoerner: If there are still more places with the catch all, they should change. Generally, unexpected exceptions in the refactoring and cancellation are supposed to bubble up.

Rubberduck.Core/UI/CodeExplorer/Commands/ImportCommand.cs
MDoerner: I have added a todo for it.

Rubberduck.Core/UI/CodeExplorer/Commands/ReplaceProjectContentsFromFilesCommand.cs
MDoerner: I'll rename to _Declines_ then.

Rubberduck.Core/UI/CodeExplorer/Commands/UpdateFromFileCommand.cs
MDoerner: I do not think that any of those names would be right. It is more a `ExistingModulesAreNotInEligible`. That `True` is returned does not mean that they are OK, only that they are not deemed not OK.

Rubberduck.Parsing/VBA/IParseManager.cs
MDoerner: I'll add a separate flag for the failure around the read lock. I was honestly surprised to see that it used `UnexpectedError` although it is a known error condition. All other cases how the suspension itself can fail already have a separate outcome.

Rubberduck.Resources/RubberduckUI.de.resx
MDoerner: Yes, it is supposed to contain line breaks. They are necessary to get one blank line between the reason for the failure and the statement that the import is aborted in the message box.

Rubberduck.VBEEditor/Extensions/ComponentTypeExtensions.cs
MDoerner: `.frx` is not VB6 only. It is also used for the binary file of user forms.

Rubberduck.Core/UI/CodeExplorer/Commands/UpdateFromFileCommand.cs
bclothier: I see. The trouble is that `AreGenerallyOk` doesn't actually convey the reason why they are "OK". We also want to avoid the double negation as well. I think `MayBeImported` at least convey the possibility that they would be OK without necessarily saying it will import successfully. The other alternative I can think is to use words like `PreCheckPassed`, which would avoid suggesting that it's the final answer.

MDoerner:
Any additional comments?

PR #5342, opened 2020-01-07
Stop linefeed additions when refactoring

### LINKED ISSUE #5335: Move closer to usage adds carriage return to end of module when line continuation character is used
Version 2.5.0.16568
Host Version: 16.0.12325.20240
Host Executable: EXCEL.EXE

**Description**
As title says.

**To Reproduce**
1. Paste code into module
```
Option Explicit

Public Sub MVCE()
    Dim Foo1 As String, Foo2 As String, Foo3 As String, Foo4 As String, Foo5 As String, Foo6 As String
    
    Foo1 = "foo1" & _
            "after continuation"
    Foo2 = "foo2" & _
            "after continuation"
    Foo3 = "foo3" & _
            "after continuation"
    Foo4 = "foo4" & _
            "after continuation"
    Foo5 = "foo5" & _
            "after continuation"
    Foo6 = "foo6" & _
            "after continuation"
End Sub
```
2. Ensure no extra carriage returns are after `End Sub`.
3. Use Move Closer To Usage on one of the variables.
4. Place cursor at end of module and see that there is now extra carriage returns for every time a variable was moved.

**Expected behavior**
Not have the added carriage returns added.

IvenBach:
Close #5335. Any extra linefeeds at the end of a rewrite are removed.

We start with the code below. Note the inability to bookmark in the image.
```
Option Explicit

Public Sub MVCE()
    Dim Foo1 As String, Foo2 As String
    
    Foo1 = "foo1" & _
            "after continuation"
    Foo2 = "foo2" & _
            "after continuation"
End Sub

```
Last line is unable to be bookmarked
01 Last line is unable to be bookmarked

Performing a rename on `Foo2` to become `Foo22`.
```
Public Sub MVCE()
    Dim Foo1 As String, Foo22 As String
    
    Foo1 = "foo1" & _
            "after continuation"
    Foo22 = "foo2" & _
            "after continuation"
End Sub


```
What was the last line is now able to be bookmarked.
02 What was last line is now able to be bookmarked

That's because a linefeed was added shifting it down one.
03 Last line unable to be bookmarked has shifted down a line

For every refactoring performed the end of the file would ever shift downwards. This PR checks for those new lines and removes them.

This may be an issue when extra linefeeds are intentional and they are suddenly removed.


Rubberduck.Parsing/Rewriter/ModuleRewriter.cs
```diff
@@ -33,7 +33,14 @@ public void Rewrite()
                 return;
             }
 
-            var newCode = _rewriter.GetText();
+            var tentativeCode = _rewriter.GetText();
+
+            while (tentativeCode.EndsWith(Environment.NewLine))
+            {
+                tentativeCode = tentativeCode.Remove(tentativeCode.Length - 2);
```
Vogel612: Should be `- Environment.NewLine.Length`

Vogel612:
As it stands, this change is breaking some unit tests that were expecting a trailing newline to be present. I would also argue in favor of a module ending with a single newline to simplify editing it.

The unit tests need to be fixed before this can be merged, I don't particularly care about the newline at the end, but there is merit to having an EOL-EOF combination at the end of a module, if only because it's somewhat conventional to have that.

IvenBach:
I'm fine with an additional `NewLine` at the end. However when it kept adding them for every refactoring that felt wrong.

PR #5390, opened 2020-02-19
Remove the project SCW from ProjectDeclaration

### LINKED ISSUE #4095: Seemingly random Unicode characters in the output of Code Inspection details.
From the code inspection window click the "Copy to clipboard" icon at the top.
Paste in to a text editor that can handle Unicode characters.
For inspections that include the workbook name in the results I am seeing the junk characters after the workbook name.

EXAMPLES:
> Suggestion: Consider renaming variable 'vbCrLf' - (Fairly long workbook name version number.xlsb - [L«ÿ}TºU|Ñ) Projectname.modulename, line 89

> Warning: Member 'Value' was not found on the compile-time interface for type 'Range'. - (Fairly long workbook name version number.xlsb - [L­^PÊº£$'HòÅekËøù™s|÷Å) Projectname.modulename, line 286

Version 2.2.6738.40126
Host Version: 15.0.4997.1000
Host Executable: EXCEL.EXE

### LINKED ISSUE #4811: Odd characters in Code Explorer
Version 2.4.0.4533
Host Version: 16.0.4795.1000
Host Executable: EXCEL.EXE

**Description**
When an Excel file name is too long to be completely displayed in the Code Explorer, I'm seeing odd non-English characters in my English Excel installation

**To Reproduce**
1. Give an Excel file a long name
2. Open the Code Explorer in a pane that's not wide enough to display the whole name
3. Profit! (or not)
4. oh... See error

**Expected behavior**
One would expect an ellipse `...` when a long name is truncated.

**Screenshots**
2019-02-18 07_40_57-microsoft visual basic for applications - satisfaction summary master v3 31 - de

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/2875627/RubberduckLog.txt)

**Additional context**
Noted this a build or two ago, but never had a chance to report it.

### LINKED ISSUE #5206: Remove Declaration.Project
I propose to remove the member `Project` from `Declaration` and especially from `ProjectDeclaration`. This will have to consequence that `ProjectDisplayName` will also have to be removed from these.

The problem with this member is that it returns a stored `IVBProject` instance. As managing all references to declarations is near impossible, we hand a project from the `ProjectsRepository` to the declaration. Ufortunately, this means that the project will get released on the next parse and thus will be invalid. This is a RCW already released COM exception just waiting to happen. 

Because of the release issues, we should in general not store COM wrappers anywhere but in the `ProjectsRepository`, with the exception of the `initial IVBE` instance we pass in through DI.

### LINKED ISSUE #5292: Code Inspection's "Copy to Clipboard" doesn't
```
Version 2.4.1.5200
Host Version: 16.0.4873.1000
Host Executable: MSACCESS.EXE
```

**Description**
The `Copy to Clipboard` icon on the _Code Inspections_ window copies text to the clipboard, but it isn't the currently selected inspection, the current group of inspections or even all the inspections.

**To Reproduce**
1. Parse & Inspect a project with issues
2. Open Code Inspections window
3. Click `Copy to Clipboard` button
4. Paste text somewhere handy (Notepad, Code pane, email, whatever...)
5. Revel in the fact that you've got no issues found

**Expected behavior**
Ideally, the selected inspection should be copied. If no single inspection is selected, the currently expanded section(s) should be copied, if no section is expanded (or all sections are) all inspections should be copied.

**Screenshots**
My errors: the green arrow shows a selected inspection, click the `copy` button:
2019-11-20 08_55_10-Code Inspections
Results in:
```
Rubberduck Code Inspections - 11/20/2019 08:56:44 0 issues found.
--
Type | Project | Component | Issue | Line | Column
```
That was pasted directly into the browser window (FF) which is oddly different than what I see when I paste it into N++:
2019-11-20 08_57_36-_new 12 - Notepad++

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/3869615/RubberduckLog.txt)

**Additional context**
I'm pretty sure this worked not long ago - I rarely use this, but I recall seeing someone had pasted inspection results into an issue within the last few weeks.

MDoerner:
Closes #5206
Closes #4811
Closes #4095
Closes #5292

This PR removes the `IVbProject` SCW from the `ProjectDeclaration`, which is the last COM wrapper remaining on declarations. All users of the corresponding property were refactored to use an `IProjectdProvider` instead to get the project based on the project id. The same is true for `ProjectDisplayName` on the `ProjectDeclaration`. There is the exception of the `InspectionResultFormatter`, which now gets the document name fed as an additional constructor parameter.

When implementing the according changes, I realized that this implementation of `IExportable` and the one in the ToDo Explorer were not wired up correctly, which should be the root cause of #5292. After the adjustments, clipboard exports worked again as expected.

In addition, the implementations of `IVbProject.ProjectDisplayName` have been enhanced to never show characters from random memory and to show the correct document name in next to all cases. Now, we first try to get it from the filename. If that fails, moste likely because the document has never been saved, we try to use the caption, but with an enhanced handling of the case that triggers the VBE API bug. Finally, if we only get part of the name, we fall back to the `BuildFileName` without the extension.

### DIFF
--- a/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
+++ b/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
@@ -5,6 +5,7 @@ using Rubberduck.AddRemoveReferences;
 using Rubberduck.Navigation.Folders;
 using Rubberduck.Parsing.Symbols;
 using Rubberduck.Parsing.VBA;
+using Rubberduck.VBEditor.ComManagement;
 using Rubberduck.VBEditor.SafeComWrappers;
 using Rubberduck.VBEditor.SafeComWrappers.Abstract;
 
@@ -21,17 +22,20 @@ namespace Rubberduck.Navigation.CodeExplorer
         };
 
         private readonly IVBE _vbe;
+        private readonly IProjectsProvider _projectsProvider;
 
         public CodeExplorerProjectViewModel(
             Declaration project, 
             ref List<Declaration> declarations, 
             RubberduckParserState state, 
             IVBE vbe, 
+            IProjectsProvider projectsProvider,
             bool references = true) 
             : base(null, project)
         {
-            State = state;         
+            State = state;
             _vbe = vbe;
+            _projectsProvider = projectsProvider;
             ShowReferences = references;
 
             SetName();
@@ -58,7 +62,7 @@ namespace Rubberduck.Navigation.CodeExplorer
                     return base.FontWeight;
                 }
 
-                var project = State.ProjectsProvider.Project(Declaration.ProjectId);
+                var project = _projectsProvider.Project(Declaration.ProjectId);
                 if (project == null)
                 {
                     return base.FontWeight;
@@ -158,7 +162,7 @@ namespace Rubberduck.Navigation.CodeExplorer
                 return new List<ReferenceModel>();
             }
 
-            var project = State.ProjectsProvider.Project(Declaration.ProjectId);
+            var project = _projectsProvider.Project(Declaration.ProjectId);
             if (project == null)
             {
                 return new List<ReferenceModel>();
--- a/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
+++ b/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
@@ -1,5 +1,4 @@
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Linq;
 using System.Windows;
 using Rubberduck.AddRemoveReferences;
--- a/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
+++ b/Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
@@ -22,7 +22,13 @@ namespace Rubberduck.Navigation.CodeExplorer
 
         private readonly IVBE _vbe;
 
-        public CodeExplorerProjectViewModel(Declaration project, ref List<Declaration> declarations, RubberduckParserState state, IVBE vbe, bool references = true) : base(null, project)
+        public CodeExplorerProjectViewModel(
+            Declaration project, 
+            ref List<Declaration> declarations, 
+            RubberduckParserState state, 
+            IVBE vbe, 
+            bool references = true) 
+            : base(null, project)
         {
             State = state;         
             _vbe = vbe;
@@ -47,7 +53,13 @@ namespace Rubberduck.Navigation.CodeExplorer
         {
             get
             {
-                if (_vbe.Kind == VBEKind.Hosted || Declaration.Project == null)
+                if (_vbe.Kind == VBEKind.Hosted || Declaration == null)
+                {
+                    return base.FontWeight;
+                }
+
+                var project = State.ProjectsProvider.Project(Declaration.ProjectId);
+                if (project == null)
                 {
                     return base.FontWeight;
                 }
@@ -55,7 +67,9 @@ namespace Rubberduck.Navigation.CodeExplorer
                 using (var vbProjects = _vbe.VBProjects)
                 using (var startProject = vbProjects?.StartProject)
                 {
-                    return Declaration.Project.Equals(startProject) ? FontWeights.Bold : base.FontWeight;
+                    return project.Equals(startProject) 
+                        ? FontWeights.Bold 
+                        : base.FontWeight;
                 }
             }
         }
@@ -139,7 +153,12 @@ namespace Rubberduck.Navigation.CodeExplorer
 
         private List<ReferenceModel> GetProjectReferenceModels()
         {
-            var project = Declaration?.Project;
+            if (Declaration == null)
+            {
+                return new List<ReferenceModel>();
+            }
+
+            var project = State.ProjectsProvider.Project(Declaration.ProjectId);
             if (project == null)
             {
                 return new List<ReferenceModel>();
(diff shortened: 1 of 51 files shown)


MDoerner:
I forgot, this PR has a conflict with PR #5387. This will be easy to fix, but none of the two should be merged right after the other one.

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProject.cs
```diff
@@ -186,32 +185,116 @@ public string ProjectDisplayName
...
+            {
+                try
+                {
+                    if (ProjectId != activeProject.ProjectId)
+                    {
+                        vbe.ActiveVBProject = this;
+                    }
+
+                    var caption = mainWindow.Caption;
+                    if (caption.Length > 99)
```
bclothier: I think we should consider the fact that apparently we get more caption from Win32 API than we would from `mainWindow.Caption`. Using Spy++, I was able to get up to 159 characters where `mainWindow.Caption` only gets us 98 characters. To use Win32 API, we'd need to p/invoke [`GetWindowText`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowtexta).

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
```diff
@@ -1,10 +1,12 @@
 using System.Collections.Generic;
+using System.ComponentModel;
```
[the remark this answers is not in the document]
MDoerner: No, will remove again.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -187,32 +186,116 @@ public string ProjectDisplayName
...
+                    }
+                }
+
+                return _displayName;
+            }
+        }
+
+        private string DisplayNameFromFileName()
+        {
+            return Path.GetFileName(FileName);
```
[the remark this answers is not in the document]
MDoerner: Is this possible in VB6? I got the impression that it was intentional that the VB6 version of `FileName` does not have the try catch block.

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProject.cs
MDoerner: I kind of don't want to do that for something that is only a fallback anyway, which again has another fallback.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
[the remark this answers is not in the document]
mansellan: Will check later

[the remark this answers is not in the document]
mansellan: Confimed that it doesn't throw, it just returns null.

Rubberduck.Core/Navigation/CodeExplorer/CodeExplorerProjectViewModel.cs
```diff
@@ -47,15 +58,23 @@ public override FontWeight FontWeight
         {
             get
```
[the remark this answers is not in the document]
mansellan: Confirmed that the emboldening of startup project still doesn't work. It's pre-existing though, I'll raise a fresh issue for it.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -187,32 +186,116 @@ public string ProjectDisplayName
...
+
+        private string DisplayNameFromBuildFileName()
+        {
+            var pseudoDllName = Path.GetFileName(BuildFileName);
+            return pseudoDllName == null || pseudoDllName.Length <= 4 //Should not happen as the string should always end in .DLL.
+                ? string.Empty
+                : pseudoDllName.Substring(0, pseudoDllName.Length - 4);
+        }
+
+        private static readonly Regex CaptionProjectRegex = new Regex(@"^(?:[^-]+)(?:\s-\s)(?<project>.+)(?:\s-\s.*)?$");
```
mansellan: As discussed in chat, VB6 has the caption the other way around from VBA.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
```diff
@@ -187,32 +186,116 @@ public string ProjectDisplayName
...
+                {
+                    if (ProjectId != activeProject.ProjectId)
+                    {
+                        vbe.ActiveVBProject = this;
+                    }
+
+                    var caption = mainWindow.Caption;
+                    if (caption.Length > 99)
+                    {
+                        //The value returned will be truncated at character 99 and the rest is garbage due to a bug in the VBE API.
```
mansellan: As discussed, project name for VB6 seems limited to 40 chars so this *probably* won't apply (except perhaps in locales where the application name is lengthy?)

MDoerner:
@mansellan Could you have a look whether the latest change makes it work correctly for VB6?

Rubberduck.VBEditor.VBA/SafeComWrappers/VB/VBProject.cs
mansellan: Given that this is a fallback, I think truncation to 99 characters is reasonable. You'd have to have the CE crazy-wide to even see the difference...

mansellan:
Yep, VB6 now matches PE behaviour.

Rubberduck.VBEditor.VB6/SafeComWrappers/VB/VBProject.cs
mansellan: Fixed

PR #5497, opened 2020-05-15
Annotate Declaration Command

### LINKED ISSUE #4790: Surfacing Annotations

### LINKED ISSUE #5460: False Positive Inspection: 'TestMethod' was expected to have less arguments.
Version 2.5.0.5410
Host Version: 16.0.12430.20184
Host Executable: EXCEL.EXE

**Description**
The 'Code Inspections' pane reports a **False Positive** inspection result.
> Warning: The annotation 'TestMethod' was expected to have less arguments. 

Even if the following annotation is used
```
'@Ignore SuperfluousAnnotationArgument
```

**To Reproduce**
Steps to reproduce the behaviour:
1.  create a standard module (.bas)
1.  insert the code (below)
1.  refresh the 'Code Inspections' pane
1.  view the false positive result
1.  open the 'Test Explorer'
1.  the category field on the test _correctly_ shows: 'TheCategory'
1.  back in the  'Code Inspections' pane, select 'Ignore Once'
1.  refresh the 'Code Inspections' pane
1.  view the false positive result

**Example Code**
```
'@TestModule
'@Folder("Tests")

Option Explicit
Option Private Module

'@TestMethod("TheCategory")
Private Sub TheTestMethod()                        'TODO Rename test
    On Error GoTo TestFail
    
    'Arrange:

    'Act:

    'Assert:
    'Assert.Succeed

TestExit:
    'Exit Sub
TestFail:
    'Assert.Fail "Test raised an error: #" & Err.number & " - " & Err.Description
End Sub
```
**Expected behaviour**
Code inspection pane should not report an exception.

**Screenshot**
Screenshot

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/4479886/RubberduckLog.txt)

**Additional context**
n/a

MDoerner:
Closes #4790
Closes #5460

This PR introduces a refactoring to annotate declaration, including a UI.

I open this draft PR now to ask some questions on how to proceed especially regarding some points surrounding the UI design. 

Currently, the UI looks as follows:
AnnotateDeclaration_v1

There are a few problems:
 - I think that the `Add` and `Remove` buttons push the values control to the side lets the annotation combobox look too wide. However, shrinking it to the width of the arguments box also looks strange. 
 - I am not too happy about how the selected declaration is displayed.
 - The DemiBold labels seems to actually render as Bold.
 - The inspections are in a seemingly random order. I have found no nice way to handle this. The displayed values come from a value converter getting the localized names. Unfortunately, for some mysterious reason there does seem to be a way to specify a sort order on the displayed items in WPF. The only option to do this seems to be to sort in the view model. However, I would like to avoid this since the sort has to refresh when the display culture changes.
- The error image (the one we use in the rename dialog) cannot be seen because it is outside the bounds of the data grid. If somebody knows how to handle this, I would be grateful.

In addition, I have a question regarding the code pane context menu:
 - Should we add annotate module and annotate member commands and move the current `AnnotateDeclaration` as a _selected_ option into a submenu like _Indent_? 
 
Finally, since this is my first user control, I would like to ask whether the organization, in particular in the XAML, makes sense. I would appreciate any advice improving things.

I plan to add a parameterized CE command in which the annotation type is pre-selected based on the item in the _Annotate_ submenu to come. The visibility of the items in there will be determined by the type of the selected item.

One more possible enhancement is to add a preview box at the bottom. That might also open up the option to move the _Add_ and _Remove_ buttons below the arguments data grid. (Currently, that looks really wrong.) However, I think that would be work for a follow up PR.


A9G-Data-Droid:
I would move the Add\Remove buttons above the arguments box, tied to the upper right corner. Then allow the arguments to fill the fill width of the window. This way the combobox and the arguments box are the same, full, width.

With the new width of the data grid you can add some space for the error image to appear. Maybe pad the side until the error image is visible within the same element where the datagrid lives.

Rubberduck.Core/UI/Refactorings/AnnotateDeclaration/AnnotationArgumentViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+                    break;
+            }
+
+            return errors;
+        }
+
+        private bool ContainsNewline(string argumentText)
+        {
+            return argumentText.Contains('\n') || argumentText.Contains('\r');
+        }
```
BZngr, line 178: Question/Curious: Is this equivalent to `argumentText.Contains(Environment.Newline);` or are there other considerations?

Rubberduck.Resources/RubberduckUI.resx
```diff
@@ -1819,4 +1819,84 @@ Import aborted.</value>
     <value>The component '{0}' would have been affected by the refactoring, but its state in Rubberduck's memory is stale. Please refresh Rubberduck and try again.</value>
     <comment>{0} stale module</comment>
   </data>
+  <data name="ImplementInterface_TODO" xml:space="preserve">
+    <value>'TODO implement interface member</value>
```
BZngr, line 1823: This was recently moved to `Refactorings.resx`.  Looks like it is trying to sneak back into the `RubberduckUI.resx`.  :)

Rubberduck.Core/UI/Refactorings/AnnotateDeclaration/AnnotationArgumentViewModel.cs
retailcoder: It's not equivalent: matching `Environment.NewLine` would match `\r\n` but not `\r` and not `\n`; this matches `\r`, `\n`, and `\r\n`.

MDoerner: On windows, they should not be equivalent since `Environment.Newline == "\r\n"`. That would not detect a newline copied in from a Unix style formatted text file. However, the next validation would complain about the control character.

Rubberduck.Resources/RubberduckUI.resx
MDoerner: I'll remove it. That also explains why there was a merge conflict at all.

MDoerner:
I have moved the Add and Remove buttons as suggested and the result looks a lot better.
AnnotateDeclaration_v2

In addition, the inspection names are now sorted by localized name. I had to do this in the view model since WPF does not support sorting by converted items in a collection view. This means that a change in the UI culture is not reflected immediately in the sort order. However, this should not be a problem since the dialog is modal; there is no way to reach the settings dialog to change the culture.

In order to allow ignoring all inspections via the ignore annotation, I have removed the easter egg names for the ThunderCode inspections; these did not describe what they are about.

Furthermore, the CE now has an _Annotate_ sub-menu, that contains exactly those annotations applicable to the node, i.e. modules get those with target `Module`, members those with target `Member` and variables and constants those with target `Variable`. In addition, declarations that cannot have attributes do not show attribute annotations. Annotations that are already there and do not allow multiple are deactivated.  
AnnotateDeclaration_CE

I have also adjusted the `TextBoxErrorTemplate` to properly emulate the current behaviour of the rename dialog. However, I am not happy with the result. In a DataGrid, something that pops out of the cells themselves looks strange, overlaps into other cells and gets clipped at the grid border. (See the first screenshot.) 
The standard WPF behaviour is to display a red inner border, which is kind of ugly, thb. Does anybody have a suggestion to get a useful error marker?
In addition to the template itself, the DataGrid seems to have its own row template, which adds the red exclamation mark to the left. Should I try to remove it or leave it there?

One more little annoyance to me is that the text box for text arguments does not adjust its height to the cell height when there are multiple applicable argument types. Using `VerticalAlignment="Stretch"` on the containing `StackPanel` just aligns it to the top, which looks even worse.

Apart from adjusting the error template based on feedback, all I wanted to put into this initial PR for `AnnotateDeclaration` is done now.

Possible future follow ups are:
 - Add a preview box.
 - Add a checkbox to align the attribute value for attribute annotations.
 - Align the error handling in the rename and move to folder UIs.

Rubberduck.Refactorings/AnnotateDeclaration/AnnotateDeclarationModel.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using Rubberduck.Parsing.Annotations;
+using Rubberduck.Parsing.Symbols;
+
+namespace Rubberduck.Refactorings.AnnotateDeclaration
+{
+    //This is introduced in favor of a value tuple in order to ba able to bind to the components in XAML.
```
Vogel612, line 7: typo: "to be able".

MDoerner:
I have moved the code pane command into its own sub menu with a choice of annotating the selected declaration, selected member or selected module, similar to the `Indent` sub menu.
AnnotateDeclaration_ContextMenu


This also fixes that the display order of the context menu was wrong, for ages. (Since our menu items know their display position, we need separate ones for different parent menus.)

A9G-Data-Droid:
@MDoerner I like the little red exclamation points. For error\incomplete textboxen I generally change the background color. I like to use a yellow to highlight a text field that needs correcting. If there are different ways for the text to be wrong then it's important to have some sort of feedback to the user about the error. When you don't have room to give that feedback it's generally done in a tooltip.

MDoerner:
While waiting for reviews, I took the liberty to add the third of the possible future follow ups to this PR, aligning the error handling in the rename and move (to) folder dialogs. 

These now use the `TextBoxErrorStyle`, which means that they look the same but have error tool tips. 
In addition, there is now a warning when move folder will merge folders, for the dialog and the drag and drop. The user can still proceed by confirming the dialog.

MDoerner:
I have added some margin to the right of the bottom grid. I guess I forgot to add that after moving the buttons from its right to its top.

AnnotateDeclaration_v3

Rubberduck.Core/UI/CodeExplorer/Commands/AnnotateDeclarationCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            _annotateAction = annotateAction;
+            _failureNotifier = failureNotifier;
+            _userInteraction = userInteraction;
+            _state = state;
+
+            AddToCanExecuteEvaluation(SpecialEvaluateCanExecute);
+        }
+
+        public override IEnumerable<Type> ApplicableNodeTypes => new[] { typeof(System.ValueTuple<IAnnotation, ICodeExplorerNode>) };
```
Vogel612, line 46: This seems to be unused in the diff... it also honestly doesn't seem to make a whole lot of sense to me

PR #5502, opened 2020-05-23
AssignmentNotUsedInspection reduce false positives due to jump statements

### LINKED ISSUE #5456: False Positive `AssignmentNotUsed` inspection
```
Version 2.5.0.5410
Host Version: 16.0.4966.1000
Host Executable: EXCEL.EXE
```

**Description**
A variable assignment in an error handler is flagged as `Not Used`, even though that's the Function's "return" variable and it _is_ used in cleaning up the return.

**To Reproduce**
```
Public Function Foo() as Long
  Dim Sum as Long
  On Error Goto ErrorHandler
  Sum = 1/0
CleanExit:
  Foo = Sum
  Exit Function
ErrorHandler:
  Sum = -1 
  Resume CleanExit
End Function
```

**Expected behavior**
The `Sum = -1` assignment in the ErrorHandler should not be flagged as unused.

**Screenshots**
N/A

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/4428722/RubberduckLog.txt)

**Additional context**
This will probably require full CPA to fix.

BZngr:
Closes #5456

Modifies `IsResultReference(...)` to avoid flagging an unused assignment that precedes a `GoTo/Resume` statement that branches execution to a location in the procedure where the assignment could be used.


Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -98,7 +100,8 @@ public static List<IdentifierReference> UnusedAssignmentReferences(INode node)
 
         protected override bool IsResultReference(IdentifierReference reference, DeclarationFinder finder)
         {
-            return !IsAssignmentOfNothing(reference);
+            return !(IsAssignmentOfNothing(reference)
+                        || IsPotentiallyUsedViaResumeOrGoToExecutionBranch(reference, finder));
```
Vogel612, line 104: That name feels a bit long-winded... how about `IsPotentiallyUsedViaJump`?

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -108,6 +111,140 @@ private static bool IsAssignmentOfNothing(IdentifierReference reference)
...
+        /// Excludes Assignment references that meet the following conditions:
+        /// 1. Preceed a GoTo or Resume statement that branches execution to a line before the 
+        ///     assignment reference, and
+        /// 2. A non-assignment reference is present on a line that is:
+        ///     a) At or below the start of the execution branch, and 
+        ///     b) Above the next ExitStatement line (if one exists) or the end of the procedure
+        /// </remarks>
+        /// <param name="resultCandidate"></param>
+        /// <param name="finder"></param>
+        /// <returns></returns>
```
Vogel612, line 128:
```suggestion
```

I'd prefer omitting these empty documentation elements

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -108,6 +111,140 @@ private static bool IsAssignmentOfNothing(IdentifierReference reference)
...
+        private static bool GotoExecutionBranchPotentiallyUsesVariable(IdentifierReference resultCandidate, IEnumerable<(string, int)> labelIdLineNumberPairs)
+        {
+            var gotoCtxts = resultCandidate.ParentScoping.Context.GetDescendents<VBAParser.GoToStmtContext>()
+                .Where(gotoCtxt => gotoCtxt.Start.Line > resultCandidate.Context.Start.Line);
+
+            if (!gotoCtxts.Any()) { return false; }
+
+            var gotoStmt = GetFirstContextAfterLine(gotoCtxts, resultCandidate.Context.Start.Line);
+
+            if (gotoStmt == null) { return false; }
```
Vogel612, line 150: This could possibly be reformulated using a `Sort` and `FirstOrDefault` something like the following air-code

```
var gotoStmt = resultCandidate.ParentScoping.Context.GetDescendents<VBAParser.GoToStmtContext>()
  .Where(gotoCtxt => {
     var gotoStart = gotoCtxt.Start; var resultStart = resultCandidate.Context.Start;
     return gotoStart.Line > resultStart.Line || gotoStart.Line == resultStart.Line && gotoStart.Column > resultStart.Column
  })
  .OrderBy(ctx => ctx.Start.Line)
  .FirstOrDefault();
if (gotoStmt == default) { return false; }
```

But I'd understand if in the service of readability this were abandoned.

Note that I expanded the goto context check to account for same-line GoTo's to account for that edge-case.

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -108,6 +111,140 @@ private static bool IsAssignmentOfNothing(IdentifierReference reference)
...
+
+            var executionBranchLine = DetermineExecutionBranchLine(gotoStmt.expression().GetText(), labelIdLineNumberPairs);
+
+            return IsPotentiallyUsedAssignment(resultCandidate, executionBranchLine);
+        }
+
+        private static bool ResumeExecutionBranchPotentiallyUsesVariable(IdentifierReference resultCandidate, IEnumerable<(string IdentifierName, int Line)> labelIdLineNumberPairs)
+        {
+            var resumeStmtCtxts = resultCandidate.ParentScoping.Context.GetDescendents<VBAParser.ResumeStmtContext>()
+                .Where(jumpCtxt => jumpCtxt.Start.Line > resultCandidate.Context.Start.Line);
```
Vogel612, line 160: this has a similar context ordering problem as the goto check in that it doesn't account for multiple statements on a single line

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -108,6 +111,140 @@ private static bool IsAssignmentOfNothing(IdentifierReference reference)
...
+            if (!resumeStmtCtxts.Any()) { return false; }
+
+            var onErrorGotoStatements = resultCandidate.ParentScoping.Context.GetDescendents<VBAParser.OnErrorStmtContext>()
+                                .Where(errorStmtCtxt => !errorStmtCtxt.expression().GetText().Equals("0"))
+                                .ToDictionary(k => k.expression()?.GetText() ?? "0", v => v.Start.Line);
+
+            var errorHandlerLabelsAndLines = labelIdLineNumberPairs.Where(pair => onErrorGotoStatements.ContainsKey(pair.IdentifierName));
+            
+            //If the resultCandidate line preceeds all ErrorHandlers/Resume statements - it is not evaluated
+            if (errorHandlerLabelsAndLines.All(s => s.Line > resultCandidate.Context.Start.Line))
```
Vogel612, line 171: IIUC this is correct, because a line-label needs to be at the start of a line, can you confirm that?

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
```diff
@@ -108,6 +111,140 @@ private static bool IsAssignmentOfNothing(IdentifierReference reference)
...
+
+            int? executionBranchLine = null;
+
+            var expression = resumeStmt.expression()?.GetText();
+
+            //For Resume and Resume Next, expression() is null
+            if (string.IsNullOrEmpty(expression))
+            {
+                //Get info for the errorHandlerLabel above the Resume statement
+                (string IdentifierName, int Line)? errorHandlerLabel = labelIdLineNumberPairs
```
Vogel612, line 188: the `Line` isn't used subsequently... it might be better to perform a `Select` to remove it?

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
BZngr: Agreed - changed

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
BZngr: Removed

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
BZngr: Handling the same-line `GoTo` and `Resume` scenarios drove a fair amount of changes - all for the better.

TBH I have never used or come across the use of the ":" to place multiple statements on the same line (...kind of unbelievable).  Your initial comment/example had me a confused for a little while.  Thank you for spotting that missing use case.

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
BZngr: From MS-VBAL

> 5.4.1.1 Statement Labels 
>  
> statement-label-definition = LINE-START ((identifier-statement-label ":") / (line-numberlabel [":"] )) 

Admittedly, I often mis-interpret MS-VBAL, but I believe the use of the `LINE-START` terminal symbol establishes that the statement-label-definition begins at the start of a line.  And...I couldn't place them anywhere but at the start of a line in my Excel VBE.

Rubberduck.CodeAnalysis/Inspections/Concrete/AssignmentNotUsedInspection.cs
BZngr: Yes...refactored

PR #5540, opened 2020-07-11
Introduce ConflictDetectionSession

BZngr:
This PR consolidates conflict detection for proposed renames, relocations, or new-to-be-inserted declarations.  The `ConflictDetectionSession`  relieves the `DeclarationFinder` of rename conflict detection 'collateral duty'. 

The `ConflictDetectionSession` will generate a non-conflicting name, as part of the conflict evaluation, based on the proposed rename, proposed new Declaration, or proposed Declaration relocation/move.  The generated name is used for providing a non-conflict initial identifier to the user (e.g., `ExtractInterfaceRefactoring`), or generating non-user-defined identifiers (e.g., backing field identifiers in `EncapsulateFieldRefactoring`)

The ConflictDetectionSession provides conflict analysis for multiple Declaration identifier (and/or location changes) per session. When checking for conflicts, it takes into account identifiers previously proposed within the session.

The PR removes conflict detection code (and tests) from the `DeclarationFinder` and incorporates a `ConflictSession` where the `DeclarationFinder.FindNewDeclarationNameConflicts` was used within `RenameRefactoring` and `AssignedByValParameterMakeLocalCopyQuickFix`.  `ExtractInterfaceRefactoring` was also updated in order to evaluate the user-provided identifier for the new interface class module.

A PR to incorporate use of a `ConflictDetectionSession` in `EncapsulateFieldRefactoring` will follow once this PR has been merged.  Changes required by `EncapsulateFieldRefactoring` are fairly extensive - so it was excluded from this PR.


BZngr:
At the moment, the value added by this PR is is significantly limited by the fact that the two refactorings that would it benefit most are not ready to use it.    And, when they are, will probably drive non-trivial changes to the current implementation:

- `EncapsulateFieldRefactoring` : #5578 should precede this PR.
- `MoveMemberRefactoring`: is still on my local machine and will be refactored significantly based on the outcome of the `EncapsulateFieldRefactoring` PR.

PR #5547, opened 2020-07-24
Fench Translations

retailcoder:


retailcoder:
Includes commits of #5479

PR #5555, opened 2020-08-13
Clean-up blank lines left behind by EncapsulateFieldRefactoring and MoveCloserToUsage

BZngr:
The `EncapsulateFieldRefactoring` potentially leaves large blocks of blank lines in the scenarios when the user elects to encapsulate numerous fields by wrapping them in a UDT.

Residual newlines also occur in the `MoveCloserToUsage` refactoring and quickfix.  It is less noticeable since these actions delete declarations one at a time.  But, running the refactoring/quickfix against many declarations in the same module can potentially result in a large block of residual blank lines.

To accomplish the newline removals, this PR introduces an `IModuleRewriterExtensions` class in Refactorings.Common.  The purpose of the extension methods are to provide a convenient mechanism to clear residual newlines left behind by the removal of variable and member declarations.  

`IModuleRewrite.Remove` does not remove the `EndOfStmtContext` which can result in orphaned newlines.  The extension method related to variable declarations also correctly handles removal of a variable declaration list in the scenario where are all variables in the list are identified for removal. (calling `IModuleRewrite.Remove` on each variable in a list leaves the `VisibilityContext` behind)


Rubberduck.Refactorings/Common/IModuleRewriterExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            var fieldsByListContext = toRemove.Distinct()
+                                                .GroupBy(f => f.Context.GetAncestor<VBAParser.VariableListStmtContext>());
+
+            foreach (var fieldsGroup in fieldsByListContext)
+            {
+                var variables = fieldsGroup.Key.children.Where(ch => ch is VBAParser.VariableSubStmtContext);
+                if (variables.Count() == fieldsGroup.Count())
+                {
+                    if (fieldsGroup.First().ParentDeclaration.DeclarationType.HasFlag(DeclarationType.Module))
```
MDoerner, line 38: This is a second iteration of `fieldsGroup`. Accordingly, it should be materialized via `ToList` where it is defined.

Rubberduck.Refactorings/EncapsulateField/EncapsulateFieldRefactoring.cs
```diff
@@ -91,26 +90,26 @@ protected override EncapsulateFieldModel InitializeModel(Declaration target)
 
         protected override void RefactorImpl(EncapsulateFieldModel model)
         {
-            var refactorRewriteSession = new EncapsulateFieldRewriteSession(_rewritingManager.CheckOutCodePaneSession()) as IEncapsulateFieldRewriteSession;
+            var refactorRewriteSession = _rewritingManager.CheckOutCodePaneSession();
 
-            refactorRewriteSession = RefactorRewrite(model, refactorRewriteSession);
+            RefactorRewrite(model, refactorRewriteSession as IRewriteSession);
```
MDoerner, line 95: The cast seems to be redundant; the return type of `CheckOutCodePaneSession` extends `IRewrtiteSession`.
Or is there another overload taking an `IExecutableRewriteSession`?

Rubberduck.Refactorings/EncapsulateField/EncapsulationStrategies/ConvertFieldsToUDTMembers.cs
```diff
@@ -27,14 +18,11 @@ public ConvertFieldsToUDTMembers(IDeclarationFinderProvider declarationFinderPro
...
-        protected override void ModifyFields(IEncapsulateFieldRewriteSession refactorRewriteSession)
+        protected override void ModifyFields(IRewriteSession refactorRewriteSession)
         {
             var rewriter = refactorRewriteSession.CheckOutModuleRewriter(_targetQMN);
 
-            foreach (var field in SelectedFields)
-            {
-                refactorRewriteSession.Remove(field.Declaration, rewriter);
-            }
+            rewriter.RemoveVariables(SelectedFields.Select(f => f.Declaration).Cast<VariableDeclaration>());
```
MDoerner, line 25: I would add a line break for each LINQ call. However, that is a question of style.

Rubberduck.Refactorings/MoveCloserToUsage/MoveCloserToUsageRefactoringAction.cs
```diff
@@ -16,12 +17,17 @@ public MoveCloserToUsageRefactoringAction(IRewritingManager rewritingManager)
         protected override void Refactor(MoveCloserToUsageModel model, IRewriteSession rewriteSession)
         {
             var target = model.Target;
-            InsertNewDeclaration(target, rewriteSession);
-            RemoveOldDeclaration(target, rewriteSession);
-            UpdateQualifiedCalls(target, rewriteSession);
+            if (target is VariableDeclaration variable)
```
MDoerner, line 20: That should probably be inverted. The result is a guard clause and no need to use `return`.

Rubberduck.Refactorings/Common/IModuleRewriterExtensions.cs
BZngr: I understand the comment..though I am curious as how to best materialize the `List` since `fieldsGroup` is defined within a `foreach (var fieldsGroup in fieldsByListContext)` statement.  Currently, I've added `var fieldsToDelete = fieldsGroup.ToList();`  and use `fieldsToDelete` for the `Count()` and `First()` calls which (I believe) resolves the comment.

Is there another more preferred way in this scenario?...perhaps using `ToLookup`  to instantiate `fieldsByListContext` instead of `GroupBy`  resolves this as well without calling `ToList` on `fieldsGroup`?

Rubberduck.Refactorings/Common/IModuleRewriterExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+using Rubberduck.Parsing.Grammar;
+using Rubberduck.Parsing.Rewriter;
+using Rubberduck.Parsing.Symbols;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Rubberduck.Refactorings.Common
+{
+    public static class IModuleRewriterExtensions
```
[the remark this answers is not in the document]
BZngr: Since RD 'owns' `ModuleRewriter`, adding the methods to `IModuleRewriter` is an available choice and certainly avoids the drawbacks you mention.  Even though I used an extension class, I'm not advocating that extension methods are the _best_ choice here.  For a PR submission, I am very reluctant to modify such a ubiquitous interface as `IModuleRewriter` without specific direction...So, extension methods provide a less impactful way to initially add/propose new behavior.  I'm good with making the extension methods part of the `IModuleRewriter` if that is preferred.

Rubberduck.Refactorings/Common/IModuleRewriterExtensions.cs
MDoerner: That resolves the issue. You could alternatively use `ToDictionary` on `fieldsByListContext`. There should be an extension method turning it into a `Dictionary<TKey, List<TValue>>`.

BZngr:
Code is updated per the comments.  Retained the `IModuleRewriterExtensions` class rather than modify `IModuleRewriter`.

PR #5578, opened 2020-09-21
EncapsulateFieldRefactoring - Split Refactoring

BZngr:
This PR refactors/rewrites the `EncapsulateFieldRefactoring` (EF) to conform to the Split Refactoring design (#5387).  

The `EncapsulateFieldRefactoringAction` is now a facade object to coordinate two underlying refactoring actions: the `EncapsulateFieldUseBackingFieldRefactoringAction` and the `EncapsulateFieldUseBackingUDTMemberRefactoringAction`.

The PR is focused primarily on reducing the technical debt of being the only 'non-split' refactoring.  Consequently, it does not add any new features or close any issues.  The intent is that the EF will operate identically before and after it is merged.  Even so, as part of the migration to the split refactoring design, the PR includes the following internal improvements:

- Incorporates use of CW/DI and factories for creating objects.  
- Incorporates use of `IRefactoringPreviewProvider` to generate previews 
- Retains the existing battery of EF tests.  (Some re-organization but otherwise left essentially unchanged).    
- Retains the properties and method signatures used by the `EncapsulateFieldView` and `EncapsulateFieldViewModel` to isolate the UI from all changes.   Not an improvement per se, but there is/was some work going on with the UI, and I wanted the PR to avoid impacting that work.

Note: A parallel/personal goal of this refactoring was to evaluate and make every class associated with the refactoring more focused and, as a result, smaller.  Consequently, I had some hope for a net decrease in lines of code (LOC) but it turned out that there was a net increase instead.  Some quick analysis revealed that the addition of tests,  factories, and preview providers account for nearly the entire net LOC increase - with the added test content comprising most of the total.


Rubberduck.Refactorings/EncapsulateField/FieldCandidates/ArrayCandidate.cs
```diff
@@ -1,64 +1,15 @@
...
-    {
-        private string _subscripts;
-        public ArrayCandidate(Declaration declaration, IValidateVBAIdentifiers validator)
-            :base(declaration, validator)
+        public ArrayCandidate(Declaration declaration)
+            :base(declaration)
         {
-            ImplementLet = false;
-            ImplementSet = false;
             PropertyAsTypeName = Tokens.Variant;
```
bclothier, line 11: Arrays can be typed to something besides `Variant`? Why are we setting it `Variant` here and not looking at declaration's type?

Rubberduck.Refactorings/EncapsulateField/FieldCandidates/EncapsulateFieldCandidate.cs
```diff
@@ -21,217 +18,122 @@ public interface IEncapsulateFieldCandidate : IEncapsulateFieldRefactoringElemen
...
-            else if (Declaration.IsObject)
-            {
-                ImplementLet = false;
-                ImplementSet = true;
-            }
-        }
+            //5.3.1 The declared type of a function declaration may not be a private enum.
+            PropertyAsTypeName = declaration.IsEnumField() && declaration.AsTypeDeclaration.HasPrivateAccessibility()
+                ? Tokens.Long
+                : declaration.AsTypeName;
```
bclothier, line 53: Will the preview issue a warning in this case? I'm inclined to think that it's usually a bug or at least unintentional to expose a member that corresponds to a private enum.

Rubberduck.Refactorings/ReplacePrivateUDTMemberReferences/UserDefinedTypeInstance.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var prCtxt = accessExprContext as ParserRuleContext;
+                if (prCtxt == goalContext)
+                {
+                    return true;
+                }
+                accessExprContext = accessExprContext.GetAncestor<T>();
+            }
+
+            Debug.Assert(guard < 100);
```
bclothier, line 64: I think 100 should be made into a named constant.... `MaxAncestryDescend` or something.

RubberduckTests/Refactoring/EncapsulateField/EncapsulateFieldTestComponentResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+using Rubberduck.Refactorings.ReplacePrivateUDTMemberReferences;
+using Rubberduck.Refactorings.ReplaceDeclarationIdentifier;
+using Rubberduck.Refactorings.EncapsulateFieldUseBackingUDTMember;
+using Rubberduck.Refactorings.EncapsulateFieldUseBackingField;
+using Rubberduck.Refactorings.EncapsulateFieldInsertNewCode;
+using System;
+
+namespace RubberduckTests.Refactoring.EncapsulateField
+{
+    public class EncapsulateFieldTestComponentResolver
```
bclothier, line 16: Is this to work with the lack of CW? Consider using a stripped down CW for the tests? That would give a better fidelity compare the actual code and the unit tests.

bclothier (commented):
Apologies for not having commented on this stellar PR sooner. This looks very great and I look forward to merging it. My comments are mostly nitpicks. Thank you!

Rubberduck.Refactorings/Common/CodeBuilder.cs
```diff
@@ -34,56 +34,72 @@ public interface ICodeBuilder
...
         /// <param name="parameterIdentifier">Defaults to '<paramref name="propertyIdentifier"/>Value' unless otherwise specified</param>
         bool TryBuildPropertyGetCodeBlock(Declaration prototype,
-                                            string propertyIdentifier,
-                                            out string codeBlock,
-                                            string accessibility = null,
-                                            string content = null);
+            string propertyIdentifier,
+            out string codeBlock,
+            string accessibility = null,
+            string content = null);
```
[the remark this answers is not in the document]
MDoerner: We actually already have an `Accesdibility` enum in the parsing project.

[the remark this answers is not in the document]
BZngr: Yes.  Should be using the `Accessibility` enum (default is `Public` to match current behavior).  Changing the parameter type will cause this PR to drag the `AddInterfaceImplementationsRefactoringAction` into this PR, but I think it should be done.

Rubberduck.Refactorings/EncapsulateField/FieldCandidates/ArrayCandidate.cs
BZngr: AIUI, we will lose Office97 back compatibility :smirk:.  Yes, it would be better to return the array's declared type.  This falls a bit outside the goal of the PR, but looks fairly straightforward to change.

Rubberduck.Refactorings/EncapsulateField/FieldCandidates/EncapsulateFieldCandidate.cs
BZngr: In the current version, if the user selected a field for encapsulation (`Private` or `Public`), then the refactoring 'assumed' that the intent was to make them all publicly accessible.  A regrettable design choice in retrospect.  Exposing a `Private` field as a `Public Property` would be better implemented as an option on a per field basis.  #5421, which identifies this issue, was created based on your feedback (in chat IIRC) some time ago.  Since the focus this PR is on technical debt and changing this likely requires a change in the UI, I'd like to defer resolution of this comment to issue #5421.

RubberduckTests/Refactoring/EncapsulateField/EncapsulateFieldTestComponentResolver.cs
BZngr: Yes...this class/approach helped me to manage and evaluate my ctor injected dependencies, understand a bit more about what CW provides, and made object instantiations for testing quite a bit easier.  That said, I never considered that I could/should use a stripped down version of CW for the tests.  I'll check into it - and it will be, no doubt, a learning experience.  TBH, CW is still quite mysterious to me.

bclothier: There is [an example](https://github.com/rubberduck-vba/Rubberduck/tree/next/RubberduckTests/Refactoring/MockIoC), inspired by [this article](https://blog.ploeh.dk/2013/03/11/auto-mocking-container/). HTH!

Rubberduck.Refactorings/EncapsulateField/FieldCandidates/ArrayCandidate.cs
bclothier: Hmm, can’t tell if we still can support O97 (we require .NET 4.6.1, and that doesn’t work on older OSes, think O97 still can be installed on W10, but unsure). Assuming yes, I am fine with leaving it as is with a comment for a future PR to conditionally branch based on Office version.

Rubberduck.Refactorings/EncapsulateField/EncapsulateFieldInsertNewCode/EncapsulateFieldCodeBuilder.cs
```diff
@@ -0,0 +1,106 @@
...
+        string BuildFieldDeclaration(Declaration target, string identifier);
+    }
+
+    /// <summary>
+    /// EncapsulateFieldCodeBuilder wraps an ICodeBuilder instance to extend it for the 
+    /// specific needs of an EncapsulateField refactoring action.
+    /// </summary>
+    public class EncapsulateFieldCodeBuilder : IEncapsulateFieldCodeBuilder
+    {
+        private const string FourSpaces = "    ";
```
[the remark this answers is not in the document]
BZngr: Agreed...IMO indenter settings are the right way to go _especially_ where this (or any refactoring) generates brand new code block(s).  When inserting into existing content - the choice is not quite so clear (at least to me)...`EncapsulateFieldRefactoring` needs to support both use cases.  

So, as I've explored how to address this comment, I've come to the decision-point whether to ctor inject an `IIndenter` interface into the `CodeBuilder` (making an Indenter instance _always_ available), or use method injection to allow callers to dictate the 'indention style'.  

I'm leaning toward method injection since it relieves the `CodeBuilder` of the responsibility to determine _when_ to use the `Indenter` (it simply by checking if an `IIndenter` interface has been provided). e.g., use case: using `ICodeBuilder` to generate a line of code to insert into an existing block - where the caller might prefer to apply an indention style that matches what is currently used within the block.  

The downside of method injection is that it adds, yet another, optional parameter to a handful of the `ICodeBuilder` methods - which already have several.  (e.g., 
```
public bool TryBuildPropertyLetCodeBlock(Declaration prototype, 
               string propertyIdentifier, 
               out string codeBlock,
               Accessibility accessibility = Accessibility.Public,
               string content = null,
               string parameterIdentifier = null,
               IIndenter indenter = null) 
```
Thoughts/Reactions/Opinions?...Time to inject a `struct`/DTO?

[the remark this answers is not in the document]
bclothier: I'm inclining toward KISS / YAGNI - while it might be cool for the callers to change indentation style, I feel this encourage inconsistent style enforcement. I mean, they should be setting the indenter's behavior before they refactor at all, and if the code they're working doesn't conform already, it should conform (and obey 😉 ) to the current indenter's setting. So I'd always provide to the code builder and tell users to go to the settings if they don't like it.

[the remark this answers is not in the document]
BZngr: Glad I asked...Admittedly, though I was leaning towards the method injection approach, I had my doubts because it seemed intuitively 'correct' that a `CodeBuilder` object _should_ always have access to the current `Indenter` settings.  The approach you are advocating implies that when editing an existing block (e.g., inserting new members into a `UserDefinedType Declaration`, a brand new declaration block will be generated, indented (with both new and existing UDT members), and replace the existing `UserDefinedType Declaration` in its entirety...This is different than the current implementation, but sounds good to me.

Rubberduck.Refactorings/Common/CodeBuilder.cs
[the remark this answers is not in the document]
BZngr: parameter updated in the latest sync

Rubberduck.Refactorings/EncapsulateField/EncapsulateFieldInsertNewCode/EncapsulateFieldCodeBuilder.cs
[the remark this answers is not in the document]
BZngr: integrated use of `IIndenter` in the latest sync

Rubberduck.Refactorings/EncapsulateField/EncapsulateFieldInsertNewCode/EncapsulateFieldCodeBuilder.cs
```diff
@@ -0,0 +1,106 @@
...
+                    $"{FourSpaces}{Tokens.Set} {propertyAttributes.PropertyName} = {propertyAttributes.BackingField}",
+                    Tokens.Else,
+                    $"{FourSpaces}{propertyAttributes.PropertyName} = {propertyAttributes.BackingField}",
+                    $"{Tokens.End} {Tokens.If}",
+                    Environment.NewLine);
+            }
+
+            if (!_codeBuilder.TryBuildPropertyGetCodeBlock(propertyAttributes.Declaration, propertyAttributes.PropertyName, out var propertyGet, content: $"{FourSpaces}{getterContent}"))
+            {
+                throw new ArgumentException();
```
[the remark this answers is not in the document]
BZngr: addressed in the latest sync

Rubberduck.Refactorings/EncapsulateField/FieldCandidates/ArrayCandidate.cs
BZngr: Comment added in latest sync

BZngr:
Latest update to this PR (Oct 7) addresses most comments (primarily related to `CodeBuilder` and `Indenter` use).  Will be working on the CW items next along with a couple other comments above where I have yet to indicate that they have been addressed.

Rubberduck.Refactorings/EncapsulateField/EncapsulateFieldInsertNewCode/NewContentAggregator.cs
```diff
@@ -0,0 +1,141 @@
...
+        int NewLineLimit { set; get; }
+    }
+
+    /// <summary>
+    /// NewContentAggregator provides a repository for caching generated code blocks
+    /// and retrieving them as an aggregated single block of code organized by <c>NewContentType</c>.
+    /// </summary>
+    public class NewContentAggregator : INewContentAggregator
+    {
+        private readonly static string _doubleSpace = $"{Environment.NewLine}{Environment.NewLine}";
```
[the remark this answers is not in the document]
BZngr: Not sure I understand the motivation behind making this a public static string.  Since `$"{Environment.NewLine}{Environment.NewLine}"` is such a small and inherently clear expression as-is, there seems to be little gain in terms of 're-use' to define this publicly somewhere else...Can you say a bit more about what you were considering?  

Perhaps introducing a `string` extension function like `<string>.AppendNewLines(int count)` would be an option.  For the use case above, `_doubleSpace = string.Empty.AppendNewLines(2);`  in the constructor works...Or, get rid of `_doubleSpace` altogether and use the extension method inline.  `PrependNewLines(int)` would then be convenient to have as well.

Also, since `Environment.NewLine` is a static property/function - I do not see a way to define it as a resource expression directly.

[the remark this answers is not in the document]
bclothier: The reason I suggested that way was because I have a feeling I've seen that usage somewhere and instead of having multiple definition of "add 2 blank lines", define it once for common use.

I didn't mean as a resource. You would add to [`ApplicationConstants`](https://github.com/rubberduck-vba/Rubberduck/blob/next/Rubberduck.Resources/ApplicationConstants.cs), perhaps, or as a new class.

RubberduckTests/Refactoring/EncapsulateField/EncapsulateFieldTestComponentResolver.cs
BZngr: Created a `WindsorContainer` that managed dependencies for the tests.  

I can see (now in retrospect) that the real value of a container to support testing is highest if it exists right from the start.  Since design/development introduces a lot of churn to various constructor signatures...the test code would have simply continued to compile without modification after changes to the refactoring's constructors (which was frequent when refactoring to use DI).  I think a container to support testing would be overkill in most cases - but I think it is well worth having for this refactoring.

BZngr:
All comments have been addressed.  And, short of inspiring new comments, they are (at least functionally) resolved.  Rather than add short responses to each comment above, I think/hope the the last six commit comments are sufficient to explain what was done.

PR #5590, opened 2020-10-03
Add additional checks to ensure that selection is valid

KonH:
Related to https://github.com/rubberduck-vba/Rubberduck/issues/5553


Vogel612 (approved):
While this is only a symptomatic fix, it's probably the best we can hope for right now. Since the UI is difficult if not impossible to unit-test I'd be happy to merge this :)

retailcoder (approved):
Well done! This ensures "no selection" doesn't leave stale content in the UI, as making the whole block conditional on `cs` being non-null would. Very happy to merge this very welcome contribution - by all means keep 'em coming! 😀👍

PR #5595, opened 2020-10-13
Fix german typo

Tanarri:
spelling mistake


MDoerner:
This fix is already part of the open PR #5585. So, I close this one.

PR #5606, opened 2020-10-25
Enhance ExtractInterfaceRefactoring with implementation options

### LINKED ISSUE #5542: ExtractInterface Refactoring can provide implementation options
**Justification**
`ExtractInterfaceRefactoring` can offer to provide an implementation of the extracted interface to avoid requiring the user to manipulate the generated content.  

**Description**
From [The Builder Walkthrough Post](https://rubberduckvba.wordpress.com/2020/05/29/builder-walkthrough):
>The scary part is that it feels as though if Extract Interface accounted for the presence of a Private Type in a similar way Encapsulate Field does, then even the TODO placeholder bits could be fully automated. Might be something to explore there…

Lets fully automate replacement of the "placeholder bits".  This can be done based on the content of the default interface members that have been used to defined the new interface.

It seems that there are four user options:
1. **Forward object member calls to the interface members**. _Use Case:_ Does not break external references to current member(s) yet complies with _DRY_ coding practices.
2. **Forward interface member calls to the Object members**. _Use Case:_ User is partitioning a 'wide' default interface into more focused interface definitions.  Does not break references to current member(s) yet complies with _DRY_ coding practices.
3. **Copy the object member implementations to the interface members** (some renaming required) and delete the object members.  _Use Case:_ User is creating a new class and only wants to expose state and capabilities using an interface.  (This is basically the _Builder Walkthrough_ scenario).  This option only makes sense if the default interface members have yet to be externally referenced.  If references exist, see options 1 & 2.
4. **Do not implement the interface** and insert the 'TODO placeholder bits' (the current implementation of `ExtractInterfaceRefactoring`)

My current thinking is that Option 1 would be the default option.

**Additional context**
Using a portion of the `User` class API of the _Builder Walkthrough_ post as an example:
```
Option Explicit
Public Id As String
Public UserName As String
Public FirstName As String
```
After `EncapsulateFieldRefactoring` becomes:

```
Option Explicit

Private Type TUser
    Id As String
    UserName As String
    FirstName As String
End Type

Private this As TUser

Public Property Get Id() As String
    Id = this.Id
End Property
Public Property Let Id(ByVal value As String)
    this.Id = value
End Property
Public Property Get UserName() As String
    UserName = this.UserName
End Property
Public Property Let UserName(ByVal value As String)
    this.UserName = value
End Property
Public Property Get FirstName() As String
    FirstName = this.FirstName
End Property
Public Property Let FirstName(ByVal value As String)
    this.FirstName = value
End Property
```
Using `ExtractInterfaceRefactoring` with:
1.  User selects **Forward object member calls to the interface members**
2. Selecting just the _getters_ (as was done in the post) 

the delivered content becomes...free of any user edits:
```
Option Explicit

Implements IUser

Private Type TUser
    Id As String
    UserName As String
    FirstName As String
End Type

Private this As TUser

Public Property Get Id() As String
    Id = IUser_Id
End Property
Public Property Let Id(ByVal value As String)
    this.Id = value
End Property
Public Property Get UserName() As String
    UserName =  IUser_UserName
End Property
Public Property Let UserName(ByVal value As String)
    this.UserName = value
End Property
Public Property Get FirstName() As String
    FirstName = IUser_FirstName
End Property
Public Property Let FirstName(ByVal value As String)
    this.FirstName = value
End Property

Private Property Get IUser_Id() As String
    IUser_Id = this.Id
End Property
Private Property Get IUser_UserName() As String
    IUser_UserName= this.UserName
End Property
Private Property Get IUser_FirstName() As String
    IUser_FirstName= this.FirstName
End Property
```

BZngr:
Closes #5542

The implementation options are:
1. `ForwardObjectMembersToInterface`
2. `ForwardInterfaceToObjectMembers`
3. `NoInterfaceImplementation` (the current default behavior) 
4. `ReplaceObjectMembersWithInterface`

The default option presented to the user is `ForwardInterfaceToObjectMembers`.  

The implementation of the PR matches the behavior as described in #5542 with the exception of `ReplaceObjectMembersWithInterface`.  In the case where a member has external references, the refactoring retains the referenced member(s) forwarding the calls to the associated interface member (rather than refusing the request (per #5542).  This avoids creating uncompilable code and moves the user _further_ towards his/her original intent to completely eliminating the member from the class.

In addition, the following improvements are also included:
1. Adds a ConflictFinder to prevent the user from generating an interface class with a conflicting identifier.  
2. Prevents the selection of Members with underscores in their name to be part of the interface. 
3. Uses the `Indenter` settings via the `CodeBuilder`


retailcoder (approved):
LGTM

PR #5623, opened 2020-11-03
Move ToDoExplorer resources

### LINKED ISSUE #4541: .resx needs a clean-up

IvenBach:
Close #4541.


BZngr (approved):
LGTM

PR #5643, opened 2020-11-30
Include enums in ImplicitPublicMemberInspection

### LINKED ISSUE #5642: Flag implicitly public enumerations as such
**What**
Enumeration declarations aren't included as part of checking for implicitly public members.

**Why**
Enumerations should be included to inform the user of their existence.

**Example**

```vb
Enum Foo
    Member
End Enum
```

IvenBach:
Close #5642.

### DIFF
--- a/Rubberduck.CodeAnalysis/Inspections/Concrete/ImplicitPublicMemberInspection.cs
+++ b/Rubberduck.CodeAnalysis/Inspections/Concrete/ImplicitPublicMemberInspection.cs
@@ -42,7 +42,8 @@ namespace Rubberduck.CodeAnalysis.Inspections.Concrete
             DeclarationType.Procedure,
             DeclarationType.PropertyGet,
             DeclarationType.PropertyLet,
-            DeclarationType.PropertySet
+            DeclarationType.PropertySet,
+            DeclarationType.Enumeration
         };
 
         protected override bool IsResultDeclaration(Declaration declaration, DeclarationFinder finder)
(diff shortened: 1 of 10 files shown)


Rubberduck.CodeAnalysis/Inspections/Concrete/ImplicitPublicMemberInspection.cs
```diff
@@ -42,7 +42,8 @@ public ImplicitPublicMemberInspection(IDeclarationFinderProvider declarationFind
             DeclarationType.Procedure,
             DeclarationType.PropertyGet,
             DeclarationType.PropertyLet,
-            DeclarationType.PropertySet
+            DeclarationType.PropertySet,
+            DeclarationType.Enumeration
```
BZngr, line 46: Seems that we do not check for implicit `Public Fields`, `Constants`, or `UserDefinedTypes` either.  Should they be included in this inspection/PR?

IvenBach: Brief testing leads me to believe the omission is intentional for fields and constants. Within a standard module `DeclaredIn` everything was declared and subsequently called from `AccessedBy` When trying to access them implicitly:
```vb
'DeclaredIn module
Const FooConstant As String = "FooConstantValue" 'Method or data member not found without `Public` modifier

Dim FooField As String 'Method or data member not found with `Dim`.

Type FooType
    Name As String
    Age As Long
End Type
```

```vb
'AccessedBy module
Public Sub DoBar()
    Debug.Print DeclaredIn.FooConstant 'Requires `Public`
    Debug.Print DeclaredIn.FooField 'Requires `Public`
    
    Dim userType As DeclaredIn.FooType 'Implicitly access just fine.
    userType.Name = "Morty McMorterson"
    userType.Age = 1
End Sub
```

Based on the above I assume `UserDefinedType` could be included.

retailcoder (approved):
Well done, LGTM!

PR #5647, opened 2020-12-02
Fix "Variable is implicitly Variant" quickfix results for array declarations

### LINKED ISSUE #5646: Fix for "Variable is implicitly Variant" breaks array declaration

Version 2.5.1.5676
Host Version: 16.0.12527.21330
Host Executable: EXCEL.EXE

**Description**
fix for inspection breaks code

**To Reproduce**
`Private myARRAY(1 To 3)` 

will be changed to

`Private myARRAY As Variant(1 To 3)`

**Expected behavior**
`Private myARRAY(1 To 3) As Variant`

**Screenshots**
.
**Logfile**
.

**Additional context**
.

BZngr:
Closes #5646 

PR modifies the `DeclareAsExplicitVariant` to leverage the `ImplicitTypeToExplicitRefactoringAction` and removes the  `DeclareAsExplicitVariant` tests as they are now redundant to the tests for the refactoring action.


Vogel612 (approved):

MDoerner (commented):
Nice! I really like that the quickfix now uses a refactoring action.
My only real concern with the PR is that the old tests for the quickfix have been removed and instead of being extended. We also have tests for the other refactoring-based quickfixes.

BZngr:
Thanks for questioning the use of `dynamic`.  Even though it _worked_ as part of the origin quickfix,  your comment made me consider/realize something.  Now that the work is done in a refactoring action, the model parameter is not  guaranteed to be coming from (and vetted by) the inspection/quickfix processing.  So, removed the use of `dynamic` and added a guard clause to enforce explicit input expectations. 

I've also added unit tests for the two quickfixes.  They basically test that the quickfix can successfully invoke the refactoring action.  All the other various scenarios from the original quickfix were moved to the refactoring action unit tests.

PR #5675, opened 2021-02-21
Potentially fix the AV when running Rubberduck with Access

bclothier:
There is an access violation when one tries to parse too soon or at wrong time which then leaks the pointer and causes Access to becomes unstable and thus can crash when exiting normally. This helps fixes and also 2 minor changes for developer's convenience.


Vogel612:
love the overview of compiler constants we have in the solution

PR #5714, opened 2021-04-02
UnitTests: Fix AddTestMethod commands to generate compilable code

### LINKED ISSUE #4169: New Unit Test Added inside existing Unit Test
Just to be clear, this is what happened when I attempted to add a 2nd Test Method using the Test Explorer UI. 

I can confirm that happens when the cursor is located:
 - Above
 - Inside, or
 - Below 

the original Unit Test.

Word 2010, Rubberduck v2.2.0.3468-pre

```
'@TestMethod
Public Sub ValidateIndexOneIsLevel1() ' Expected: IsValid = True
    On Error GoTo TestFail
    
    'Arrange:
    
    'Act:

    'Assert:
    Assert.Inconclusive

TestExit:
    Exit Sub
TestFail:
    Assert.Fail "Test raised an error: #" & Err.Number & " - " & Err.description
'@TestMethod
    On Error GoTo TestFail
    
    'Arrange:

    'Act:

    'Assert:
    Assert.Inconclusive

TestExit:
    Exit Sub
TestFail:
    Assert.Fail "Test raised an error: #" & Err.Number & " - " & Err.description
End Sub

End Sub
```

BZngr:
Closes #4169

This PR addresses the uncompilable code generated when adding a UnitTest TestMethod.

The fundamental issue involved was an incorrect insertion point for the generated code.  

1. Noticed that the commands did not use the `RewritingManager`.  So, they now do.
2. The two 'Add Test' command classes (`AddTestMethodCommand` and `AddTestMethodExpectedErrorCommand`) had identical code except for a single function call.  So, the PR introduces a base class to make these commands more DRY and consistent.
3. There were a number of `'TODO` comments in the tests regarding temporal coupling.  These were addressed as well.

This is my first PR within the UnitTest area.   I'm hoping, no basic design/architecture 'rules' have not been violated by my changes.  The commands did not use the `RewritingManager` so is it possible that this area may be a little behind in other areas?  I started wondering if calling upon a `RefactoringAction` would now be a preferred approach for generating/adding unit test code and modules.

### DIFF
--- a/Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestMethodBase.cs
+++ b/Rubberduck.Core/UI/UnitTesting/ComCommands/AddTestMethodBase.cs
@@ -1,0 +1,126 @@
+﻿using System;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Rubberduck.Parsing.Annotations.Concrete;
+using Rubberduck.Parsing.Rewriter;
+using Rubberduck.Parsing.Symbols;
+using Rubberduck.Parsing.VBA;
+using Rubberduck.UI.Command.ComCommands;
+using Rubberduck.UnitTesting;
+using Rubberduck.UnitTesting.CodeGeneration;
+using Rubberduck.VBEditor.Events;
+using Rubberduck.VBEditor.Extensions;
+using Rubberduck.VBEditor.SafeComWrappers.Abstract;
+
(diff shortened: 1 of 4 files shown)


retailcoder (approved):
Good stuff, LGTM. Indeed, unit testing has been needing some love, this is great!

PR #5754, opened 2021-04-30
Italian translation added

PhilCattivocaratere:
Italian translation added but not the installer user interface yet.


PhilCattivocaratere (commented):
It loaded what I wrote. I already viewed them

bclothier:
Awesome thank you!

Just in case you're wondering about the installer interface, you need to create a `.iss` file. Make sure to use UTF-8 BOM (the BOM part is important).

You can find it [here](https://github.com/rubberduck-vba/Rubberduck/tree/next/Rubberduck.Deployment/InnoSetup/Includes). Easiest to copy from one of the `***.CustomMessages.iss` and then translate into Italian. 

Thank you for the contribution!

retailcoder (approved):
Thank you so much for this contribution! I see the new `Language_` key was added, perfect!

PR #5861, opened 2021-09-30
Fix return value discarded inspection

### LINKED ISSUE #5853: Spurious "Return value of '...' gets discarded" inspection result

    Version 2.5.2.5906
    Host Version: 16.0.13801.20864
    Host Executable: EXCEL.EXE

**Description**
Rubberduck code inspection yields `Return value of '...' gets discarded` for a function that does not discard the return value.

**To Reproduce**
Run inspection on this code:
```VB
Private Function Credentials(ByVal Username As String, ByVal Password As String) As Dictionary
    Set Credentials = New Dictionary
    Credentials.Add "username", Username
    Credentials.Add "password", Password
End Function
```

**Expected behavior**
Spurious inspection result should not be shown.

MDoerner:
Closes #5853

Previously, we did not always determine correctly whether there is a member access on the return value of a function. The problem  was that the reference in a member access expression goes to the unrestricted identifier while it goes to the whole simple name expression in case of a simple name expression. This makes it necessary to deal with them in different ways; the first member access parent of a reference to a member access is the member access itself while it is a further member access for all other kinds of references.

This affected both kinds of function return value discarded inspections, since they use the same logic to determine whether a reference is used.

### DIFF
--- a/Rubberduck.Parsing/ParserRuleContextExtensions.cs
+++ b/Rubberduck.Parsing/ParserRuleContextExtensions.cs
@@ -30,10 +30,13 @@ namespace Rubberduck.Parsing
                                  context.Stop.EndColumn() + 1);
         }
 
-        public static bool Contains(this ParserRuleContext context, ParserRuleContext otherContext)
+        /// <summary>
+        ///  Returns whether the other context from the same parse tree is wholly contained in the current context
+        /// </summary>
+        public static bool Contains(this ParserRuleContext context, ParserRuleContext otherContextInSameParseTree)
         {
-            return context.start.TokenIndex <= otherContext.start.TokenIndex
-                   && context.stop.TokenIndex >= otherContext.stop.TokenIndex;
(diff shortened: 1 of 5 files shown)


Rubberduck.Parsing/ParserRuleContextExtensions.cs
```diff
@@ -30,6 +30,12 @@ public static Selection GetSelection(this ParserRuleContext context)
                                  context.Stop.EndColumn() + 1);
         }
 
+        public static bool Contains(this ParserRuleContext context, ParserRuleContext otherContext)
```
Vogel612, line 33: I would like to have some short doccomment on this to just allow IDE tooling to make clear that this is not an `Intersects` in usages, but that's honestly just a nitpick :)

Vogel612 (approved):

Rubberduck.Parsing/ParserRuleContextExtensions.cs
MDoerner: Sure, I will add that tomorrow.

Issue #5844, opened 2021-07-30
Add new refactoring "Inject Definition Module Prefix"...


jim-oflaherty-jr-qalocate-com:
**Justification**
As a codebase grows, the probability of a naming conflict arising among public subs and functions as external modules and/or new modules are added increases quickly. There is a dynamic tension between using simple straightforward names for subs and functions within a module and those names having already been chosen in another code-base's modules which now must be merged and integrated. Manually adding the "Module Prefix" with a dot to clarify the definition context becomes mandatory in resolving the naming conflicts in at least one of the two sides of the conflict.

**Description**
When activating the refactoring, the context of the sub or function is used to determine its definition module's name (including the one within which the refactoring is occurring). And then the module's name is inserted prior to the sub or function name followed by a dot. This disambiguates the sub or function ownership.

For example...
```vba
'Code inside module "mFoo"
Dim headerNames() As String
Dim headerNamesSize As Long

headerNamesSize = size(headerNames) 'Original unrefactored code - "size" is a function in module "mArrays"
headerNamesSize = mArrays.size(headerNames) 'Following the first refactoring pass, the "size" function has the module's name, "mArrays" within which it is defined, prepended
headerNamesSize = Geocoder.mArrays.size(headerNames) 'Following the second refactoring pass, the "size" function has the project's name, "Geocoder" within which the "mArrays" module is defined, prepended
```

**Additional context**
Assuming the context of a successfully compiled state, the existing parsed state already holds EXACTLY these relationships. In fact, that is precisely the information being displayed to the right of the "Ready" button.

I don't have enough experience refactoring Excel yet to know if also providing the project name prefix with a dot in front of the module would be as useful. If so, then perhaps this refactoring could prepend a module name to a sub or function. And then if there is already a module name, prepend the project name to the already existing module name.

I also sense there is a possibility of adding some sort of inspection which identifies calls to simpler named subs and functions be prepended with its defining module to proactively avoid the naming conflicts and refactoring efforts later.

MDoerner:
I like this suggestion. I would simply call it `Qualify`. It would also apply to properties and fields on modules. We could also qualify with the project, if the reference points to a referenced library. 

This should not be hard to implement since we store the necessary information on the references anyway. The biggest problem is probably figuring out in an efficient way when the refactoring should be available. 

[...]

PR #5864, opened 2021-10-02
Modify RenameRefactoring to use ReplaceReferencesRefactoringAction

BZngr:
This PR accomplishes a portion of the opportunities identified by #5844.  The changes modify `RenameRefactoring's` default behavior to qualify all external references of a renamed declaration. 

 If this default is acceptable, then it can lead to less restrictive rename conflict detection logic since any rename collisions in external modules would be eliminated.

### DIFF
--- a/Rubberduck.Refactorings/Rename/RenameCodeDefinedIdentifierRefactoringAction.cs
+++ b/Rubberduck.Refactorings/Rename/RenameCodeDefinedIdentifierRefactoringAction.cs
@@ -181,9 +181,9 @@ namespace Rubberduck.Refactorings.Rename
                 ModuleQualifyExternalReferences = true
             };
 
-            foreach (var rf in target.References)
+            foreach (var reference in target.References)
             {
-                replaceReferencesModel.AssignReferenceReplacementExpression(rf, newName);
+                replaceReferencesModel.AssignReferenceReplacementExpression(reference, newName);
             }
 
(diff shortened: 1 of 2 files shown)


Rubberduck.Refactorings/Rename/RenameCodeDefinedIdentifierRefactoringAction.cs
```diff
@@ -172,21 +176,17 @@ private void RenameStandardElements(Declaration target, string newName, IRewrite
...
-                    && !reference.IsArrayAccess
-                    && !reference.IsDefaultMemberAccess)
-                .GroupBy(r => r.QualifiedModuleName);
+            var replaceReferencesModel = new ReplaceReferencesModel()
+            {
+                ModuleQualifyExternalReferences = true
+            };
 
-            foreach (var grouping in modules)
+            foreach (var rf in target.References)
```
MDoerner, line 184: The general rule in the code base is not to shorten identifier names. So, this should probably be `reference` instead of `rf`.

MDoerner (approved):
I have to say that I like the change.

I just have one nit-pick regarding the variable naming.

Still, I would like to hear from others, especially those experienced with multiple hosts, what they think about module qualifying on rename.

Vogel612 (approved):

Vogel612:
In accordance with the steps outlined in #5692 I'm merging this so it can be synced into that PR.

PR #5891, opened 2021-10-22
Appveyor.yml shenanigans

Vogel612:
This hopefully restores coverage report uploads to our external coverage provider thingy.


Vogel612:
Forcing a merge of this PR, even without the signature verification for the node-based uploader, mostly because otherwise the CI will break when the chocolatey package is phased out.

PR #5899, opened 2021-11-16
Non-disposal decorators for more COM wrapper safety

MDoerner:
This PR adds decorators for several com wrapper types to suppress disposal attempts on them.

This is relevant in two places. First, the wrappers handed out by the `ProjectsRepository` are references to cashed wrappers; only the repository itself should be allowed to dispose the instances when it clears its cache. 
Second, disposal of the com wrappers bound in CW is non-deterministic. In order to avoid running into concurrency issues on shutdown, CW should not be allowed to dispose the wrappers. This will than happen deterministically when disposing the `ComSafe`.


Vogel612 (approved):
LGTM, did not review the changes to the tests, though. I also can't immediately see the reason why some of the wrapped members are only exposed as readonly, while some others are not. I presume that follows the wrapped interface logic though (since the interface is inherited as well).

retailcoder (approved):

PR #5951, opened 2022-02-23
Use env to generate release build URL

Vogel612:


retailcoder (approved):
Ha, I thought it was the GH org + repo, but indeed it's the AV account.. Thanks!

PR #5980, opened 2022-05-08
Update Italian translation.

PhilCattivocaratere:


Vogel612 (approved):
I can't tell whether these translations are any good, but other than that this looks great :)

Issue #5973, opened 2022-04-23
Code Inspection for Property/Function/Procedure not used  cannot be disabled


FullValueRider:
**Rubberduck version information**
Version 2.5.2.6062
OS: Microsoft Windows NT 10.0.22598.0, x64
Host Product: Microsoft Office x64
Host Version: 16.0.15028.20204
Host Executable: EXCEL.EXE

**Description**
I want to disable the Code Quality inspections for Property/Procedure/Function is not used.  However I can't do this as these items do not appear in the Rubberduck settings list for Code Quality Inspections

**Expected behavior**
If a rule appears in the Code inspections it should be possible to disable it

BZngr:
The Code Quality inspection for 'Property/Procedure/Function is not used' is identified as "_Procedure is not referred to_" in the Code Inspection/Code Quality Settings list.  

Alternatively, if you select a flagged result for this (or any) inspection result in the Code Inspections window, a '_disable this inspection_' link is provided in the lower pane  which will also change the setting for the inspection to "Do Not Show".

FullValueRider:
Thanks for the clarification.  In which case I'd argue that the text that appears in the code inspection pane be updated to reflect that  of the code inspection text in the code inspection settings pane or vice vera.  I'd prefer  'is not used' to 'is not referred to'

PR #5986, opened 2022-06-14
Improve resource consistency for `XXXNotUsedInspections'

BZngr:
Issue #5973 surfaced an inconsistency between the content/phrasing of the `Code Inspections` result description and the `ProcedureNotUsedInspection's` `InspectionName` resource used by the `Settings` editor.

Inconsistencies were found for all of the current 'XXXNotUsedInspection' types:

- `ConstantNotUsedInspection`
- `LineLabelNotUsedInspection`
- `ParameterNotUsedInspection`
- `ProcedureNotUsedInspection`
- `VariableNotUsedInspection`

This PR proposes to:

1. Modify the `InspectionName` resource to a phrase of the form "XXX is not used" in preference to "XXX is not referred to".
2. Delete translated "XXXNotUsedInspection" resources for both `InspectionNames` and `InspectionResults`. Re-visiting the translations looks warranted. For example, the German translations contained three forms of the "XXX not used/referred" message across the 5 inspections. Inconsistencies appear to exist in the other translations as well.  Updating the resource translations for these inspections at the same time seems like the most reliable route to getting everything self-consistent.

### DIFF
--- a/Rubberduck.Resources/Inspections/InspectionNames.Designer.cs
+++ b/Rubberduck.Resources/Inspections/InspectionNames.Designer.cs
@@ -808,7 +808,7 @@ namespace Rubberduck.Resources.Inspections {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Parameter is not referred to..
+        ///   Looks up a localized string similar to Parameter is not used..
         /// </summary>
         public static string ParameterNotUsedInspection {
             get {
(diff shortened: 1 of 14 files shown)


Vogel612 (approved):
I'm not super happy about outright deleting the old resources, but I don't have a better alternative, soo ... LGTM

BZngr:
@Vogel612  First - thanks for taking the time to review this PR.  I'm in agreement with your sentiments about deleting the translated resources.  In the end, I opted for what I understood the standard approach.  Which is, when the English version of a resource is changed, the translation(s) are deleted to force an update.  

FWIW, an alternative that I considered was to leave the translations as-is, but create an Issue for the resources that needed to be re-evaluated.  That, of course, runs the risk that the Issue is forgotten/missed when putting together 2.5.3.  On the positive side, the translated resources remain available...just a bit inconsistent.  I'm happy to take this approach if it sounds better.

PhilCattivocaratere:
Italian resources are under my care and will be updated as soon as possible. We will not loose anything.

Issue #5974, opened 2022-04-25
Unused and Un-assigned UDT Member inspections 


BZngr:
Version 2.5.2.x (debug)
OS: Microsoft Windows NT 10.0.19044.0, x64
Host Product: Microsoft Office x64
Host Version: 16.0.15028.20228
Host Executable: EXCEL.EXE

**Description**
RD does not flag unused or unassigned User Defined Type members.  

**To Reproduce**
The following code generates 'VariableNotUsed' and 'VariableNotAssigned' inspection results for `mNotUsed`, but no flagged results for the `SecondVal` UDT Member.
```
Option Explicit

Private Type TMCVE
    FirstVal As Long
    SecondVal As Double
End Type

Private this As TMCVE

Private mUsed As Long
Private mNotUsed As Double

Private Sub Class_Initialize()
    mUsed = 1
    this.FirstVal = 1
End Sub

Private Function Sum() As Long
    Sum = mUsed + this.FirstVal
End Function
```

**Expected behavior**
In the above example, RD should flag `SecondVal` as `UDTMemberNotUsed` and `UDTMemberNotAssigned` (if implemented as a dedicated inspection) or flag `SecondVal` as part of the `VariableNotUsed/Assigned` inspection result sets.

MDoerner:
The part of the issue regarding the unused UDT member seems to be fixed already by the linked PR.
Since there is currently no `UDTMemberNotAssignedInspection` I relabel this an enhancement.

PR #6004, opened 2022-07-20
Introduce UDTMemberNotUsedInspection

BZngr:
Related issue: #5974

Rubberduck provides both tools and encouragement to use `Private User Defined Type` (UDT) members in preference to module-level variables.  However, RD lacks any inspections for UDT members that are similar to those for Fields.  

This PR inspects for UDT Members that are declared but never assigned or used.  The `RemoveUnusedDeclarationQuickFix` supports the removal of the declarations flagged by this inspection.


retailcoder (approved):
Awesome, thanks!

PR #6005, opened 2022-07-23
Comments translation

PhilCattivocaratere:
I realized I never translated a lot of comments in resource files. Now most of them are.


Vogel612:
@PhilCattivocaratere the comments are never visible to the user. They are only visible to developers and translators and are intended to explain why a resource string exists and what is put into the relevant placeholders. As such translating them only benefits ... developers or translators that prefer italian.

Thank you very much for working on this, but don't worry if there's untranslated comments :)

PhilCattivocaratere:
> ... but don't worry if there's untranslated comments :)
Ah! I thought they were visibile to user too! As you can understand I known Rubberduck in every little detail.
Comments are done, not so difficult. Ready for next resource strings!
Thank you for your patience with my mess in every PR.

IvenBach:
Every contributing duck is a volunteer. We're appreciative of your translation effort.

Issue #4342, opened 2018-08-31
Indenter doesn't properly handle line-continuation in If statement


daFreeMan:
Win10, Excel2016(desktop), RD .3739

Initial code:

```VBA
Public Function TotalVeryDissatisfied(ByRef year As YearFlag, ByRef printStateFlag As PrintState) As Long

  Dim myClinicItem As ClinicItem
  Dim Sum As Long
  
  Sum = 0
  On Error GoTo ErrHandle
  For Each myClinicItem In this.ClinicItems
    If myClinicItem.GenerateSatSurvey Then
      If (printStateFlag = psPrintState) _
    Or (printStateFlag = psDoNotPrintState And myClinicItem.Name <> STATE_EXCLUSION) Then
        If year = yfLastYear Then
          Sum = Sum + myClinicItem.LastYearVD
        Else
          Sum = Sum + myClinicItem.YtdVD
        End If
      End If
    End If
  Next
  
ExitHere:
  TotalVeryDissatisfied = Sum
  Exit Function
  
ErrHandle:
  Sum = -1
  Resume ExitHere

End Function
```

Note the position of the line beginning with `Or`. After indenting the procedure, I get:

```VBA
Public Function TotalVeryDissatisfied(ByRef year As YearFlag, ByRef printStateFlag As PrintState) As Long

  Dim myClinicItem As ClinicItem
  Dim Sum As Long
  
  Sum = 0
  On Error GoTo ErrHandle
  For Each myClinicItem In this.ClinicItems
    If myClinicItem.GenerateSatSurvey Then
      If (printStateFlag = psPrintState) _
    Or (printStateFlag = psDoNotPrintState And myClinicItem.Name <> STATE_EXCLUSION) Then
        If year = yfLastYear Then
          Sum = Sum + myClinicItem.LastYearVD
        Else
          Sum = Sum + myClinicItem.YtdVD
        End If
      End If
    End If
  Next
  
ExitHere:
  TotalVeryDissatisfied = Sum
  Exit Function
  
ErrHandle:
  Sum = -1
  Resume ExitHere

End Function
```

Again, note the position of the line starting with `Or...` - it's in the exact same column. I do know that the indenter ran because it also gobbled up my leading/trailing blank line that is supposed to leave.

However, 2 MCVEs of:

```VBA
Private Sub glarp()

  Dim foo As Boolean
  Dim bar As Boolean
  Dim baz As Boolean

  If foo Then
    If bar _
[...]

Issue #5960, opened 2022-03-16
More inconsistent indentation with line continuation 


daFreeMan:
**Rubberduck version information**
```
Version 2.5.2.6062
OS: Microsoft Windows NT 10.0.19042.0, x64
Host Product: Microsoft Office x64
Host Version: 16.0.14326.20784
Host Executable: MSACCESS.EXE
```

**Description**
Another instance of inconsistent indentation by the indenter. In this case, it's a line that concatenates the results of `string` returning functions that doesn't work.

**To Reproduce**
```
  EmailFooter = Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter1") & _
Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter2") & _
Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter3")
```
indents to
```
  EmailFooter = Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter1") & _
                                                                                                         Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter2") & _
                                                                                                         Tools.GetGeneralConfig("EmailRecord.Class_Initalize", "SurveyEmail", "OccHealthFooter3")
```

However, this code:
```
  If DevelopmentMode Then
    body = "This is a TEST EMAIL" & SINGLE_LINE_BREAK & _
"It would have been emailed to: " & DeliveryEmailAddress(True) & "<hr>" & DOUBLE_LINE_BREAK & body
  End If
```
indents to:
```
  If DevelopmentMode Then
    body = "This is a TEST EMAIL" & SINGLE_LINE_BREAK & _
           "It would have been emailed to: " & DeliveryEmailAddress(True) & "<hr>" & DOUBLE_LINE_BREAK & body
  End If
```
**Expected behavior**
consistently lining up line continuations on the `=` as in the second example

**Screenshots**
N/A

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/8266836/RubberduckLog.txt)
NOTE: I've been doing a fair bit of editing - there is going to be a lot of unrelated stuff. Any errors should be at/near the end of the log.

**Additional context**
Related: #5852 , #4632 , #4342

PR #6009, opened 2022-08-19
Fix inconsistant indentation #5960

awb95:
Fixes Indenter #5960 .
On ')" - Function Call End - the Alignment got resetted, whereby often the LineConttinuiation '_" was taken as Alignment position.

After that I had to adjust one Test that worked not as intended i think.

### DIFF
--- a/Rubberduck.SmartIndenter/LogicalCodeLine.cs
+++ b/Rubberduck.SmartIndenter/LogicalCodeLine.cs
@@ -220,7 +220,7 @@ namespace Rubberduck.SmartIndenter
                         {
                             var finished = _alignment.Count == stackPos + 1;                            
                             var token =_alignment.Pop();
-                            if (token.Type == AlignmentTokenType.Function && token.NestingDepth == _nestingDepth - 1)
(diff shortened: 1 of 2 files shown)


retailcoder (approved):
Thank you for these fixes!

awb95:
I think this PR Fixes #4342 also.

BZngr:
@awb95  Thank you for all the fixes regarding the `Indenter`.  If you're not certain that the issues identified above are fixed, one option is to submit a new PR that adds unit tests replicating the issues' examples.  Assuming the tests succeed without other code changes, then the new PR can be used to close the above issues with more confidence.

awb95:
I discovered these additional problems when I was looking for problems with the indenter. I tested the given examples in the VBE. I didn't count all the spaces, but at first glance they looked much better than before.
Maybe it's a good idea to add some general tests for continued bracket expressions/function calls. But taking the whole examples as unit tests makes little sense in my eyes.

BZngr:
OK...If your checks in the VBE looked good to you, then it would seem OK to go ahead and close those additional issues as resolved.  4 issues with 1 PR..nice!

PR #6040, opened 2022-10-20
Italian translation for new resources from #6035 PR

PhilCattivocaratere:
Italian translation for new resources from #6035 PR


Vogel612:
lightning fast, as always :)

Issue #5961, opened 2022-03-19
Building RD fails on olewoo


AndreasMatthias:
Trying to build Rubberduck I get the following error message concerning olewoo:

 C:\Users\andreas\Rubberduck\Rubberduck.Deployment\Rubberduck.Deployment.csproj(49,5): error MSB4062: The "RubberduckPreBuildTask" task could not be loaded from the assembly C:\Users\andreas\Rubberduck\Rubberduck.Deployment\..\Rubberduck.Deployment.Build\bin\Rubberduck.Deployment.Build.dll. Could not load file or assembly 'olewoo, Version=1.2.6710.37528, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.

Any help on this is greatly appreciated.

My build environment:

Microsoft Visual Studio Community 2022
Version 17.1.1
VisualStudio.17.Release/17.1.1+32228.430
Microsoft .NET Framework
Version 4.8.04084

Vogel612:
I can reproduce this on a fresh clone on a machine without MSOffice installation. We should really switch to explicitly referencing this as a nuget package. This seems to be tangentially related to #5969, the base issue seems to be that we're using a positively ancient version of Microsoft.Build.Framework and related.

This should be fixed by pulling the involved dependencies to an up to date version (i.e. 17.2.0 at time of writing) and updating the Rubberduck.Deployment.Build tasks accordingly.

pflugs30:
I get this error when building a fresh pull of RD next branch in VS2022 **with** an MSOffice installation.  I've also been able to implement all the changes found in PR #5982 on my local machine (since the PR isn't yet merged).  However, I can't get past this 'olewoo' error.  Here's my version:

> The "RubberduckPreBuildTask" task could not be loaded from the assembly **C:\dev\personal\Rubberduck\Rubberduck.Deployment\..\Rubberduck.Deployment.Build\bin\Rubberduck.Deployment.Build.dll**. Could not load file or assembly 'olewoo, Version=1.2.6710.37528, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.

[...]

PR #6063, opened 2022-12-29
Fixing various build issues in VS2022 (64-bit)

pflugs30:
- Updated `TargetFramework` in `RubberduckBaseProject.csproj` from `net461` to `net462` since 4.6.2 is the latest supported 4.6 version.
- Updated `TargetFramework` in `RubberduckBaseMetaProject.csproj` from `net46` to `net462` since 4.6.2 is the latest supported 4.6 version.
- Updated `RubberduckPostBuildTask.NetToolsDir` in `Rubberduck.Deployment.csproj` to point to `NETFX 4.6.2 Tools`.
- Converted the `stdole` reference in both the `Rubberduck.VBEditor` and `Rubberduck.VBEditor.VBA` projects from using the MS Build Tools Program Files install to using the `stdole` NuGet package (version 7.0.3300.0) to correct a reference error.
- Added 64-bit versions of `olewoo.dll` and `olewoo_interop.dll` as received from @bclothier.


pflugs30:
Yes, I think we have a build error due to the x86 / x64 versions of `olewoo.dll` tested in this PR.  I will see if I can include both DLL versions and use them conditionally.

retailcoder:
Hm, I wonder if it would be possible to target 4.8 now 🤔

pflugs30:
The build fails due to this error:
```
C:\projects\rubberduck\Rubberduck.Deployment\Rubberduck.Deployment.csproj(49,5): error MSB4062: 
The "RubberduckPreBuildTask" task could not be loaded from the assembly C:\projects\rubberduck\Rubberduck.Deployment\..\Rubberduck.Deployment.Build\bin\Rubberduck.Deployment.Build.dll. 
Could not load file or assembly 'olewoo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. 
The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.
```

I think that's related to [this comment](https://github.com/rubberduck-vba/Rubberduck/issues/5961#issuecomment-1367434254).  I need to do some more testing to make that happen.

pflugs30:
Okay, well THAT change didn't work.  Something about this build process is still running in 32-bit mode.  Guess I'll have to do more investigation into the Appveyor process, or I need to figure out how the build process works locally (or both).

retailcoder:
No wait, I think that one's the token. It failed trying to push a comment on the PR. I just merged the updated token, is this branch up-to-date?

pflugs30:
Yes, I branched yesterday from next.  Should be up to date.  We may have multiple failures.

retailcoder:
Yeah next has the correct token as of a couple of minutes/hours ago 🤖

bclothier:
Please don't merge until confirmed that this builds OK on VS 2022 64-bit. It did build OK on VS 2019 32-bit but that's only half of the equation.

RubberduckTests/Symbols/SelectedDeclarationProviderTests.cs
```diff
@@ -527,7 +527,7 @@ public void Identify_NamedParameter_Parameter_FromExcel()
                 .AddProjectToVbeBuilder()
                 .Build();
 
-            var (expected, actual) = DeclarationsFromParse(vbe.Object, DeclarationType.Parameter, "Link", "EXCEL.EXE;Excel._Worksheet.Paste");
+            var (expected, actual) = DeclarationsFromParse(vbe.Object, DeclarationType.Parameter, "Link", "EXCEL.EXE;Excel.Worksheet.Paste");
```
bclothier, line 530: Do we really need to change this? Seems unrelated to the PR and it's failing in AppVeyor.

Vogel612:
This PR replaces #5982 (by completely subsuming it)

RubberduckTests/Symbols/SelectedDeclarationProviderTests.cs
pflugs30: I changed it because it was the only test out of the several thousand (and I am thrilled to see so many tests!) that failed.  I investigated the test, and I felt it was a typo or mistake.  I'm accustomed to making sure all active tests pass on my PRs, so I changed it.  I didn't mean anything else by it.  :-)

pflugs30:
> Please don't merge until confirmed that this builds OK on VS 2022 64-bit. It did build OK on VS 2019 32-bit but that's only half of the equation.

Thank you for picking up the ball and advancing further.  Unfortunately, I can't compile the `Rubberduck.sln` solution anymore with the same error as before:
image

I checked the output from `RubberduckMeta.sln`, and the build process copied the Olewoo x86 DLLs:
image

Looking at the clever change to `Rubberduck.Deployment.Build`, it sure seems the process should have used the x86 versions:
image

So, I will try to see why the 64-bit VS2022 is still building the x86 version into the Meta solution.  I wanted to get some response out to the team ASAP.

Also, this may sound picky, but I recommend we use "x86" instead of "x32".  It's more correct:
image

bclothier:
> I checked the output from RubberduckMeta.sln, and the build process copied the Olewoo x86 DLLs

Just to rule this out - have you tried from a clean build? I'm not sure it'll be smart enough to replace the DLLs if it exists already from prior builds.

> Also, this may sound picky, but I recommend we use "x86" instead of "x32". 

FWIW, I dislike it because it sorts higher than the 64-bit and could be read as 86 > 64 when in fact it's 64 > 32 and the term "x86" could easily encompass both 32-bit and 64-bit (it even started out as 16-bit). That said, I'm not terribly attached to the "x32" naming convention as long as it's consistent everywhere else.

pflugs30:
Yes, I did try clean builds.  Here are the build outputs:
[Uploading BuildOutput.zip…]()

> FWIW, I dislike it because it sorts higher than the 64-bit and could be read as 86 > 64 when in fact it's 64 > 32 and the term "x86" could easily encompass both 32-bit and 64-bit (it even started out as 16-bit). That said, I'm not terribly attached to the "x32" naming convention as long as it's consistent everywhere else.

I completely understand that perspective.  I wish "the powers that be" hadn't named it "x86".  But since it is that way, I prefer consistency with the industry.  Again, that being said, I'm the new guy, and I will humbly defer.  :-D

pflugs30:
Looking through the `RubberduckMeta.sln` build output, I found this snippet where the process resolved the references:
```
2>    Primary reference "olewoo, Version=1.2.6710.37528, Culture=neutral, PublicKeyToken=null".
2>        Resolved file path is "C:\dev\personal\Rubberduck\Rubberduck.Deployment.Build\OleWoo\x32\olewoo.dll".
2>        Reference found at search path location "{CandidateAssemblyFiles}".
2>        Found related file "C:\dev\personal\Rubberduck\Rubberduck.Deployment.Build\OleWoo\x32\olewoo.dll.config".
2>    Primary reference "olewoo_interop, Version=1.0.6710.35728, Culture=neutral, PublicKeyToken=null".
2>        Resolved file path is "C:\dev\personal\Rubberduck\Rubberduck.Deployment.Build\OleWoo\x32\olewoo_interop.dll".
2>        Reference found at search path location "{CandidateAssemblyFiles}".
2>            For SearchPath "{CandidateAssemblyFiles}".
2>            Considered "OleWoo\x32\olewoo.dll",
2>    			but its name "olewoo"
2>    			didn't match the expected name "olewoo_interop, Version=1.0.6710.35728, Culture=neutral, PublicKeyToken=null".
```

Seems it determined the bitness as 32-bit.  Perhaps we're barking up the wrong tree?  Is there something instead about the `Rubberduck.Deployment` build process that uses 64-bit instead of the 32-bit process?

pflugs30:
I'd like to report another test which failed.  I set up an x86 platform for both solutions, and I added an `Architecture` attribute to the `Rubberduck.Deployment` project:

```
  <UsingTask TaskName="RubberduckPreBuildTask" AssemblyFile="..\Rubberduck.Deployment.Build\bin\Rubberduck.Deployment.Build.dll" Architecture="x86" />
  <UsingTask TaskName="RubberduckPostBuildTask" AssemblyFile="..\Rubberduck.Deployment.Build\bin\Rubberduck.Deployment.Build.dll" Architecture="x86" />
```

Even after cleaning and building, the Olweoo error still persists.  I might be at the extent of what I know.... :-)

pflugs30:
I think we've run into an actual bug with MSBuild...
https://github.com/dotnet/msbuild/issues/2360

bclothier:
Actually, I don't think the MSBuild bug as reported is applicable here. The project is `Any CPU` that's referencing a unmanaged C++ DLL that is not a part of the build process (e.g. we're just copying around the DLL). I got a VS2022 instance set up and from my testing, it appears that the issue is that Visual Studio won't update the reference if it already is present. Therefore, we must force it to remove the existing reference, then re-add it and it will then use the correct reference and build it. `HintPath` as the name implies is just a "hint" and not an actual directive to "use this one, not that  one". 

Therefore, Visual Studio needs to be smacked out of its stupidity then everything will be A-OK. This seems to build OK on both my VS 2019 32-bit and VS 2022 64-bit. If you can confirm it does for yours, we could consider this completed.

bclothier:
Just as a FYI - even though the AV's VS is 2022, it apparently uses [32-bit MSBuild](https://ci.appveyor.com/project/rubberduck-vba-releasebot/rubberduck/builds/45805026?fullLog=true#L589). On my VS2022, it came out with "x64". Go figure. 

>  Resolved MSBuild Bitness: x32

pflugs30:
Thanks for sticking with this issue to the end.  I pulled all changes and was able to build all projects successfully!  I'm going to try the tests again, but I wanted you to know ASAP.

pflugs30:
FWIW, I'm still getting the failure in the one test:
image

```
 Identify_NamedParameter_Parameter_FromExcel
   Source: SelectedDeclarationProviderTests.cs line 515
   Duration: 1.7 sec

  Message: 
  Expected: <Rubberduck.Parsing.Symbols.ParameterDeclaration>
  But was:  <Rubberduck.Parsing.Symbols.ParameterDeclaration>


  Stack Trace: 
SelectedDeclarationProviderTests.Identify_NamedParameter_Parameter_FromExcel() line 532
```

When I debug through the test, I see the following:
image

So, the `expected` Scope and ParentScope are `EXCEL.EXE;Excel._Worksheet.Paste"` and the `actual` Scope and ParentScope are `EXCEL.EXE;Excel.Worksheet.Paste"`.  I haven't explored these tests too much.  Is there any chance this issue is related to 64-bit vs 32-bit?

bclothier:
I would think it unlikely. On my VS2022, it passes. I would treat this as a separate issue, beyond the scope of the PR.

PR #6118, opened 2023-06-21
fix codecov yml

Vogel612:
This should finally get us some actual coverage back :D


Vogel612:
for whatever it's worth, at least the quoting of wildcard strings should not have been necessary per yaml spec, but alas, the yaml parsing ecosystem is insufficiently consolidated...

Issue #6187, opened 2023-12-10
Recent Resolver Error on VBE-compilable code


pflugs30:
**Rubberduck version information**
The info below can be copy-paste-completed from the first lines of Rubberduck's log or the About box:

Version 2.5.9.6314
OS: Microsoft Windows NT 10.0.22631.0, x64
Host Product: Microsoft Office x64
Host Version: 16.0.16130.20846
Host Executable: EXCEL.EXE

**Description**
After updating to the recent public release of Rubberduck, I started seeing a Resolver Error on compilable code which used to be just fine.  Using the log files, I isolated the issue to a single method.

**To Reproduce**
1. Open Excel and create a new blank workbook.
2. Open the VBE and add a single code module.
3. Paste the following code: [Module1.zip](https://github.com/rubberduck-vba/Rubberduck/files/13626617/Module1.zip)
4. Compile the code to show the VBE can parse it.
5. Click the Rubberduck menu --> Refresh
6. Observe Rubberduck will show a **Resolver Error**.

**Expected behavior**
Rubberduck should be able to parse and resolve this code just fine.

**Screenshots**
N/A

**Logfile**
[RubberduckLog.txt](https://github.com/rubberduck-vba/Rubberduck/files/13626614/RubberduckLog.txt)

**Additional context**
Specifically, I've discovered these lines to be the problem:
```
        If utc_HasOffset Then
            ParseIso = ParseIso - utc_Offset
        End If
```

When these lines are commented, then RD can parse and resolve all day.  All these variables are defined and found in the function.

MDoerner:
Thanks for the report. 
I think I have an idea why there is an exception in the resolution. However, if that code is the culprit, I think something imight be already wrong when parsing the code. The resolver  seems to fail on a single line if statement with empty then statement.

I will have a look later and see what I can do.

pflugs30:
Please let me know if you'd like me to send a broader log file.  I think what I sent covers the parsing, but I could be mistaken.

jimmyli97:
If anyone else has this issue then the workaround for now is to remove the empty line e.g. ``If utc_NegativeOffset Then utc_Offset = -utc_Offset`` instead of ``If utc_NegativeOffset Then: utc_Offset = -utc_Offset``

Irubataru:
[...]

PR #6196, opened 2024-01-27
Recent Grammar issues

tommy9:
Fixes for #6187, #6194

Single line if statements had been made too flexible, in particular the rule ifWithEmptyThen where all the elements following the THEN were optional, letting the singleLineIfStmt match what is really an ifStmt:

`ifWithEmptyThen : IF whiteSpace? booleanExpression whiteSpace? THEN whiteSpace? emptyThenStatement? singleLineElseClause?;`

Now the rule requires at least one of emptyThenStatement or singleLineElseClause.

The user defined type member rule has been modified to allow the asTypeClause to be omitted but only if the member is an array. This ensures valid (though not defined in the language specification) examples through as in #6187 

`optionalArrayClause : ((arrayDim whiteSpace)? asTypeClause | arrayDim);`

### DIFF
--- a/Rubberduck.Parsing/Grammar/VBAParser.g4
+++ b/Rubberduck.Parsing/Grammar/VBAParser.g4
@@ -543,6 +543,7 @@ subroutineName : identifier;
 
 // 5.2.3.3 User Defined Type Declarations
 // member list includes trailing endOfStatement
+// To support actual VBA behaviour, had to change optionalArrayClause to allow a standalone arrayDim without the asTypeClause - see issue #6194
 udtDeclaration : (visibility whiteSpace)? TYPE whiteSpace untypedIdentifier endOfStatement udtMemberList END_TYPE;  
 udtMemberList : (udtMember endOfStatement)+; 
 udtMember : reservedNameMemberDeclaration | untypedNameMemberDeclaration;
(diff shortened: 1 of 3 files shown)


MDoerner:
Thanks for working on the mess I caused. I would have liked to do it myself for a whole, but my real work is in crunch mode at the moment.

I think the fix actually does not address the full problem with the single line if statement. It certainly removes the rule confusion on multi-line if statements, but it does not address the resolver issue on single line if statements with empty then clauses. I think that is caused by me forgetting to put a null check into the method handling single line if statement contexts in the resolver when I made the one part optional. I believe that still needs to be added.

tommy9:
Thanks, I didn't think to change the lack of null checking because the grammar fix should mean it there should never be an empty else clause and an empty then clause. However, who knows what odd code could end up taking that path, so I agree it is safest to add a check in. I'll update the pull request.

MDoerner (approved):
Thanks for dealing with this.

pflugs30:
@retailcoder something is wrong with the release URL.  When I try to download the updated build from the website for testing, I get a 404 on the redirect to Github.
image

The link with the 404 is:
https://github.com/rubberduck-vba/Rubberduck/releases/download/Prerelease-v2.5.9.6296/Rubberduck.Setup.2.5.9.6319-pre.exe

It should be:
https://github.com/rubberduck-vba/Rubberduck/releases/download/**Prerelease-v2.5.9.6319**/Rubberduck.Setup.2.5.9.6319-pre.exe

(without the asterisks, of course :-) )

retailcoder:
Argh, my bad - with the web API being down I'm manually inserting the records in the database... I messed that one up, will update over lunch (30-60 minutes)

retailcoder:
Fixed!

PR #6257, opened 2025-02-23
Updated Italian translation.

PhilCattivocaratere:
Updated Italian translation.


Vogel612 (approved):

Vogel612:
Thank you for your continued work on the italian localization :)

PhilCattivocaratere:
> Thank you for your continued work on the italian localization :)

I am not here so frequently but Rubberduck is always in my mind.

PR #6265, opened 2025-03-30
Export function now allows to choose a file name

Kristupasc:
PR to fix issue [export function should allow to change the file name #6135 ](https://github.com/rubberduck-vba/Rubberduck/issues/6135)

New behavior: The file now saves with the name and extension given in the "Save as" file explorer window instead of the module name inside the code explorer. If no file name is given, it defaults to the module name just as it used to work before.

Explanation of the changes: 
`ExportAsSourceFile` now has a conditional check where it sees whether a custom file name was provided. 
`PromptFileNameAndExport` now passes the full path (including the file name), instead of only the directory path.

NOTE: Due to these changes, the unit test `ExportModule_ExpectExecution` in `RubberduckTests\CodeExplorer\CodeExplorerViewModelTests.cs` was updated. The test now expects that when a custom file name is provided, ExportAsSourceFile is called with the full file path (folder plus filename) instead of just the folder.

### DIFF
--- a/Rubberduck.Core/UI/CodeExplorer/Commands/ExportCommand.cs
+++ b/Rubberduck.Core/UI/CodeExplorer/Commands/ExportCommand.cs
@@ -101,7 +101,6 @@ namespace Rubberduck.UI.CodeExplorer.Commands
                 var component = ProjectsProvider.Component(qualifiedModule);
                 try
                 {
-                    //var path = Path.GetDirectoryName(dialog.FileName);
                     var path = dialog.FileName; // This makes it that the file is named not as the module name, but as the file name the user provided in the explorer.
                     component.ExportAsSourceFile(path, false, true); // skipped optional parameters interfere with mock setup
                 }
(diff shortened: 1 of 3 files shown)


Rubberduck.Core/UI/CodeExplorer/Commands/ExportCommand.cs
```diff
@@ -101,7 +101,8 @@ public bool PromptFileNameAndExport(QualifiedModuleName qualifiedModule)
                 var component = ProjectsProvider.Component(qualifiedModule);
                 try
                 {
-                    var path = Path.GetDirectoryName(dialog.FileName);
+                    //var path = Path.GetDirectoryName(dialog.FileName);
```
MDoerner: Can you please delete the code instead of only commenting it out?

MDoerner (commented):
Thanks for the PR. It looks good to me in general.

Just to be sure:
Have you tested that exporting projects and syncing projects still works as before?
I cannot recall whether the export command is involved there.

Rubberduck.Core/UI/CodeExplorer/Commands/ExportCommand.cs
Kristupasc: Good catch! I removed it in the new commit

Kristupasc:
> Thanks for the PR. It looks good to me in general.
> 
> Just to be sure: Have you tested that exporting projects and syncing projects still works as before? I cannot recall whether the export command is involved there.

Yes, I have looked through all of the usages for this function and I tried around some example projects practically, and they all seem to function exactly the same.

MDoerner (approved):
Thanks for also checking the other export behaviour.

retailcoder (approved):

